PR #2, opened 2015-12-14
[Ninja] Support -j (jobs) parameter to Ninja.

grp:
Many uses of Ninja pass -j to specify how many jobs to run in
parallel. Replace the non-standard --parallel and --no-parallel
options with -j (and an expanded form --jobs) to match Ninja.

### DIFF
--- a/lib/Commands/NinjaBuildCommand.cpp
+++ b/lib/Commands/NinjaBuildCommand.cpp
@@ -97,12 +97,10 @@ static void usage(int exitCode=1) {
           "load the manifest at PATH [default='build.ninja']");
   fprintf(stderr, "  %-*s %s\n", optionWidth, "-k <N>",
           "keep building until N commands fail [default=1]");
-  fprintf(stderr, "  %-*s %s\n", optionWidth, "--no-parallel",
-          "build commands serially");
-  fprintf(stderr, "  %-*s %s\n", optionWidth, "--parallel",
-          "build commands in parallel [default]");
   fprintf(stderr, "  %-*s %s\n", optionWidth, "-t, --tool <TOOL>",
           "run a ninja tool. use 'list' to list available tools.");
+  fprintf(stderr, "  %-*s %s\n", optionWidth, "-j, --jobs <JOBS>",
+          "number of jobs to build in parallel [default=cpu dependent]");
   fprintf(stderr, "  %-*s %s\n", optionWidth, "--no-regenerate",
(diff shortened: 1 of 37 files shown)


ddunbar:
Can you comment a bit on why you want this?

I will probably cave and support it to be compatible with Ninja, but the reason I went this route was because I really wanted llbuild to just DTRT w.r.t. building as quickly as possible (and I didn't want to commit to implementing a system which was based around limiting to a particular concurrent process count).

I know that one common use of -j is when using a system which is backed by cloud based build tools, so you really want the engine to just dispatch as much work as is ever possible. Is that the use case you are in, or something else?

gribozavr:
Another usecase is to limit the number of compiler invocations to be lower than the number of cores, if your system is RAM constrained.

ddunbar:
That, however, is exactly the kind of use case I want llbuild to just do the right thing automatically -- for example, by monitoring the resource usage of the jobs being spawned and the system memory pressure, and adapting accordingly.

gottesmm:
> On Dec 14, 2015, at 10:57 PM, Daniel Dunbar <notifications@github.com> wrote:
> 
> That, however, is exactly the kind of use case I want llbuild to just do the right thing automatically -- for example, by monitoring the resource usage of the jobs being spawned and the system memory pressure, and adapting accordingly.
> 
> 

Can you guarantee that llbuild will always do the right thing? Why not give power users an escape hatch that enables them to stay in llbuild instead of using something else (or worse) becoming disillusioned?

grp:
The main reason on my side was compatibility. We (Facebook) have a variety of scripts internally that call Ninja, and it's much easier to try using llbuild immediately if I can assume that they work without modification there.

I've used `-j` specifically for concurrent builds of unrelated projects. For example, when testing a build tool on a large real-world project while preparing a new version of the build tool (with additional logging). In that case, I estimated the number of jobs to favor the the real-world project, in the hope that when I finish analyzing the output from the previous version of the build tool, the next version would be ready. To @gottesmm's point, it would be hard for an intelligent scheduler to figure out an intention like that even if it did the right thing for the vast majority of cases.

ddunbar:
Makes sense, as I said earlier I am fine taking it for compatibility with Ninja. Can you rebase this PR though so that it merges cleanly?

grp:
Rebased.

ddunbar:
Thanks Grant!

PR #3, opened 2015-12-14
[Ninja] Document -t, --tool option and add the "list" tool.

grp:
The -t option to Ninja is used to run subtools. One subtool, "targets",
was already implemented, but this support was not in the help. Also, to
determine which tools are available, the "-t list" option is useful;
especially so while only a subset of the Ninja subtools are implemented.


ddunbar:
Thanks! Feel like adding a test for -t list, just to make sure it exists and at least does something?

grp:
Good idea. I'll sent out another PR for that.

PR #5, opened 2015-12-14
[Ninja] Rename --simulate to -n and --dry-run to match Ninja.

grp:
Existing uses of Ninja expect certain options, this provides the -n
option used for dry runs. The existing --simulate option is renamed
--dry-run to match the Ninja option.


ddunbar:
It isn't actually true that --simulate is a substitute for -n, which is why I haven't done this rename. --simulate will assume all cached results in the database are valid (which means it won't print some things that would happen), and will bring the database up to date if there aren't results (which means the things that it does print might not run on the next actual build, which is bad).

This is an option I found useful for development and benchmarking, but it probably should be removed from the Ninja tool (or at least hidden).

I agree we generally also should support -n, but it will require a different implementation.

grp:
Thanks, I'll take a look at what an implementation of the real `-n` option would need.

PR #6, opened 2015-12-18
More verbosity if -v is specified repeatedly

mxcl:
Currently the only usage of twice -v is in the swift build frontend where we pass -v through to swiftc.

### DIFF
--- a/include/llbuild/BuildSystem/BuildSystem.h
+++ b/include/llbuild/BuildSystem/BuildSystem.h
@@ -93,6 +93,8 @@ public:
   /// Called to determine whether to show verbose status.
   virtual bool showVerboseStatus() = 0;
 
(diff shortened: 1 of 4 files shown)


ddunbar:
I'm not sure if this is worth putting in llbuild proper -- couldn't we just make the package manager put the necessary "-v" directly into the other-args passed to swiftc?

mxcl:
Oh, indeed. Didn't occur to me.

mxcl:
Well, this way I need to rebuild the yaml based on the command line arguments passed. Currently I rebuild the yaml every time anyway, but this is non-ideal.

I'll push this anyway as it is less intrusive to llbuild, but I worry we'll introduce a bug later because of this.

PR #7, opened 2016-02-22
[docs] Clang tool does not execute bin/sh

modocache:
Unlike the shell tool, the Clang tool only takes a single argument. That argument is not executed using `/bin/sh`.

---

I wasn't able to find many examples of usages of the Clang tool, aside from `tests/BuildSystem/Build/discovered-compiler-deps.llbuild`. Am I right in understanding that the Clang tool merely executes a shell command, and doesn't have any Clang-specific logic? I feel like I'm misunderstanding something, so advice appreciated. :bow:

### DIFF
--- a/docs/buildsystem.rst
+++ b/docs/buildsystem.rst
@@ -337,8 +337,7 @@ A tool used to invoke the Clang compiler.
      - Description
 
    * - args
-     - A string or string list indicating the command line to be executed. If a
-       single string is provided, it will be executed using ``/bin/sh -c``.
+     - A string indicating the command line to be executed.
 
    * - deps
      - The path to a Makefile fragment (presumed to be output by the compiler)


ddunbar:
The Clang tool also has support for discovered dependencies, under the:
  https://github.com/apple/swift-llbuild/blob/master/docs/buildsystem.rst#clang-tool

It hasn't been updated for the addition of string list parsing support, which is why the args key is a manually split string instead of a string list as it should be. This is just a bug.

modocache:
Oops, my bad! I'll send another pull request if I think of something to add--perhaps your note on discovered dependencies, once I grok how those work.

To explain my reasoning yesterday when sending this: reading the documentation, the `clang` and `shell` tools had such similar descriptions, I initially assumed they were copy-pasted and potentially incorrect. The fact that the `clang` tool invoked `sh` and not `clang`, for example, really threw me off.

Thanks for the explanations! I'll keep poking around to get a better understanding of llbuild.

ddunbar:
Cool, let me know if you have questions.

There is actually a ton more documentation that I want to write, but haven't had time to get to -- some pokes in the right direction might help.

Maybe the docs here could just clarify the intent of the tool. The idea of having a Clang specific tool was that, eventually, it could be a place to house a lot of custom compiler behaviors if we had **deep** integration with Clang (e.g., if we could invoke it via a library interface).

There is also a set of useful common behaviors provided by all "GCC compatible" compilers, for example the handling of the -MM/-MD/-MF flags that build systems use to get the discovered dependencies. Those behaviors eventually should be factored into a tool which is more specific than just the ShellTool, but less specific than Clang.

modocache:
Awesome, yes! This is exactly the kind of insight I was hoping to tease out with a pull request :stuck_out_tongue_winking_eye: 

I'm excited to learn more about llbuild and see how it'll integrate with SwiftPM and Swift's own build scripts over time.

ddunbar:
Encoded basically those notes into: 18996d152abbc08e4aea13e925fc6063b0bdb5a8

You can already use llbuild to build Swift if you `ln -s path/to/llbuild ninja` in your PATH. :)

PR #11, opened 2016-03-24
Add “archive” tool

mxcl:
Constructs an archive from provided inputs, intended for static library archives.

Takes `inputs` and `outputs` only.

Any virtual nodes are ignored when constructing the `ar` instantiation.

I consider this PR incomplete:

- [x] documentation (where to put it?)
- [x] using `inputs`/`outputs` as the only attributes seems inconsistent, so should I do something else?
- [x] `ArchiveShellCommand` probably is lacking some basic checks
- [x] should I add tests, if so, where?

So I await some feedback.

### DIFF
--- a/docs/buildsystem.rst
+++ b/docs/buildsystem.rst
@@ -564,3 +564,21 @@ exact version of the Swift compiler in use (as reported by ``swiftc
      - A string or string list indicating other arguments passed to the
        ``swiftc`` executable. Examples of individual values include
        ``"-enable-testing"`` or ``"-Onone"``.
+
+Archive Tool
+------------
+
+**Identifier**: *archive*
+
+A tool used to create an archive (``.a``)
+
+All non-virtual inputs are archived. Only one non-virtual output may be
+specified, this is inferred to be the archive file that this tool produces.
+
+A typical use for this tool is creating static libraries.
+
+.. note::
+
+   FIXME: currently the archive is always recreated entirely, it would be
+   preferable in future to correctly update/delete/create the archive file
+   as required.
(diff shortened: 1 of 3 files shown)


ddunbar:
Documentation goes in `docs/buildsystem.rst` currently, check the other tools for precedent.

Tests go in `tests/BuildSystem/Build`, let me know if you have trouble running them. Check that dir for some examples of what kinds of stuff I usually check (typical is (a) initial build, (b) build with no change, (c) build with an incremental change).

Will do a more complete review ASAP, not sure yet about the inputs/outputs question.

ddunbar:
After pondering on it a bit I'm fine with the current strategy of just deriving the inputs and outputs from the declared dependency edges. As the FIXME in the code notes I would probably add an error check (you can do this when the "Command" is run) that at least one output and one input exists.

We may eventually want to split these out, but I don't see a reason to overcomplicate it now and that is consistent with what "mkdir" and "symlink" are currently doing.

mxcl:
Revised, please advise.

ddunbar:
Can you make it so the change will still respect a `description` field, if set, and otherwise fall back to the default here? This is something I want clients to be able to have control over.

Also, can you squash?

mxcl:
Done. Will squash, rebase and merge if you OK all the changes.

ddunbar:
LGTM, thanks!!

mxcl:
@swift-ci Please test

mxcl:
@swift-ci Please test

mxcl:
@swift-ci Please test

PR #12, opened 2016-04-05
Add .clang-format file

beltex:
Hello `swift-llbuild`! :)

#### What's in this pull request?

Adds a `.clang-format` file using LLVM style as does the [Swift compiler](https://github.com/apple/swift/blob/master/.clang-format). Assuming there is a desire to use it! If not, my apologies!

#### Why merge this pull request?

Defines the coding style desired, and will be picked up by `clang-format` if used (with the `-style=file` option that is).

#### What are the downsides to merging this pull request?

If `clang-format` is never used, it's a wasted file.

#### Resolved bug number: (None)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+BasedOnStyle: LLVM
+DerivePointerAlignment: true
+PointerAlignment: Left


ddunbar:
Thanks for the PR!

I'm ok with having a .clang-format file, but I would like it to match the convention such that the diffs it would generate wouldn't change the style (just fix inconsistencies).

One major thing which is different in the llbuild style versus the LLVM style is that llbuild uses Google style for '*' and '&' placement. Do you know if it possible to override this element of the style? I haven't followed clang-format development and last I looked I did not see an option for it.

beltex:
Hello @ddunbar!

> Thanks for the PR!

np!

> I'm ok with having a .clang-format file, but I would like it to match the convention such that the diffs it would generate wouldn't change the style (just fix inconsistencies).

Cool!

> One major thing which is different in the llbuild style versus the LLVM style is that llbuild uses Google style for '*' and '&' placement.

Using the latest `clang-format` (`3.9.0`) and dumping the configs for each via `clang-format -style=<STYLE> -dump-config`, believe these are the options your referring to, correct? (these are the Google settings, the LLVM ones are `false` and `Right` respectively.)

```
DerivePointerAlignment: true
PointerAlignment: Left
```

In which `DerivePointerAlignment ` is defined as, _"If `true`, analyze the formatted file for the most common alignment of `&` and `*`. `PointerAlignment` is then used only as fallback."_ **[1]**

> Do you know if it possible to override this element of the style?

I believe so! Simply defining them in the `.clang-format` file should override the `BasedOnStyle` setting (did a quick test to confirm) **[2]**. If those are indeed the options mentioned above, I can make the change.

[[1 - Clang-Format Style Options - Configurable Format Style Options]](http://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options)
[[2 - Clang-Format Style Options - Configuring Style with clang-format]](http://clang.llvm.org/docs/ClangFormatStyleOptions.html#configuring-style-with-clang-format)

ddunbar:
Ah yes, that does make the total diff from forcing everything to conform go down considerably. Will merge if you update...

beltex:
Awesome, done!

PR #14, opened 2016-04-11
[gardening] Fix cosmetic and non-cosmetic typos

hitstergtd:
**Changes made:**
* `s/uniqu/unique/`
* `s/for/from/` in one place because latter makes contextual sense

Should not affect the build.


hitstergtd:
@ddunbar, thanks! 👍

PR #15, opened 2016-04-12
Optimize storage of dependency keys

aciidb0mb3r:
Bug: https://bugs.swift.org/browse/SR-1140

@ddunbar This is working except that no entry from keys table is deleted right now. 
Can you check if this is right approach.

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -34,7 +34,7 @@ using namespace llbuild::core;
 namespace {
 
 class SQLiteBuildDB : public BuildDB {
-  static const int currentSchemaVersion = 4;
+  static const int currentSchemaVersion = 5;
 
   sqlite3 *db = nullptr;
 
@@ -126,12 +126,20 @@ public:
       }
       if (result == SQLITE_OK) {
         result = sqlite3_exec(
+          db, ("CREATE TABLE key_names ("
+               "id INTEGER PRIMARY KEY, "
+               "key STRING UNIQUE);"),
+          nullptr, nullptr, &cError);
+      }
+      if (result == SQLITE_OK) {
+        result = sqlite3_exec(
           db, ("CREATE TABLE rule_results ("
                "id INTEGER PRIMARY KEY, "
-               "key STRING, "
+               "key_id INTEGER, "
                "value BLOB, "
                "built_at INTEGER, "
-               "computed_at INTEGER);"),
+               "computed_at INTEGER, "
+               "FOREIGN KEY(key_id) REFERENCES key_names(id));"),
           nullptr, nullptr, &cError);
       }
       if (result == SQLITE_OK) {
@@ -145,9 +153,10 @@ public:
         result = sqlite3_exec(
           db, ("CREATE TABLE rule_dependencies ("
                "rule_id INTEGER, "
-               "key STRING, "
-               "PRIMARY KEY (rule_id, key) "
-               "FOREIGN KEY(rule_id) REFERENCES rule_info(id)) "
+               "key_id INTEGER, "
+               "PRIMARY KEY (rule_id, key_id) "
+               "FOREIGN KEY(rule_id) REFERENCES rule_info(id) "
+               "FOREIGN KEY(key_id) REFERENCES key_names(id)) "
 #if SQLITE_VERSION_NUMBER >= 3008002
                "WITHOUT ROWID"
 #endif
@@ -160,7 +169,7 @@ public:
         // Create an index to be used for efficiently looking up rule
         // information from a key.
         result = sqlite3_exec(
-          db, "CREATE UNIQUE INDEX rule_results_idx ON rule_results (key);",
+               db, "CREATE UNIQUE INDEX rule_results_idx ON rule_results (key_id);",
           nullptr, nullptr, &cError);
       }
 
@@ -180,10 +189,25 @@ public:
 
     // Initialize prepared statements.
     result = sqlite3_prepare_v2(
+      db, findKeyIDForKeyStmtSQL,
+      -1, &findKeyIDForKeyStmt, nullptr);
+    assert(result == SQLITE_OK);
+
+    result = sqlite3_prepare_v2(
+      db, findKeyForKeyIDStmtSQL,
+      -1, &findKeyForKeyIDStmt, nullptr);
+    assert(result == SQLITE_OK);
+    
+    result = sqlite3_prepare_v2(
       db, findIDForKeyInRuleResultsStmtSQL,
       -1, &findIDForKeyInRuleResultsStmt, nullptr);
     assert(result == SQLITE_OK);
-
+    
+    result = sqlite3_prepare_v2(
+      db, insertIntoKeysStmtSQL,
+      -1, &insertIntoKeysStmt, nullptr);
+    assert(result == SQLITE_OK);
+    
     result = sqlite3_prepare_v2(
       db, insertIntoRuleResultsStmtSQL,
       -1, &insertIntoRuleResultsStmt, nullptr);
@@ -193,7 +217,12 @@ public:
       db, insertIntoRuleDependenciesStmtSQL,
       -1, &insertIntoRuleDependenciesStmt, nullptr);
     assert(result == SQLITE_OK);
-
+    
+    result = sqlite3_prepare_v2(
+      db, deleteFromKeysStmtSQL,
+      -1, &deleteFromKeysStmt, nullptr);
+    assert(result == SQLITE_OK);
+    
     result = sqlite3_prepare_v2(
       db, deleteFromRuleResultsStmtSQL,
       -1, &deleteFromRuleResultsStmt, nullptr);
@@ -203,7 +232,7 @@ public:
       db, deleteFromRuleDependenciesStmtSQL,
       -1, &deleteFromRuleDependenciesStmt, nullptr);
     assert(result == SQLITE_OK);
-
+    
     result = sqlite3_prepare_v2(
       db, findRuleResultStmtSQL,
       -1, &findRuleResultStmt, nullptr);
@@ -221,10 +250,14 @@ public:
     assert(db);
 
     // Destroy prepared statements.
+    sqlite3_finalize(findKeyForKeyIDStmt);
+    sqlite3_finalize(findKeyIDForKeyStmt);
     sqlite3_finalize(findRuleDependenciesStmt);
     sqlite3_finalize(findRuleResultStmt);
+    sqlite3_finalize(deleteFromKeysStmt);
     sqlite3_finalize(deleteFromRuleDependenciesStmt);
     sqlite3_finalize(deleteFromRuleResultsStmt);
+    sqlite3_finalize(insertIntoKeysStmt);
     sqlite3_finalize(insertIntoRuleDependenciesStmt);
     sqlite3_finalize(insertIntoRuleResultsStmt);
     sqlite3_finalize(findIDForKeyInRuleResultsStmt);
@@ -278,13 +311,17 @@ public:
     sqlite3_finalize(stmt);
   }
 
+  static constexpr const char *deleteFromKeysStmtSQL =
+  "DELETE FROM key_names WHERE key == ?;";
+  sqlite3_stmt* deleteFromKeysStmt = nullptr;
+  
   static constexpr const char *findRuleResultStmtSQL =
     "SELECT id, value, built_at, computed_at FROM rule_results "
-    "WHERE key == ? LIMIT 1;";
+    "WHERE key_id == ? LIMIT 1;";
   sqlite3_stmt* findRuleResultStmt = nullptr;
 
   static constexpr const char *findRuleDependenciesStmtSQL =
-    "SELECT key FROM rule_dependencies "
+    "SELECT key_id FROM rule_dependencies "
     "WHERE rule_id == ?;";
   sqlite3_stmt* findRuleDependenciesStmt = nullptr;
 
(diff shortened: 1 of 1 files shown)


lib/Core/SQLiteBuildDB.cpp
```diff
@@ -109,13 +109,13 @@ class SQLiteBuildDB : public BuildDB {
...
-	       "id INTEGER PRIMARY KEY, "
-	       "version INTEGER, "
-	       "client_version INTEGER, "
-	       "iteration INTEGER);"),
-	  nullptr, nullptr, &cError);
+               "id INTEGER PRIMARY KEY, "
+               "version INTEGER, "
+               "client_version INTEGER, "
+               "iteration INTEGER);"),
+          nullptr, nullptr, &cError);
```
ddunbar: Can you put the indentation into a separate patch? Actually I am fine if you just want to fix non-code changes like this on trunk.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -126,12 +126,20 @@ class SQLiteBuildDB : public BuildDB {
       }
       if (result == SQLITE_OK) {
         result = sqlite3_exec(
+          db, ("CREATE TABLE keys ("
```
ddunbar: I'd probably call this key_names

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: Yeah sure

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -203,7 +218,12 @@ class SQLiteBuildDB : public BuildDB {
       db, deleteFromRuleDependenciesStmtSQL,
       -1, &deleteFromRuleDependenciesStmt, nullptr);
     assert(result == SQLITE_OK);
-
+    
+    result = sqlite3_prepare_v2(
+      db, findKeyIDForKeyStmtSQL,
+      -1, &findKeyIDForKeyStmt, nullptr);
+    assert(result == SQLITE_OK);
+
```
ddunbar: Can you put this next to the init of `findKeyForKeyID`?

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -126,12 +126,20 @@ class SQLiteBuildDB : public BuildDB {
       }
       if (result == SQLITE_OK) {
         result = sqlite3_exec(
+          db, ("CREATE TABLE keys ("
+               "id INTEGER PRIMARY KEY, "
+               "key STRING UNIQUE);"),
+          nullptr, nullptr, &cError);
```
ddunbar: My knowledge of SQLite breaks down here... I hope, but am not 100% sure, that SQLite will realize that the index on keys.key is a covering index and it will never actually need the primary table. It feels like there should be a way to ensure that the data is only ever present in the index (the table itself is irrelevant), but I'm not sure if that is possible.

It might be worth verifying that SQLite isn't hitting the actual table here when doing key lookups, although I would need to play around with the query explainer to even know exactly what to look for.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -293,13 +335,19 @@ class SQLiteBuildDB : public BuildDB {
 
     // Fetch the basic rule information.
     int result;
+    uint64_t keyID;
+    
+    // If the key wasn't found, we are done.
+    if(!findKeyIDForKey(rule.key, &keyID)) {
```
ddunbar: Nit: `if (`

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -338,18 +386,16 @@ class SQLiteBuildDB : public BuildDB {
         abort();
       }
       assert(sqlite3_column_count(findRuleDependenciesStmt) == 1);
-      result_out->dependencies.push_back(
-          std::string(
-              (const char*)sqlite3_column_text(findRuleDependenciesStmt, 0),
-              sqlite3_column_bytes(findRuleDependenciesStmt, 0)));
+      uint64_t keyID = sqlite3_column_int64(findRuleDependenciesStmt, 0);
```
ddunbar: I would call this `dependencyID` just so it doesn't shadow the very similar but semantically different `keyID` in the containing scope.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -373,32 +419,111 @@ class SQLiteBuildDB : public BuildDB {
     "DELETE FROM rule_dependencies WHERE rule_id == ?;";
   sqlite3_stmt* deleteFromRuleDependenciesStmt = nullptr;
 
-  virtual void setRuleResult(const Rule& rule,
-                             const Result& ruleResult) override {
+  //Return the key for a given keyID, keyID should exist in database.
```
ddunbar: Nit: `// Return`

ddunbar: Also, actually this should be a doxygen-level comment (`/// Return`).

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -373,32 +419,111 @@ class SQLiteBuildDB : public BuildDB {
...
+      return std::string(
+                (const char*)sqlite3_column_text(findKeyForKeyIDStmt, 0),
+                sqlite3_column_bytes(findKeyForKeyIDStmt, 0));
+    }
+    abort();
+  }
+  
+  //Finds keyID for a key and puts result in keyID_out
+  //returns false if key not present
+  bool findKeyIDForKey(const KeyType& key, uint64_t* keyID_out) {
```
ddunbar: Same nit, `///` with a leading space.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -373,32 +419,111 @@ class SQLiteBuildDB : public BuildDB {
...
+    assert(result == SQLITE_OK);
+    result = sqlite3_bind_text(insertIntoKeysStmt, /*index=*/1,
+                               key.data(), key.size(),
+                               SQLITE_STATIC);
+    result = sqlite3_step(insertIntoKeysStmt);
+    if (result != SQLITE_DONE)
+      abort();
+    
+    uint64_t keyID;
+    auto findResult = findKeyIDForKey(key, &keyID);
```
ddunbar: I believe that this code can use `sqlite3_last_row_id()`, which should be substantially more efficient. We should try and find a way we can do the insert without duplicate lookups.

aciidb0mb3r: I considered using that first but according to docs it won't update if there is no insertion

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -373,32 +419,111 @@ class SQLiteBuildDB : public BuildDB {
...
+  }
+  
+  virtual void setRuleResult(const Rule& rule,
+                             const Result& ruleResult) override {
+    int result;
+
+    uint64_t keyID;
+    uint64_t ruleID = 0;
+    
+    //only bother to lookup ruleID if the key is present
```
ddunbar: Nit: complete sentences (leading caps and trailing punctuation, a la the [LLVM style guide](http://llvm.org/docs/CodingStandards.html))

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r:
```
** ^An [INSERT] that fails due to a constraint violation is not a
** successful [INSERT] and does not change the value returned by this
** routine.  ^Thus INSERT OR FAIL, INSERT OR IGNORE, INSERT OR ROLLBACK,
** and INSERT OR ABORT make no changes to the return value of this
** routine when their insertion fails.
```

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -373,32 +419,111 @@ class SQLiteBuildDB : public BuildDB {
...
     } else {
       abort();
     }
+    
+    return false;
+  }
+  
+  //Adds a key into keys table if not present
+  //returns the keyID of the key
+  uint64_t addKey(const KeyType& key) {
```
ddunbar: One overall general comment:

I think we should just have a single method for interacting with keys, which is to `getOrInsert(KeyType)` and it always returns an ID.

This means that we sometimes will end up creating a key even when `lookupRuleResult` will just realize the rule doesn't exist, but in practice the engine is just going to come along in a second and create that rule. This should simplify the patch and make only one thing that needs to be optimized.

aciidb0mb3r: sounds good to me

ddunbar:
Yes, this seems like the right approach to me... added a bunch of nit-level comments.

The major high level comment is needing to understand the performance impact. I would like to understand two things:
 - Performance impact on a full build (when the database is empty, and we are creating a bunch of data).
 - Performance impact on a null build (when we just need to query a bunch of the data).

My guess is that we need some kind of cache for the keys, but I'm unclear of how we should go about structuring that... I don't think I want us to cache *all* of them, that seems wasteful and also means we are just duplicating work from SQLite.

This is one area where I am a little worried that we may need to change the *interface* to the BuildDB in order to be able to get the space win and the performance we want. If we changed the way dependencies were stored at the BuildDB interface level, then the rest of the engine might only need the ID into the Rule, which would then mean we wouldn't have to do a ton of lookups just to resolve it. If that makes sense...

I need to mull on this some more, and would be interested if you have other thoughts about what we could do here...

aciidb0mb3r:
maybe we could try with a small LRU cache for keys and test the performance but I guess that won't be enough for a large build and managing cache could be more difficult than we want?

I like the idea of Rule having keyID instead of key that should definitely fix the performance issues but idk what the overall impact on the code would be, though definitely worth exploring.

Another thing I was thinking was to try with nested queries which maybe faster than two individual queries.

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: done

aciidb0mb3r:
@ddunbar I am done with nits, will look into performances

aciidb0mb3r:
I got these results so far :
Master: average: 0.332

Current patch: average: 0.360

If I use this query the average is actually increasing:
```
*findRuleDependenciesStmtSQL =
  "SELECT key FROM key_names WHERE id IN (SELECT key_id FROM rule_dependencies WHERE rule_id == ?);";
```
Query: average: 0.461

ddunbar:
My guess is you want a JOIN, not a nested query.

FWIW, I often prototype this by just using `sqlite3` on the command line to find the best query, then implement it in code.

ddunbar:
@swift-ci please test

aciidb0mb3r:
Updated to join, now:
patch -> average: 0.285
master -> average: 0.335

:D

ddunbar:
Nice! Is it worth optimizing the initial rule_result query to also use a join so only one query is needed?

Also, does this pass all tests?

aciidb0mb3r:
current patch is average: 0.243
with 
```
static constexpr const char *findRuleResultStmtSQL =
    "SELECT rule_results.id, value, built_at, computed_at FROM rule_results "
    "INNER JOIN key_names ON key_names.id = rule_results.key_id WHERE key == ?;";
```
it becomes average: 0.268
So I guess not worth it, however I removed the unused `keyForKeyID` method

there are 3 lit tests failing, looking into that.

ddunbar:
I wonder why it got slower, seems like it should be strictly less work...

ddunbar:
@swift-ci please test

aciidb0mb3r:
@ddunbar fixed the tests

aciidb0mb3r:
@swift-ci please test

ddunbar:
@swift-ci please test

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -193,7 +212,12 @@ class SQLiteBuildDB : public BuildDB {
       db, insertIntoRuleDependenciesStmtSQL,
       -1, &insertIntoRuleDependenciesStmt, nullptr);
     assert(result == SQLITE_OK);
-
+    
+    result = sqlite3_prepare_v2(
+      db, deleteFromKeysStmtSQL,
+      -1, &deleteFromKeysStmt, nullptr);
+    assert(result == SQLITE_OK);
+
```
ddunbar: This is unused in this patch... I am ok with landing this without removing keys, we should file a follow on to track that, but in practice I don't see it being a huge problem in the short term.

ddunbar:
LGTM!

lib/Core/SQLiteBuildDB.cpp
aciidb0mb3r: Done, https://bugs.swift.org/browse/SR-1244

ddunbar:
Looks like something is wrong here, my performance monitoring bots noticed a major regression in the initial build time (null build time did decrease, as expected, although unfortunately not by the same percentage).

This regression is pretty huge, unfortunately:
screen shot 2016-04-15 at 8 43 32 am

ddunbar:
@aciidb0mb3r can you take a look and see if it is something simple you think can be resolved in short order? If not we should probably revert this while the investigation takes place.

ddunbar:
Oh, also the static analyzer flagged this:
```
Assertion: Value stored to 'result' is never read
File: llbuild/lib/Core/SQLiteBuildDB.cpp:446

Introduced by Ankit Agarwal (b3b71cd)
```

PR #17, opened 2016-04-13
Update Swift bindings for llbuild to 2.2

beltex:
#### What's in this pull request?

Updates the Swift bindings for `llbuild` to version `2.2`.

#### Why merge this pull request?

To address the following deprecation warnings:

* [_Removing `var` from Function Parameters_](https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters.md)
* [_Remove the `++` and `--` operators_](https://github.com/apple/swift-evolution/blob/master/proposals/0004-remove-pre-post-inc-decrement.md)

#### What are the downsides to merging this pull request?

None that I can think of!

#### Resolved bug number: (None)

### DIFF
--- a/bindings/swift/llbuild.swift
+++ b/bindings/swift/llbuild.swift
@@ -26,9 +26,9 @@ private func stringFromData(data: llb_data_t) -> String {
   return String([UInt8](UnsafeBufferPointer(start: data.data, count: Int(data.length))))
 }
 
-private func stringFromUInt8Array(var data: [UInt8]) -> String {
+private func stringFromUInt8Array(data: [UInt8]) -> String {
   // Convert as a UTF8 string, if possible.
-  let tmp = NSData(bytes: &data, length: Int(data.count))
+  let tmp = NSData(bytes: data, length: data.count)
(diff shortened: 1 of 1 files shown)


ddunbar:
Thanks!

Out of curiosity, how did you test these (given that they aren't yet integrated into the build)?

beltex:
> Thanks!

np! 😄 

> Out of curiosity, how did you test these (given that they aren't yet integrated into the build)?

I added an additional scheme to the Xcode project for the existing `llbuild-framework` target to compile it. Then added a new **Command Line Tool** template target and scheme for `bindings` which included this source file and linked with the `llbuild.framework`. Everything compiled successfully. In addition ran a basic sanity test:

```swift
let key = Key("myKey")
let value = Value("myValue")

// Which invokes `stringFromUInt8Array()`
print(key.toString())
print(value.toString())
```

Are there some other tests that should be run? Whats the best way to integrate them into the build (the bindings)? As I imagine what I did is not ideal.

ddunbar:
It sounds reasonable to me. I would like to get proper integration of the bindings both in the Xcode project and the CMake build...

PR #21, opened 2016-04-17
Parseable swiftc output

aciidb0mb3r:
This parses the `began` and `finished` json output provided by swiftc and displays the status when compiling swift modules.

Sample output :

```
Compile Swift Module 'swifty' (5 sources)
Compile /Users/aciid/mycode/swiftpmProjects/swifty/Sources/Foo1.swift
Compile /Users/aciid/mycode/swiftpmProjects/swifty/Sources/Foo3.swift
Compiled /Users/aciid/mycode/swiftpmProjects/swifty/Sources/Foo1.swift
Compiled /Users/aciid/mycode/swiftpmProjects/swifty/Sources/Foo3.swift
Linking .build/debug/swifty
Program ended with exit code: 0
```

### DIFF
--- a/include/llbuild/BuildSystem/BuildExecutionQueue.h
+++ b/include/llbuild/BuildSystem/BuildExecutionQueue.h
@@ -114,13 +114,16 @@ public:
   virtual bool
   executeProcess(QueueJobContext* context,
                  ArrayRef<StringRef> commandLine,
-                 ArrayRef<std::pair<StringRef, StringRef>> environment) = 0;
+                 ArrayRef<std::pair<StringRef, StringRef>> environment,
+                 std::function<void(StringRef)> partialOutputCallback = nullptr) = 0;
 
   /// @}
 
   /// Execute the given command, using an inherited environment.
   bool executeProcess(QueueJobContext* context,
-                      ArrayRef<StringRef> commandLine);
+                      ArrayRef<StringRef> commandLine,
+                      std::function<void(StringRef)> partialOutputCallback = nullptr
+                      );
 
   /// Execute the given command using "/bin/sh".
   ///
--- a/lib/BuildSystem/LaneBasedExecutionQueue.cpp
+++ b/lib/BuildSystem/LaneBasedExecutionQueue.cpp
@@ -124,7 +124,8 @@ public:
   executeProcess(QueueJobContext* opaqueContext,
                  ArrayRef<StringRef> commandLine,
                  ArrayRef<std::pair<StringRef,
-                                    StringRef>> environment) override {
+                                    StringRef>> environment,
+                 std::function<void(StringRef)> partialOutputCallback) override {
     // Assign a process handle, which just needs to be unique for as long as we
     // are communicating with the delegate.
     struct BuildExecutionQueueDelegate::ProcessHandle handle;
@@ -296,7 +297,11 @@ public:
         if (numBytes == 0)
           break;
 
-        outputData.insert(outputData.end(), &buf[0], &buf[numBytes]);
+        if (partialOutputCallback) {
+          partialOutputCallback(std::string(buf, numBytes));
+        } else {
+          outputData.insert(outputData.end(), &buf[0], &buf[numBytes]);
+        }
       }
 
       // Close the read end of the pipe.
--- a/lib/BuildSystem/SwiftTools.cpp
+++ b/lib/BuildSystem/SwiftTools.cpp
@@ -174,7 +174,7 @@ class SwiftCompilerShellCommand : public ExternalCommand {
 
   /// Whether the sources are part of a library or not.
   bool isLibrary = false;
-  
+
   virtual uint64_t getSignature() override {
     uint64_t result = ExternalCommand::getSignature();
     result ^= basic::hashString(executable);
@@ -231,6 +231,8 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     for (const auto& arg: otherArgs) {
       result.push_back(arg);
     }
+
+    result.push_back("-parseable-output");
   }
   
 public:
@@ -523,7 +525,10 @@ public:
       return false;
 
     // Execute the command.
-    if (!bsci.getExecutionQueue().executeProcess(context, commandLine)) {
+
+    auto handler = std::bind(&SwiftCompilerShellCommand::partialOutputAvailable, this, std::placeholders::_1);
+
+    if (!bsci.getExecutionQueue().executeProcess(context, commandLine, handler)) {
       // If the command failed, there is no need to gather dependencies.
       return false;
     }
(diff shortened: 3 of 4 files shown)


include/llbuild/BuildSystem/BuildExecutionQueue.h
```diff
@@ -114,13 +114,16 @@ class BuildExecutionQueue {
   virtual bool
   executeProcess(QueueJobContext* context,
                  ArrayRef<StringRef> commandLine,
-                 ArrayRef<std::pair<StringRef, StringRef>> environment) = 0;
+                 ArrayRef<std::pair<StringRef, StringRef>> environment,
+                 std::function<void(StringRef)> partialOutputCallback = nullptr) = 0;
```
ddunbar: Can you break this change out into its own patch, with tests? I would probably call this `hasPartialOutput()`, which is a little more consistent with our existing callback naming conventions.

Also, this change should be added to the documentation comments here.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -296,7 +297,11 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
         if (numBytes == 0)
           break;
 
-        outputData.insert(outputData.end(), &buf[0], &buf[numBytes]);
+        if (partialOutputCallback) {
+          partialOutputCallback(std::string(buf, numBytes));
+        } else {
+          outputData.insert(outputData.end(), &buf[0], &buf[numBytes]);
+        }
```
ddunbar: This makes for a weird API, I wonder if we should clean up the overall API here instead of making this behavior conditional.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -523,7 +525,10 @@ class SwiftCompilerShellCommand : public ExternalCommand {
       return false;
 
     // Execute the command.
-    if (!bsci.getExecutionQueue().executeProcess(context, commandLine)) {
+
+    auto handler = std::bind(&SwiftCompilerShellCommand::partialOutputAvailable, this, std::placeholders::_1);
```
ddunbar: Can you just use an inline lambda here to forward? I find it more readable than `std::bind()`.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -536,6 +541,50 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+      // See if we have enough output to parse.
+      if (previousOutput.size() >= numBytes) {
+        StringRef legibleJSON = previousOutput.substr(0, numBytes);
+
+        // Do something with JSON.
+        std::string sep = "\n--------------------\n";
+        fwrite(sep.data(), sep.size(), 1, stdout);
+        fflush(stdout);
+        fwrite(legibleJSON.data(), legibleJSON.size(), 1, stdout);
+        fflush(stdout);
```
ddunbar: When building up incremental patches, I think each patch should be "committable" (i.e., not have behavior which would be problematic if it was landed without the subsequent patches). So, I would factor this patch (on top of first patch to add `hasPartialOutput()` in such a way that it just adds the facility of adding a per-JSON message handler (if enabled).

As you have done here, I would probably test it by just printing the data, but I wouldn't actually commit that. If it was complicated enough I felt like it deserved its own tests then I would design it to have a conditional logging feature the tests could use.

include/llbuild/BuildSystem/BuildExecutionQueue.h
ddunbar: Another thing we need to handle here is splitting of the stdout/stderr streams; currently they are merged, but with this change we should probably split them and supply the output for each independently to the client. This will be annoying because we will need to move to using `select()` (or something like it) to pull the data off the output pipes. However, it is important we do this so that `stderr` output from the compiler isn't lost.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -536,6 +541,50 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+        // Don't even have the first line yet.
+        if (newLinePos == std::string::npos) {
+          return;
+        }
+
+        StringRef numBytesString = previousOutput.substr(0, newLinePos);
+        if (StringRef(numBytesString).getAsInteger(10, numBytes)) {
+          numBytes = 0;
+          // Couldn't get number of bytes.
+          // Probably due to a compiler crash or bad output?
```
ddunbar: We should probably do something else here, even if it just a `// FIXME: Need logging infrastructure.` to get a supported API for logging this kind of "unexpected, but possible" error.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -536,6 +541,50 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+          return;
+        }
+
+        // Remove the message length string.
+        previousOutput = previousOutput.substr(newLinePos+1, previousOutput.size());
```
ddunbar: This is an inefficient way of processing this data, there is no need to actually move the bytes within the buffer except when we need to concatenate it for waiting for the next chunk. `StringRef` is designed to be really nice for this case, you can use it to "slide" over the underlying buffer while still making use of functions like `find` to work with the data.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -175,6 +179,10 @@ class SwiftCompilerShellCommand : public ExternalCommand {
   /// Whether the sources are part of a library or not.
   bool isLibrary = false;
 
+  /// Contains the "began" messages produced by swiftc's parseable-output
+  /// mapped to their pid.
+  std::unordered_map<int32_t, SwiftCompilerOutputMessage> outputmap;
```
ddunbar: Is there a need to store the entire message? What do we use the rest of it for?

Also, camelCase please.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -624,8 +630,322 @@ class SwiftCompilerTool : public Tool {
     return nullptr;
   }
 };
+
+class SwiftCompilerOutputMessage {
```
ddunbar: After looking at this some more, I think it should probably move to a parser + delegate model. The parser would take in the incoming stream of bytes, and it would take a delegate with a bunch of methods like:
```
virtual void actOnBegin( ... )
virtual void actOnFinished(...)
```

The reason this makes more sense if that the clients of this API are probably going to have very specialized logic for each case. It doesn't make sense to pack up all the data into a uniform structure just to turn around and have the clients unpack it.

This would also have another major advantage of being unit testable, which I would recommend actually doing. If you want an example, look at 600d1262d019003c4dbef8214224556145fdb796.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -624,8 +630,322 @@ class SwiftCompilerTool : public Tool {
...
+          return false;
+        }
+        message_out.setPid(pid);
+        ++it;
+        
+        return true;
+        
+      }
+      case Kind::Finished:
+      {
```
ddunbar: llbuild uses braces on the previous line.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -624,8 +630,322 @@ class SwiftCompilerTool : public Tool {
...
+        //get pid
```
ddunbar: Complete sentences + leading space plz!

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -624,8 +630,322 @@ class SwiftCompilerTool : public Tool {
...
+  }
+
+  if (message.getKind() == SwiftCompilerOutputMessage::Kind::Began && message.getName() == "compile") {
+    llvm::outs() << "Compile " << message.getInputs().front() << "\n";
+    outputmap.insert(std::make_pair(message.getPid(), message));
+  } else if (message.getKind() == SwiftCompilerOutputMessage::Kind::Finished) {
+    if (message.getExitStatus() == 0) {
+      auto it = outputmap.find(message.getPid());
+      if (it != outputmap.end())
+        llvm::outs() << "Compiled " << it->second.getInputs().front()  << "\n";
```
ddunbar: I wouldn't bother having two messages for start vs. stop, I think it is too noisy. I have a plan for how we can eventually show the duration of running commands nicely for all tools.

ddunbar:
Basic approach looks good to me, but I would split into a more incremental (and tested) set of patches.

ddunbar:
I'm closing this out due to age, we can resurrect it if either of us picks it up again.

PR #28, opened 2016-05-04
[BuildSystem] Teach swiftc command about -whole-module-optimization. …

aciidb0mb3r:
…When enabled swift compiler produces one object file and dependency file for the entire module instead of for each of the sources

### DIFF
--- a/lib/BuildSystem/SwiftTools.cpp
+++ b/lib/BuildSystem/SwiftTools.cpp
@@ -175,6 +175,12 @@ class SwiftCompilerShellCommand : public ExternalCommand {
   /// Whether the sources are part of a library or not.
   bool isLibrary = false;
   
+  /// Whether to enable -whole-module-optimization.
+  bool enableWholeModuleOptimization = false;
+
+  /// Enables multi-threading with the thread count if > 0.
+  std::string numThreads = "0";
+
   virtual uint64_t getSignature() override {
     uint64_t result = ExternalCommand::getSignature();
     result ^= basic::hashString(executable);
@@ -220,6 +226,11 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     if (isLibrary) {
       result.push_back("-parse-as-library");
     }
+    if (enableWholeModuleOptimization) {
+      result.push_back("-whole-module-optimization");
+    }
+    result.push_back("-num-threads");
+    result.push_back(numThreads);
     result.push_back("-c");
     for (const auto& source: sourcesList) {
       result.push_back(source);
@@ -289,12 +300,22 @@ public:
     } else if (name == "temps-path") {
       tempsPath = value;
     } else if (name == "is-library") {
-      if (value != "true" && value != "false") {
-        ctx.error("invalid value: '" + value + "' for attribute '" +
-                  name + "'");
+      if (!configureBool(ctx, isLibrary, name, value))
+        return false;
+    } else if (name == "enable-whole-module-optimization") {
+      if (!configureBool(ctx, enableWholeModuleOptimization, name, value))
+        return false;
+    } else if (name == "num-threads") {
+      int numThreadsInt = 0;
+      if (value.getAsInteger(10, numThreadsInt)) {
+        ctx.error("'" + name + "' should be an int.");
         return false;
       }
-      isLibrary = value == "true";
+      if (numThreadsInt < 0) {
+        ctx.error("'" + name + "' should be greater than or equal to zero.");
+        return false;
+      }
+      numThreads = value;
     } else if (name == "other-args") {
       SmallVector<StringRef, 32> args;
       StringRef(value).split(args, " ", /*MaxSplit=*/-1,
@@ -306,7 +327,19 @@ public:
 
     return true;
   }
-  
+
+  // Extracts and stores the bool value of an attribute inside "to" variable.
+  // Returns true on success and false on error.
+  bool configureBool(const ConfigureContext& ctx, bool& to, StringRef name, StringRef value) {
+    if (value != "true" && value != "false") {
+      ctx.error("invalid value: '" + value + "' for attribute '" +
+          name + "'");
+      return false;
+    }
+    to = value == "true";
+    return true;
+  }
+
   virtual bool configureAttribute(const ConfigureContext& ctx, StringRef name,
                                   ArrayRef<StringRef> values) override {
     if (name == "sources") {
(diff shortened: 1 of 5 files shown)


lib/BuildSystem/SwiftTools.cpp
```diff
@@ -288,13 +294,17 @@ class SwiftCompilerShellCommand : public ExternalCommand {
       importPaths = std::vector<std::string>(imports.begin(), imports.end());
     } else if (name == "temps-path") {
       tempsPath = value;
-    } else if (name == "is-library") {
+    } else if (name == "is-library" || name == "enable-wmo") {
```
ddunbar: I'd prefer to use the full name, even though it is verbose. "wmo" is too opaque

Also, please update the docs when you add new keys!

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -288,13 +294,17 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
         ctx.error("invalid value: '" + value + "' for attribute '" +
                   name + "'");
         return false;
       }
-      isLibrary = value == "true";
+      bool boolValue = value == "true";
+      if (name == "is-library")
+        isLibrary = boolValue;
+      else if (name == "enable-wmo")
+        enableWholeModuleOptimization = boolValue;
```
ddunbar: I would rather factor out a helper function for bool reading than lump these into one `else if`, I find that less readable.

lib/BuildSystem/SwiftTools.cpp
aciidb0mb3r: should this be enable-whole-module-optimization or enable-wholeModuleOptimization

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -348,37 +358,47 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     }
 
     os << "{\n";
+
+    // Write out the entries for each source file if WMO is not enabled.
```
ddunbar: I think there are cases where WMO still uses the per-file entries (with parallel WMO). I unfortunately forget the details, but there are multiple modes, one where WMO runs as a single process, and one where WMO runs with one thread/process to do the codegen of the .o files (I think there is an explicit option for it).

Xcode's default these days is to use the one which does the codegen in parallel, so we should match that. In that mode, I believe the individual .o files are still used.

aciidb0mb3r: Ah cool. Let me see if I can find that out

lib/BuildSystem/SwiftTools.cpp
aciidb0mb3r: done

ddunbar:
We should also have a test case for this.

You can teach the "fake" swiftc how to validate these options a tad. That doesn't test against the real thing, but it does protect against modifications to the code having unintended consequences.

aciidb0mb3r:
ah cool

lib/BuildSystem/SwiftTools.cpp
aciidb0mb3r: done

lib/BuildSystem/SwiftTools.cpp
aciidb0mb3r: Looks like the parallelism is decided by the `-num-threads` flag

aciidb0mb3r: I'll just remove the if for now, not sure what value to pass to it.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -220,6 +223,11 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     if (isLibrary) {
       result.push_back("-parse-as-library");
     }
+    if (enableWholeModuleOptimization) {
+      result.push_back("-whole-module-optimization");
+      result.push_back("-num-threads");
+      result.push_back("1");
```
aciidb0mb3r: @ddunbar how/what to set a value here.

ddunbar: We don't have a good answer currently, for now I think the best thing is to make it a separate option and handle both cases of parallel or non-parallel WMO. Then clients can pick if they want to hard code a threads value even if llbuild itself isn't scheduling it "properly"

aciidb0mb3r: Ok. I dont think its going to matter if output file has entries for individual objects when non-parallel WMO is used. Though I'll double check to be sure.

ddunbar: I think it is ok to leave them there, if it is convenient, but I think we need to be careful to only link the correct `.o` files (for which I think there will just be one).

aciidb0mb3r: The object file for entire module was present in both modes iirc.

ddunbar: That would surprise me, I would make sure it wasn't just left over from before.

aciidb0mb3r: Just checked, you're right. Lots of ifs for swift-build then \._./

drewcrawford: atbuild would be interested in setting a value for num-threads, so that's a datapoint.

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -353,6 +381,15 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     SmallString<16> masterDepsPath;
     llvm::sys::path::append(masterDepsPath, tempsPath, "master.swiftdeps");
     os << "  \"\": {\n";
+    if (enableWholeModuleOptimization) {
```
ddunbar: Don't we only need this if WMO && !parallel-WMO?

lib/BuildSystem/SwiftTools.cpp
```diff
@@ -361,18 +398,20 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
       llvm::sys::path::append(partialModulePath, tempsPath,
                               sourceStem + "~partial.swiftmodule");
       SmallString<16> swiftDepsPath;
       llvm::sys::path::append(swiftDepsPath, tempsPath,
                               sourceStem + ".swiftdeps");
-      depsFiles_out.push_back(depsPath.str());
       
       os << "  \"" << source << "\": {\n";
-      os << "    \"dependencies\": \"" << depsPath << "\",\n";
+      if (!enableWholeModuleOptimization) {
```
ddunbar: Don't we need this if !WMO || parallel-WMO?

lib/BuildSystem/SwiftTools.cpp
aciidb0mb3r: we need depsPath in both cases
object is not needed for !parallel-WMO == true but it can just stay there

lib/BuildSystem/SwiftTools.cpp
aciidb0mb3r: in both cases only one .d is generated for entire module

aciidb0mb3r:
Added test for both WMO and WMO parallel

aciidb0mb3r:
@swift-ci please test

aciidb0mb3r:
@swift-ci please test

ddunbar:

aciidb0mb3r:

aciidb0mb3r:
@swift-ci please test

jrose-apple:
I think it's not worth investing in this; `-whole-module-optimization` *plus* the `-num-threads` option goes back to producing several .o files. cc @eeckstein

aciidb0mb3r:
@ddunbar CI passed but @jrose-apple says it's not worth doing?

ddunbar:
I think @jrose-apple was just referring to WMO with !num-threads, as it is "deprecated". I still think we should make it behave correctly as long as the compiler supports it.

PR #29, opened 2016-05-06
[Basic] Add utility method to create a shell quoted strings

aciidb0mb3r:
Fixes https://bugs.swift.org/browse/SR-1174

### DIFF
--- a/include/llbuild/Basic/ShellUtility.h
+++ b/include/llbuild/Basic/ShellUtility.h
@@ -1,0 +1,43 @@
+//===- ShellUtility.h -------------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLBUILD_BASIC_SHELL_UTILITY_H
+#define LLBUILD_BASIC_SHELL_UTILITY_H
+
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/raw_ostream.h"
+
+using namespace llvm;
+
+namespace llbuild {
+namespace basic {
+
+/// Appends a shell escaped string to an output stream.
+/// For e.g. hello -> hello, hello$world -> 'hello$world', input A -> 'input A'
+///
+/// \param os Reference of the output stream to append to.
+///
+/// \param string The string to be escaped and appended.
+/// 
+void appendShellEscapedString(llvm::raw_ostream& os, StringRef string);
+
+/// Creates and returns a shell escaped string of the input.
+///
+/// \param string The string to be escaped.
+///
+/// \returns escaped string.
+std::string shellEscaped(StringRef string);
+
+}
+}
+
+#endif
--- a/lib/Basic/ShellUtility.cpp
+++ b/lib/Basic/ShellUtility.cpp
@@ -1,0 +1,62 @@
+//===-- ShellUtility.cpp --------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#include "llbuild/Basic/ShellUtility.h"
+#include "llvm/ADT/SmallString.h"
+
+namespace llbuild {
+namespace basic {
+
+void appendShellEscapedString(llvm::raw_ostream& os, StringRef string) {
+
+  static const std::string whitelist = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-_/:@#%+=.,";
+  auto pos = string.find_first_not_of(whitelist);
+
+  // We don't need any escaping just append the string and return.
+  if (pos == std::string::npos) {
+    os << string;
+    return;
+  }
+
+  // We only need to escape the single quote, if it isn't present we can
+  // escape using single quotes.
+  auto singleQuotePos = string.find_first_of("'", pos);
+  if (singleQuotePos == std::string::npos) {
+    os << "'";
+    os << string;
+    os << "'";
+    return;
+  }
+
+  // Otherwise iterate and escape all the single quotes.
+  os << "'";
+  os << string.slice(0, singleQuotePos);
+  for (auto idx = singleQuotePos; idx < string.size(); idx++) {
+    if (string[idx] == '\'') {
+      os << "'\\''";
+    } else {
+      os << string[idx];
+    }
+  }
+  os << "'";
+}
+
+std::string shellEscaped(StringRef string) {
+  SmallString<16> out;
+  llvm::raw_svector_ostream os(out);
+  appendShellEscapedString(os, string);
+  os.flush();
+  return out.str();
+}
+
+}
+}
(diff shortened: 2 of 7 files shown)


include/llbuild/Basic/Utility.h
```diff
@@ -0,0 +1,35 @@
...
+namespace llbuild {
+namespace basic {
+
+/// Escapes control characters in a string to make them displayable on terminal.
+/// For e.g. hello\nworld -> hello\\nworld, input A -> input\ A
+///
+/// \param argument Reference of string to be escaped.
+///
+/// \returns Escaped string.
+std::string shellEscaped(StringRef argument);
```
ddunbar: We should make this render into a `SmallVector<char>` instead of returning an `std::string`. This lets the client avoid heap allocation if desired, and is a common pattern in LLVM/Clang/Swift. `FileSystem.h` has examples of the pattern I am talking about.

include/llbuild/Basic/Utility.h
```diff
@@ -0,0 +1,35 @@
...
+#include "llvm/ADT/StringRef.h"
+#include <string>
+
+using namespace llvm;
+
+namespace llbuild {
+namespace basic {
+
+/// Escapes control characters in a string to make them displayable on terminal.
+/// For e.g. hello\nworld -> hello\\nworld, input A -> input\ A
```
ddunbar: To be a little more explicit, this escapes the string for parsing by the shell.

Note that the `\\n` example isn't actually correct, most shells don't actually honor those kind of escape sequences. Using spaces or characters like `$` as the examples is probably better.

include/llbuild/Basic/Utility.h
```diff
@@ -0,0 +1,35 @@
+//===- Utility.h ------------------------------------------------*- C++ -*-===//
```
ddunbar: Rather than making this `Utility`, which is a little too generic for my tastes, I suggest using `ShellUtility`.

lib/Basic/Utility.cpp
```diff
@@ -0,0 +1,40 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "llbuild/Basic/Utility.h"
+#include <set>
+#include <unordered_map>
+
+namespace llbuild {
+namespace basic {
+
+std::string shellEscaped(StringRef argument) {
```
ddunbar: I recommend a couple changes here... they don't have to be done in one PR but it might make the most sense:

1. I would recommend using whitelists not blacklists for characters. So optimize for the common case where the entire string consists only of characters which we know don't need to be escaped, basically `[-a-zA-Z0-9_/:]`(maybe a couple other common ones I forgot). This is the "fast path".

2. If that test fails, I would recommend by probably starting out returning a single-character quoted string. The reason why is that single-character quoted strings have fairly strict shell flexing rules, and so within such strings you only need to quote a very small number of characters (e.g., `'`) -- you can determine if you need any escapes at all during the first fast-path scan.

3. If you only need single quotes, but no escapes, you can just return `' + string + '`, which is the secondary-fast path.

4. If that fails, you do the slow thing of going over the string again character by character and quoting if necessary.

include/llbuild/Basic/Utility.h
ddunbar: It would be good to define two APIs:

1. Append the given string to a `raw_ostream` honor shell escaping rules. This is the common case clients will want, and avoids an unnecessary copy of the string data.

2. The API you have, which can be a thin wrapper around #1.

tests/BuildSystem/Build/changed-commands.llbuild
```diff
@@ -18,8 +18,8 @@
 
 ##### VERSION-BEGIN-1 #####
 
-# CHECK-VERSION-1: echo output-1 > output-1
-# CHECK-VERSION-1: cat output-1 > output
+# CHECK-VERSION-1: echo\ output-1\ \>\ output-1
+# CHECK-VERSION-1: cat\ output-1\ \>\ output
```
ddunbar: Why would these be escaped? That seems wrong the user should be able to copy paste them into terminal.

ddunbar:
Awesome, thank you for working on this!

tests/BuildSystem/Build/changed-commands.llbuild
aciidb0mb3r: the (escaped) output is `/bin/sh -c cat\ output-1\ \>\ output`

ddunbar: ah.. it ends up being unfortunate for the readability of our test cases... another reason to use single-quoted escaped style instead of inline `\\` escapes

aciidb0mb3r:
@ddunbar Implemented the comments.
@swift-ci please test

aciidb0mb3r:
@swift-ci please test

aciidb0mb3r:
@swift-ci please test

aciidb0mb3r:
@swift-ci please test osx

aciidb0mb3r:
@swift-ci please test darwin

aciidb0mb3r:
@swift-ci please test

aciidb0mb3r:
Some unexpected failure in swift stdlib set

aciidb0mb3r:
@swift-ci please test

include/llbuild/Basic/ShellUtility.h
```diff
@@ -0,0 +1,43 @@
...
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLBUILD_BASIC_SHELL_UTILITY_H
+#define LLBUILD_BASIC_SHELL_UTILITY_H
```
ddunbar: This should be `..SHELLUTILITY...` just to exactly match the file name, but it doesn't really matter and can be fixed up directly on master.

lib/Basic/ShellUtility.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "llbuild/Basic/ShellUtility.h"
+#include "llvm/ADT/SmallString.h"
+
+namespace llbuild {
+namespace basic {
+
+void appendShellEscapedString(llvm::raw_ostream& os, StringRef string) {
+
+  static const std::string whitelist = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-_/:@#%+=.,";
+  auto pos = string.find_first_not_of(whitelist);
```
ddunbar: I would prefer to use `StringRef::find_first_not_of` here.

unittests/Basic/ShellUtilityTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+  output = shellEscaped("input A B");
+  EXPECT_EQ(output, "'input A B'");
+
+  // Double Quote.
+  output = shellEscaped("input\"A");
+  EXPECT_EQ(output, "'input\"A'");
+
+  // Single Quote.
+  output = shellEscaped("input'A");
+  EXPECT_EQ(output, "'input'\\''A'");
```
ddunbar: This is the wrong answer. :)

lib/Basic/ShellUtility.cpp
```diff
@@ -0,0 +1,62 @@
...
+    os << "'";
+    return;
+  }
+
+  // Otherwise iterate and escape all the single quotes.
+  os << "'";
+  os << string.slice(0, singleQuotePos);
+  for (auto idx = singleQuotePos; idx < string.size(); idx++) {
+    if (string[idx] == '\'') {
+      os << "'\\''";
```
ddunbar: This should be `\\'` right?

lib/Basic/ShellUtility.cpp
aciidb0mb3r: string is a StringRef, or do you mean something else?

lib/Basic/ShellUtility.cpp
aciidb0mb3r: I can change it but IMO this is fine, on shell it'll read as `'some'\''arg'`...it is kind of odd to read but I think a single quote would have very less chance of being present ? There is one more way to escape single quotes but that reads even more ugly...

ddunbar: Well, with double quotes it would just be `"foo'bar"` or `"foo\"bar"` which I find easier to read. Double quotes are also a lot more common so people may be more familiar with the rules.

I agree it isn't ultimately that big of a deal though, let's take this and we can always change it later.

PR #30, opened 2016-05-15
[Core] Remove escaped spaces if present in MakefileDepsParser

aciidb0mb3r:
Fixes https://bugs.swift.org/browse/SR-1468

I am not too sure about this fix

### DIFF
--- a/include/llbuild/Core/MakefileDepsParser.h
+++ b/include/llbuild/Core/MakefileDepsParser.h
@@ -13,7 +13,10 @@
 #ifndef LLBUILD_CORE_MAKEFILEDEPSPARSER_H
 #define LLBUILD_CORE_MAKEFILEDEPSPARSER_H
 
+#include "llvm/ADT/SmallString.h"
 #include <cstdint>
+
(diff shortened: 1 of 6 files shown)


ddunbar:
This was a bad API design on my part... it matches a pattern we use in Clang but it isn't worth the complexity here. What we should do is return the unescaped word along with `actOnRuleStart` and `actOnRuleDependency`.

A good start to that would be just to have a temporary `SmallString` in `parse()`, and teach `lexWord` to always accumulate the unescaped string, and then pass it to the callbacks.

We should still pass the raw pointers too, because they can be used for diagnostics by clients (since they are guaranteed to be pointers into the original source buffer, and can be used to derived line and column info).

ddunbar:
Any update here?

aciidb0mb3r:
@ddunbar updated

ddunbar:
Landed a version of this patch here: 26206baf938d25fbb7b1778991fea745c1a530cb, which handles more escape characters correctly and uses StringRef at the API interface.

PR #33, opened 2016-07-11
Update for SE-0025 ('fileprivate' and 'private')

jrose-apple:
[SE-0025](https://github.com/apple/swift-evolution/blob/master/proposals/0025-scoped-access-level.md) makes `private` restrict access to a lexical scope; the Swift 2 behavior is now named `fileprivate`. At the time of this pull request, the compiler treats both `private` and `fileprivate` as having `fileprivate` semantics, but that will soon change.

### DIFF
--- a/products/swift-bindings/llbuild.swift
+++ b/products/swift-bindings/llbuild.swift
@@ -68,12 +68,12 @@ public struct Key: CustomStringConvertible, Equatable, Hashable {
   }
 
   /// Create a Key object from an llb_data_t.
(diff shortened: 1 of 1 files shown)


ddunbar:
Thanks! These bindings aren't currently built in CI, so I'm just merging this.

aciidb0mb3r:
Actually last I checked the bindings were not building at all because of the changes in compiler...

jrose-apple:
Yeah, I wasn't able to test this properly. I did it by hand to be one less thing you all need to fix.

PR #34, opened 2016-07-28
UnsafePointer<Void> -> UnsafeRawPointer

atrick:
Update for SE-0107 RawPointer.

This was a straightforward search/replace that I forgot to pull in with the corresponding change on swift master.


tkremenek:
@swift-ci smoke test

PR #35, opened 2016-09-16
Fixed for Cygwin

tinysun212:
Cygwin doesn't have SIGUNUSED.

### DIFF
--- a/lib/BuildSystem/LaneBasedExecutionQueue.cpp
+++ b/lib/BuildSystem/LaneBasedExecutionQueue.cpp
@@ -159,10 +159,12 @@ public:
     }
     posix_spawnattr_setsigdefault(&attributes, &mostSignals);
 #else
-    sigset_t allSignals;
-    sigfillset(&allSignals);
-    posix_spawnattr_setsigdefault(&attributes, &allSignals);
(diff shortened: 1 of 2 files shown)


ddunbar:
The SIGUNUSED comment doesn't seem relevant, since this was presumably trying to reset all signals and I guess that fails there too?

Does Cygwin really allow all signals to be set except SIGSTOP & SIGKILL?

tinysun212:
Cygwin header doesn't define the SIGUNUSED, so the `sigaddset()` loop for Linux couldn't be used.

When I tried to reset all signals as like the case for `#else`, the `posix_spawn()`  didn't work.
After adding `sigdelset()` for the signals SIGSTOP & SIGKILL, llbuild worked well.

ddunbar:
Ok, then lets just add that code to the main code path (used on Darwin)... they can't be masked, so there is no reason for trying even on OSes that allow it. This way we at least only have two paths.

tinysun212:
Moved the code to the main path.

ddunbar:
Thanks!

@swift-ci please test

ddunbar:
@swift-ci please test

PR #37, opened 2016-11-03
Add llvm-lit as a 'lit' name to search for.

trombonehero:
In the LLVM build tree, 'lit' is built as 'llvm-lit'. Add the ability for CMake to find it using this name, too.

### DIFF
--- a/cmake/modules/FindLit.cmake
+++ b/cmake/modules/FindLit.cmake
@@ -5,7 +5,7 @@
 # LIT_EXECUTABLE
 
 find_program(LIT_EXECUTABLE
-             NAMES lit
+             NAMES lit llvm-lit
              DOC "Path to 'lit' executable")
 
 # Handle REQUIRED and QUIET arguments, this will also set LIT_FOUND to true


ddunbar:
This doesn't match the intent of `llvm-lit`, though. `llvm-lit` was really just a way for `lit` to find the LLVM build products. I don't think it makes sense for llbuild to pick that up.

Is there a reason you think it should versus finding the regular lit?

trombonehero:
My rationale had been that I was using `llvm-lit` from an LLVM build directory, but I take your point about `lit` vs `llvm-lit`. Perhaps I should withdraw this pull request and simply teach the FreeBSD port to put `/usr/local/llvm38/bin` in the `PATH` to pick up `lit` and `FileCheck`.

ddunbar:
That makes more sense.

I actually am hoping to move lit clients to more of a model where they get lit via `pip install` or something, to use an explicit, installed version. Now that pip has good options for user local installation, this can even be automated by the build process pretty effectively.

PR #38, opened 2016-11-03
Add FreeBSD build support.

trombonehero:
Many of the places in the CMake files that check for Linux really ought to check for *BSD or Linux, but a few of them are FreeBSD-specific (e.g., libexecinfo).

### DIFF
--- a/lib/Ninja/CMakeLists.txt
+++ b/lib/Ninja/CMakeLists.txt
@@ -4,3 +4,5 @@ add_llbuild_library(llbuildNinja
   ManifestLoader.cpp
   Parser.cpp
   )
+
+target_link_libraries(llbuildNinja llvmSupport curses)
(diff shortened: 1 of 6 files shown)


lib/Ninja/CMakeLists.txt
```diff
@@ -4,3 +4,5 @@ add_llbuild_library(llbuildNinja
   ManifestLoader.cpp
   Parser.cpp
   )
+
+target_link_libraries(llbuildNinja llvmSupport curses)
```
ddunbar: Can you explain this line? We shouldn't depend on curses at this layer?

ddunbar:
@swift-ci please test

lib/Ninja/CMakeLists.txt
trombonehero: That's a good point: I've now moved the curses dependency to the `unittests/Ninja/NinjaTests` target rather than `lib/Ninja/llbuildNinja`.

ddunbar:
Thanks!

ddunbar:
@swift-ci please test

ddunbar:
@swift-ci please test

PR #49, opened 2016-12-13
Command skipping

m-i-r-z-a:
This allows a build system delegate to run a subset of the build by explicitly skipping commands yet allowing dependent commands to run.

As a prerequisite, this includes a separate commit to disambiguate BuildValue::SkippedCommand into FailedDependencyCommand and CancelledCommand.

### DIFF
--- a/include/llbuild/BuildSystem/BuildSystemFrontend.h
+++ b/include/llbuild/BuildSystem/BuildSystemFrontend.h
@@ -167,6 +167,21 @@ public:
   /// corresponding \see commandFinished() call.
   virtual void commandPreparing(Command*) override;
 
+  /// Called by the build system to allow the delegate to skip a command without
+  /// implicitly skipping its dependents.
+  ///
+  /// WARNING: Clients need to take special care when using this. Skipping
+  /// commands without considering their dependencies or dependents can easily
+  /// produce an inconsistent build.
+  ///
+  /// This method is called before the command starts, when the system has
+  /// identified that it will eventually need to run (after all of its inputs
+  /// have been satisfied).
+  ///
+  /// The system guarantees that all such calls will be paired with a
+  /// corresponding \see commandFinished() call.
+  virtual bool shouldCommandStart(Command*) override;
+
   /// Called by the build system to report that a declared command has started.
   ///
   /// The system guarantees that all such calls will be paired with a
--- a/include/llbuild/BuildSystem/BuildValue.h
+++ b/include/llbuild/BuildSystem/BuildValue.h
@@ -66,6 +66,9 @@ class BuildValue {
     /// A value produced by a command which was cancelled.
     CancelledCommand,
 
+    /// A value produced by a command which was skipped.
+    SkippedCommand,
+
     /// Sentinel value representing the result of "building" a top-level target.
     Target,
   };
@@ -184,6 +187,9 @@ public:
   static BuildValue makeCancelledCommand() {
     return BuildValue(Kind::CancelledCommand);
   }
+  static BuildValue makeSkippedCommand() {
+    return BuildValue(Kind::SkippedCommand);
+  }
   static BuildValue makeTarget() {
     return BuildValue(Kind::Target);
   }
@@ -204,6 +210,7 @@ public:
   bool isFailedCommand() const { return kind == Kind::FailedCommand; }
   bool isPropagatedFailureCommand() const { return kind == Kind::PropagatedFailureCommand; }
   bool isCancelledCommand() const { return kind == Kind::CancelledCommand; }
+  bool isSkippedCommand() const { return kind == Kind::SkippedCommand; }
   bool isTarget() const { return kind == Kind::Target; }
 
   bool hasMultipleOutputs() const {
--- a/include/llbuild/BuildSystem/BuildValue.h
+++ b/include/llbuild/BuildSystem/BuildValue.h
@@ -50,7 +50,7 @@ class BuildValue {
     /// missing.
     MissingOutput,
 
-    /// A value for a produced output whose command failed.
+    /// A value for a produced output whose command failed or was cancelled.
     FailedInput,
 
     /// A value produced by a successful command.
@@ -58,9 +58,13 @@ class BuildValue {
 
     /// A value produced by a failing command.
     FailedCommand,
-      
-    /// A value produced by a command which was skipped.
-    SkippedCommand,
+
+    /// A value produced by a command which was skipped because one of its
+    /// dependencies failed.
+    PropagatedFailureCommand,
+
+    /// A value produced by a command which was cancelled.
+    CancelledCommand,
 
     /// Sentinel value representing the result of "building" a top-level target.
     Target,
@@ -174,8 +178,11 @@ public:
   static BuildValue makeFailedCommand() {
     return BuildValue(Kind::FailedCommand);
   }
-  static BuildValue makeSkippedCommand() {
-    return BuildValue(Kind::SkippedCommand);
+  static BuildValue makePropagatedFailureCommand() {
+    return BuildValue(Kind::PropagatedFailureCommand);
+  }
+  static BuildValue makeCancelledCommand() {
+    return BuildValue(Kind::CancelledCommand);
   }
   static BuildValue makeTarget() {
     return BuildValue(Kind::Target);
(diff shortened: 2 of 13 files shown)


include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -52,15 +52,21 @@ class BuildValue {
...
     SuccessfulCommand,
 
     /// A value produced by a failing command.
     FailedCommand,
-      
-    /// A value produced by a command which was skipped.
-    SkippedCommand,
+
+    /// A value produced by a command which was skipped because one of its dependencies failed.
+    FailedDependencyCommand,
```
ddunbar: I don't have a better idea, but I would probably pick a different name from `FailedDependencyCommand`, just because it is hard to read when juxtaposed with `FailedCommand`... what about `UnavailableCommand` (under idea it is unavailable because it couldn't run because input was missing) or any other ideas? `SkippedCommandBecauseOfPropagatedFailure` seems too long. :)

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -193,9 +205,11 @@ class BuildValue {
 
   bool isMissingOutput() const { return kind == Kind::MissingOutput; }
   bool isFailedInput() const { return kind == Kind::FailedInput; }
+  bool isCancelledInput() const { return kind == Kind::CancelledInput; }
```
ddunbar: Is there a difference between a cancelled input and a failed input? Both seem like failure, or do we need to retain both cases?

lib/BuildSystem/ExternalCommand.cpp
```diff
@@ -242,19 +244,30 @@ void ExternalCommand::provideValue(BuildSystemCommandInterface& bsci,
...
+        return llvm::None;
+      else
+        return BuildValue::makeMissingInput();
+    }
 
-    // For anything else, this is an error and the command should be skipped.
-    return true;
+    // Propagate cancellation.
+    if (value.isCancelledInput())
+      return BuildValue::makeCancelledCommand();
```
ddunbar: This doesn't seem right to me -- a command which is never started isn't cancelled, it is the "propagated failure" case.

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -52,15 +52,21 @@ class BuildValue {
 
     /// A value for a produced output whose command failed.
     FailedInput,
+      
+    /// A value for a produced output whose command was cancelled.
+    CancelledInput,
```
ddunbar: I think this commit should technically bump `BuildSystemImpl::internalSchemaVersion`, since this effectively will change the binary encoding of values in existing databases. This will force a rebuild for anyone who runs with a database shared before and after this build (something we don't, unfortunately, have any test coverage of).

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -68,6 +68,9 @@ class BuildValue {
     /// A value produced by a command which was cancelled.
     CancelledCommand,
 
+    /// A value produced by a command which was skipped.
+    SkippedCommand,
+
```
ddunbar: This commit should also technically bump `BuildSystemImpl::internalSchemaVersion`, for the same reason as before, although I'm fine just bumping once for the PR.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -524,7 +524,18 @@ class CommandTask : public Task {
   }
 
   virtual void inputsAvailable(BuildEngine& engine) override {
-    command.inputsAvailable(getBuildSystem(engine).getCommandInterface(), this);
+    bool shouldSkip = !getBuildSystem(engine).getCommandInterface()
```
ddunbar: I wonder if we should punt this to the background. The problem with executing this code right here is that this callback is called from the core (serial) build engine loop: the build engine expects clients to immediately offload any work which might be slow. However, since this is a client callback, maybe it is safest to not foist that responsibility onto the client, and instead immediately put this in the background.

ddunbar: BTW, this is off topic but I guess we should move all the `if (bsci.getDelegate().isCancelled()) {` checks here too, so we are consistent about where we implement the BuildSystem-layer special cases. What do you think?

include/llbuild/BuildSystem/BuildSystemFrontend.h
```diff
@@ -160,6 +160,16 @@ class BuildSystemFrontendDelegate : public BuildSystemDelegate {
...
   virtual void commandPreparing(Command*) override;
 
+  /// Called by the build system to allow the delegate to skip a command.
+  ///
+  /// This method is called before the command starts, when the system has
+  /// identified that it will eventually need to run (after all of its inputs
+  /// have been satisfied).
+  ///
+  /// The system guarantees that all such calls will be paired with a
+  /// corresponding \see commandFinished() call.
```
ddunbar: I think we should put a honkin' big warning in here about how this will break the consistency of the build and you should only use it if you know what you are doing. :)

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -524,7 +524,18 @@ class CommandTask : public Task {
...
+      .getDelegate().shouldCommandStart(&command);
+
+    if (shouldSkip) {
+      getBuildSystem(engine).getCommandInterface().taskIsComplete(
+          this, BuildValue::makeSkippedCommand());
+      getBuildSystem(engine).getCommandInterface().getDelegate()
+        .commandFinished(&command);
```
ddunbar: This seems weird to me -- we never call `commandStarted` so why call `commandFinished`?

include/llbuild/BuildSystem/BuildValue.h
m-i-r-z-a: UnavailableCommand reads well, but I wonder if we can find something that's terse yet still completely explicit about what happened? PropagatedFailureCommand?

include/llbuild/BuildSystem/BuildValue.h
m-i-r-z-a: Okay! I didn't notice this. I wonder if we can encode this dependency statically somehow instead of having to remember to bump it?

include/llbuild/BuildSystem/BuildValue.h
m-i-r-z-a: Okay.

lib/BuildSystem/BuildSystem.cpp
m-i-r-z-a: Agreed. I'll track it as a separate item.

lib/BuildSystem/BuildSystem.cpp
m-i-r-z-a: The delegate doc comments guarantee that commandFinished is called after commandPreparing, and commandPreparing has been called by the time shouldCommandStart is called. It's a little confusing because the term Finished is naturally paired with Started, yet Preparing and ShouldStart guarantee Finished. I wonder if Preparing and ShouldStart should guarantee Finished at all? Or if the delegate does need an end callback to match Preparing, maybe it should be something new? Thoughts?

include/llbuild/BuildSystem/BuildSystemFrontend.h
m-i-r-z-a: Okay!

include/llbuild/BuildSystem/BuildValue.h
ddunbar: You mean automatically compute this? I don't think so, but we could of course have better testing around it by capturing example encodings for each schema, and verifying we never get different encodings without also getting a new schema.

include/llbuild/BuildSystem/BuildValue.h
m-i-r-z-a: Updated the PR. We went with PropagatedFailureCommand over chat.

lib/BuildSystem/BuildSystem.cpp
m-i-r-z-a: Sounds good! Updated the PR.

m-i-r-z-a:
@swift-ci please test

m-i-r-z-a:
(For reference, I just rebased and force pushed to update the PR and resolve the comments, but GitHub didn't add a comment about it. The commits view shows the right timestamp.)

ddunbar:
@swift-ci please test

m-i-r-z-a:
Pushed a fixup to reorder BuildSystemTests linkage to fix Linux build.

m-i-r-z-a:
@swift-ci please test

ddunbar:
@swift-ci please test

ddunbar:
@swift-ci please test

m-i-r-z-a:
Tests aborted by bot issues over the weekend...

m-i-r-z-a:
@swift-ci please test

erg:
@swift-ci please test

PR #50, opened 2016-12-14
[libllbuild] Allow clients to supply custom command signatures.

ddunbar:
- This is used by C API clients supplying custom tools to track when those
   tools (which may embed their own state) may need to be rerun, even if other
   aspects of the command haven't changed.


ddunbar:

ddunbar:
@swift-ci please test

PR #51, opened 2016-12-15
[BuildSystem] Allow inputs for Mkdir/Symlink.

ddunbar:
- These will be treated as "must follow" inputs, since the commands themselves
   do not use them, but this allows clients to control the order in which these
   commands are run.


ddunbar:
@swift-ci please test

ddunbar:
@swift-ci please test Linux

PR #57, opened 2016-12-20
Fix command line arguments passed to the MSVC compiler invocation

hughbe:
Firstly, this avoids compiler errors building LLBuild, owing to unknown compiler flags
Secondly, this silences some warnings that we don't want (e.g. unknown pragma), but also ones that LLVMSupport ignores to avoid a cluttered build log and to avoid needing to modify the llvmSupport code


ddunbar:
@swift-ci please test

PR #59, opened 2016-12-21
Port FileInfo::getInfoForPath to Windows

hughbe:


hughbe:
@ddunbar is it worth extracting this to the new PlatformUtilities file? Not sure, considering this is the only place we use this kind of code.

ddunbar:
I'm fine leaving this here, as FileInfo's purpose is really to abstract over the platform differences in any case.

lib/Basic/FileInfo.cpp
```diff
@@ -14,22 +14,93 @@
 
 #include <cassert>
 #include <cstring>
+
+#if defined(_WIN32)
+#include <windows.h>
```
ddunbar: Same WIN32_LEAN_AND_MEAN comment as in other PR...

lib/Basic/FileInfo.cpp
```diff
@@ -14,22 +14,93 @@
...
 using namespace llbuild;
 using namespace llbuild::basic;
 
 bool FileInfo::isDirectory() const {
-  return mode & S_IFDIR;
+  return (mode & S_IFDIR) != 0;
 }
 
+#if defined(_WIN32)
+static void FileTimeToTimestamp(FILETIME fileTime, FileTimestamp *stamp) {
```
ddunbar: The intent of FileInfo is not to present a normalized view of file time... I can see how it is nice to maintain a consistent reference point, but I'm not sure llbuild really should end up interpreting these times other than to compare them with one another.

In that case, we could simply decide to document that and not bother to put that conversion here. What do you think?

lib/Basic/FileInfo.cpp
```diff
@@ -14,22 +14,93 @@
...
 /// \returns True if information on the path was found.
 FileInfo FileInfo::getInfoForPath(const std::string& path, bool asLink) {
   FileInfo result;
+
+#if defined(_WIN32)
+  DWORD attributes;
+  if (asLink) {
+    attributes = FILE_FLAG_OPEN_REPARSE_POINT;
+  }
+  else {
```
ddunbar: Nit: `} else {` please!

lib/Basic/FileInfo.cpp
```diff
@@ -14,22 +14,93 @@
...
+  auto mode = S_IREAD;
+  if (fileAttributes & FILE_ATTRIBUTE_DIRECTORY)
+    mode |= S_IFDIR;
+  else
+    mode |= S_IFREG;
+
+  if ((fileAttributes & FILE_ATTRIBUTE_READONLY) == 0)
+    mode |= S_IWRITE;
+
+  result.mode = mode;
```
ddunbar: The same comments as for time apply here for mode... I'm not sure we should attempt to interpret many of the attributes, and if we do it would probably be better to lift them to be explicit platform-independent flags rather than to try and force the Windows file info to look like POSIX file info.

lib/Basic/FileInfo.cpp
```diff
@@ -14,22 +14,93 @@
...
+  if (!GetFileInformationByHandle(fileHandle, &fileInformation)) {
+    CloseHandle(fileHandle);
+    memset(&result, 0, sizeof(result));
+    assert(result.isMissing());
+    return result;
+  }
+
+  FileTimeToTimestamp(fileInformation.ftLastWriteTime, &result.modTime);
+  result.inode = 0;
+  result.device = _getdrive() - 1;
```
ddunbar: `inode` and `device` are just used for comparison purposes...

On POSIX, inode is useful for detecting cases where a file was changed (replaced) by a file with an identical timestamp and size. This is uncommon but possible, especially on file systems with poor timestamp resolution (I'm not sure what the native timestamp resolutions of NTFS or FAT are).

For the purposes here, if Windows exposes such an entity you can map or mangle it into inode and assume llbuild won't try and do anything else with it other than compare it.

lib/Basic/FileInfo.cpp
```diff
@@ -14,22 +14,93 @@
...
 /// system.
 ///
 /// \param info_out [out] On success, the important path information.
 /// \returns True if information on the path was found.
 FileInfo FileInfo::getInfoForPath(const std::string& path, bool asLink) {
   FileInfo result;
+
+#if defined(_WIN32)
+  DWORD attributes;
+  if (asLink) {
```
ddunbar: I wonder if this is necessary... we use this in cases where the build is attempt to create a symbolic link and needs to examine the attributes of the file representing the link, but IIRC Windows style links are far less useful and I'm not sure if anyone would want to use that in the same way... or if they did, they would probably want it to be done by a Windows specific tool not the general purpose one.

I'm not familiar enough with Windows-style links to know if this will be useful elsewhere, do you think it will?

lib/Basic/FileInfo.cpp
hughbe: I'll include `LeanWindows.h` when #61 is merged

lib/Basic/FileInfo.cpp
hughbe: The reason why I decided to add this conversion helper is because `FileTimestamp` is defined as follows
```
struct FileTimestamp {
  uint64_t seconds;
  uint64_t nanoseconds;
}
```

> In that case, we could simply decide to document that and not bother to put that conversion here. What do you think?

The concern I have with this is that we'd have to change `FileTimestamp` into some kind of platform specific entity. This would require explicitly defining all the comparison operators again, and would be complicated. 

I'm not really sure if its overkill or not for this. That said, you could probably call the current method overkill. Up to you!

lib/Basic/FileInfo.cpp
hughbe: Symbolic links are pretty useless on Windows. In fact, you need elevated privileges to actually create one, so they're rarely used. I know LLVM's build process uses copying instead of symlinks for this reason.

We could really simplify most of this PR if we ignore symbolic links. This would let us use the Windows function `_stat(2)`, which ignores symlinks, but also would remove the need for the conversion code.

I didn't want to diverge too much from the behaviour on POSIX systems, however. That said, these platforms are intrinsically different, so although this would create a behaviour difference between Windows and Unix, it may not be too bad. Again, your decision, I'm open either way

lib/Basic/FileInfo.cpp
hughbe: I'll bear this in mind. Fixed

lib/Basic/FileInfo.cpp
hughbe: I think we can use the `nFileIndexHigh` and `nFileIndexLow` in the following way (I've updated the PR):
```
  result.inode = ((uint64_t)fileInformation.nFileIndexHigh << 32) +
    (uint64_t)fileInformation.nFileIndexLow;
```

lib/Basic/FileInfo.cpp
hughbe: I went with this approach because this is what libGit2 goes for: https://github.com/libgit2/libgit2/blob/15769731a2411933bb9236d06c5f15e2c4e7929b/src/win32/w32_util.h#L121

Do you suggest defining our own file attributes enumeration, defining stuff like canRead, canWrite, isDirectory, isFile etc. etc. (naming to be decide).

I'm not sure if this would be overkill - it seems like `mode` is only used in `BuildValue::dump`.

lib/Basic/FileInfo.cpp
ddunbar: I vote for ignoring `isLink` here, at least to start with, under the expectation that it won't be useful on Windows and we will have to more explicitly model the actual platform behavior when we want to.

lib/Basic/FileInfo.cpp
ddunbar: It's not obvious, but we also use it in comparisons, because we directly compare the literal bytes of the `FileInfo` inside the build system.

lib/Basic/FileInfo.cpp
ddunbar: Can you update this for LeanWindows?

hughbe:
@swift-ci please smoke test

hughbe:
I've updated the PR to address your comments :)

lib/Basic/FileInfo.cpp
```diff
@@ -46,6 +48,9 @@ FileInfo FileInfo::getInfoForPath(const std::string& path, bool asLink) {
 #if defined(__APPLE__)
   auto seconds = buf.st_mtimespec.tv_sec;
   auto nanoseconds = buf.st_mtimespec.tv_nsec;
+#elif defined(_WIN32)
+  auto seconds = buf.st_mtime;
+  auto nanoseconds = 0;
```
ddunbar: Does Win32 provide no way to get finer grained timestamp resolution? 1s resolution can be problematic, and is nice to avoid if at all possible.

hughbe: Doesn't seem like it, at least not with `_stat`. I'll take a look for other APIs, though

PR #60, opened 2016-12-21
Port lib/Core to Windows/MSVC

hughbe:
> 1>C:\Users\hbellamy\Documents\GitHub\my-swift\llbuild\lib\Core\BuildEngine.cpp(974): error C2440: 'initializing': cannot convert from 'initializer list' to '`anonymous-namespace'::BuildEngineImpl::reportCycle::WorkItem'
> 1>  C:\Users\hbellamy\Documents\GitHub\my-swift\llbuild\lib\Core\BuildEngine.cpp(974): note: No constructor could take the source type, or constructor overload resolution was ambiguous
> 1>C:\Users\hbellamy\Documents\GitHub\my-swift\llbuild\lib\Core\BuildEngine.cpp(995): error C2440: 'initializing': cannot convert from 'initializer list' to '`anonymous-namespace'::BuildEngineImpl::reportCycle::WorkItem'
> 1>  C:\Users\hbellamy\Documents\GitHub\my-swift\llbuild\lib\Core\BuildEngine.cpp(995): note: No constructor could take the source type, or constructor overload resolution was ambiguous
> 1>C:\Users\hbellamy\Documents\GitHub\my-swift\llbuild\lib\Core\BuildEngine.cpp(995): error C2672: 'std::vector<`anonymous-namespace'::BuildEngineImpl::reportCycle::WorkItem,std::allocator<_Ty>>::emplace_back': no matching overloaded function found
> 1>          with
> 1>          [
> 1>              _Ty=`anonymous-namespace'::BuildEngineImpl::reportCycle::WorkItem
> 1>          ]

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -972,6 +972,8 @@ private:
     
     // Find the cycle by searching from the entry node.
     struct WorkItem {
+      WorkItem(Rule * node) { this->node = node; }
+
       Rule* node;
       unsigned predecessorIndex = 0;
     };


lib/Core/SQLiteBuildDB.cpp
```diff
@@ -86,7 +90,11 @@ class SQLiteBuildDB : public BuildDB {
       sqlite3_close(db);
 
       // Always recreate the database from scratch when the schema changes.
+#if defined(_WIN32)
+      result = _unlink(path.str().c_str());
+#else
       result = unlink(path.str().c_str());
+#endif
```
ddunbar: Same request as with `chdir`, let's define a platform independent utility function for this.

lib/Core/BuildEngine.cpp
```diff
@@ -965,6 +965,8 @@ class BuildEngineImpl {
     
     // Find the cycle by searching from the entry node.
     struct WorkItem {
+      WorkItem(Rule * node) { this->node = node; }
+
```
ddunbar: /cc @DougGregor sorry to bug you but is it not legal C++14 to initialize this type with `WorkItem{ node }`? I'm trying to understand why Clang and MSVC aren't consistent in this case. I remember there is some subtly w.r.t. single item ctors and initializer_list but I've forgotten the specifics...

hughbe: The following code works with Clang but not MSVC. I can make a connect bug report if you think its an MSVC bug. I've been trying to make bug reports for each difference between the two compilers whilst porting Swift to MSVC:
```
int main()
{
  struct WorkItem {
    int node;
    unsigned predecessorIndex = 0;
  };  

  WorkItem{ 0 };

  return 0;
}
```

hughbe: ... interestingly, removing the default value of `predecessorIndex` allows MSVC to compile this just fine

ddunbar: I don't know, I would need to read the standard for the uniform initializer behavior to understand why there might be a divergence here.

I suspect that `WorkItem{ 0, 0 }` works on both compilers?

hughbe: Nope, that doesn't work. This is because of a C++ 14 feature (Aggregate initialization with default member initializers) that isn't supported until VS 2017.

hughbe:
I've changed this PR only to deal with the MSVC fixes - this is because I've sent in #69

lib/Core/BuildEngine.cpp
ddunbar: Ah, cool thanks for tracking it down.

ddunbar:
@swift-ci please test

PR #62, opened 2016-12-21
Fix MSVC warning building llbuild

hughbe:


ddunbar:
Is it possible to disable this MSVC warning for us, for now? If not, then I would rather find a way we can get the same warnings from Clang so that we all use a compatible warning set.

ddunbar:
(we recently chose not to turn on signed conversion warnings for Clang, because I didn't feel it was actually an improvement to code quality to just do spot fixes like this)

hughbe:
I think silencing them is the best idea, turns out we already did this in #57. I have reverted all my changes apart from turning `maximumLoadAverage` from a `float` to a `double`. This is because it is set by `::strtod`, so there is an actual risk of overflow here

ddunbar:
Great!

ddunbar:
@swift-ci please test

ddunbar:
@swift-ci please test

hughbe:
I think the CI is down over christmas until 3rd Jan

ddunbar:
@swift-ci please test

PR #72, opened 2017-01-18
[BuildSystem] Fix BuildValue initialization & serialization.

ddunbar:
- The initializer was allowing uninitialized data in the commandSignature for
   sentinel values (like virtual inputs).

 - The move initialization was failing to initialize some fields (!).

 - The serialization was also failing to properly normalize values, so the
   serialized value was not properly deterministic.

### DIFF
--- a/include/llbuild/BuildSystem/BuildValue.h
+++ b/include/llbuild/BuildSystem/BuildValue.h
@@ -100,6 +100,7 @@ private:
   BuildValue() {}
   BuildValue(Kind kind) : kind(kind) {
     valueData.asOutputInfo = {};
+    commandSignature = 0;
   }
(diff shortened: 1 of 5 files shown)


ddunbar:
@swift-ci please test

ddunbar:
@m-i-r-z-a can you review? We really need some proper serialization support code, but in lieu of that this at least fixes a major problem.

ddunbar:
@swift-ci please smoke test

ddunbar:
LLVM does a SmallVector, but because we are serializing by type punning the bytes we have to directly manage all of the implementation details. It's also significantly larger in size for the 1-or-many case because of the general purpose nature.

I suspect we could clean it up in other ways though if we had explicit serialization instead of type punning.

PR #77, opened 2017-01-23
Fix LLBUILD_EXPORT for Windows

hughbe:
Windows requires the use of `__declspec(dllexport)`. This works with MSVC, Clang and GCC. This fixes the exported API on Windows.

However, MSVC doesn't understand `__attribute()` syntax, so this also fixes many build errors.


hughbe:
@swift-ci please test

PR #79, opened 2017-01-23
Invoke llvm-lit.py with Python

hughbe:
Python executables have to be executed via `python x.py` on Windows. Keep this on all platforms, to avoid platform specific code.


hughbe:
@swift-ci please test

PR #83, opened 2017-01-23
Fix MSVC warnings for unknown escape character sequence

hughbe:
> warning C4129: 'e': unrecognized character escape sequence

I'm not sure what the intention here is:
- Accidental inclusion of `\error`: there are other cases in this file of `error` without `\error`
- Use of GCC extension escape sequence `\e`
- Wanting to print `\` to the stream, ie. `\\error`

I went with the first, but let me know


hughbe:
Let's see if I'm right:

@swift-ci please test

ddunbar:
@swift-ci please smoke test

PR #86, opened 2017-01-23
Define a platform independent abstraction for handling program signals

hughbe:
Windows doesn't support `::sigaction`, as `::sigaction` is really complicated and Unix specific.
Fortunately, this isn't the end of the world, as Windows supplies `::signal` which can do the same behaviour we need.

However, the semantics of the two methods are different - `sigaction` is more powerful than `signal`, so has a more convoluted API.

This means that we need to introduce a wrapper for platform independent signalling operations.

### DIFF
--- a/lib/Basic/CMakeLists.txt
+++ b/lib/Basic/CMakeLists.txt
@@ -6,6 +6,7 @@ add_llbuild_library(llbuildBasic
   SerialQueue.cpp
   Version.cpp
   ShellUtility.cpp
+  SignalHandlerWrapper
   )
 
 if(${CMAKE_SYSTEM_NAME} MATCHES ".*BSD" OR ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
--- a/lib/Basic/SignalHandlerWrapper.cpp
+++ b/lib/Basic/SignalHandlerWrapper.cpp
@@ -1,0 +1,44 @@
+//===- Support/SignalHandlerWrapper.cpp - Signal handling -----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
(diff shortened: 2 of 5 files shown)


hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

lib/Basic/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ add_llbuild_library(llbuildBasic
   SerialQueue.cpp
   Version.cpp
   ShellUtility.cpp
+  SignalHandlerWrapper
```
ddunbar: Is .cpp not required? Guess not...

lib/Basic/SignalHandlerWrapper.cpp
```diff
@@ -0,0 +1,35 @@
...
+    previous->handler = handler;
+  }
+#else
+  struct sigaction action {};
+  action.sa_handler = handler;
+
+  struct sigaction previousHandler;
+  ::sigaction(sig, &action, &previousHandler);
+  if (previous) {
+    previous->handler = previousHandler.sa_handler;
```
ddunbar: This isn't quite the same, because it won't restore the other parts of the old handler (like the mask or flags). I suggest making this an struct with a private void member which can then be casted in the implementation to a type containing the actual instance (which will have to be allocated, for simplicity) and that can then just have a `struct sigaction` inside.

hughbe:
@swift-ci please smoke test

hughbe:
Accidentally nuked this branch... have to reopen PR

PR #89, opened 2017-01-24
Define a platform independent abstraction for handling program signals

hughbe:
Windows doesn't support ::sigaction, as ::sigaction is really complicated and Unix specific.
Fortunately, this isn't the end of the world, as Windows supplies ::signal which can do the same behaviour we need.

However, the semantics of the two methods are different - sigaction is more powerful than signal, so has a more convoluted API.

This means that we need to introduce a wrapper for platform independent signalling operations.


hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

lib/Basic/SignalHandlerWrapper.cpp
```diff
@@ -0,0 +1,44 @@
...
+using namespace llbuild::basic;
+using namespace llbuild::basic::sys;
+
+SignalHandlerWrapper::SignalHandlerWrapper(Handler handler) {
+#if defined(_WIN32)
+  innerAction = handler;
+#else
+  struct sigaction action {};
+  action.sa_handler = handler;
+  innerAction = &action;
```
ddunbar: This isn't memory safe, innerAction needs to be allocated somewhere, otherwise it is a pointer to dangling memory.

hughbe: I'm not a c++ expert, so just to clarify :D
By allocated **somewhere**, would this constructor work? As in
```
SignalHandlerWrapper::SignalHandlerWrapper(Handler handler) {
...
#else
  innerAction = new struct sigaction();
  innerAction = &action;
}
```

I assume we'd need to free this memory at some point? Something like
```
~SignalHandlerWrapper() {
// Unix only?
  delete innerAction;
}
```

ddunbar: That is correct, you will need to cast it to the correct type. Your initializer also isn't quite right, since it overwrites the pointer, it would need to be
```
*innerAction = action
```
but maybe that is what you meant.

hughbe: > but maybe that is what you meant.

It wasn't, but I'll do that :D

For the Windows variable, it's a function pointer - do I need to allocate and free memory for that?

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test OS X Platform

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

lib/Basic/SignalHandlerWrapper.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+SignalHandlerWrapper::~SignalHandlerWrapper() {
+#if !defined(_WIN32)
+  delete static_cast<struct sigaction *>(innerAction);
+#endif
+}
+
+void SignalHandlerWrapper::registerAction(int sig,
+                                          SignalHandlerWrapper *previous) {
```
ddunbar: I think it would make more sense to just take these arguments on the constructor, wouldn't it?

hughbe:
Sure I can do that - I've got a VM so can pull changes all the time and test them

hughbe:
Okay, so I've looked into making this a higher level abstraction.

In the process, I noticed a high level of code duplication between `NinjaBuildCommand` and `BuildSystemCommand`.

So I did a bit of refactoring - I think some of it's pretty ugly right now. This is because:
- we share static state: there is a different static instance of `int signalWatchingPipe[2]`, `std::atomic<bool> wasInterrupted`. I'm fairly sure the answer to this is no, but should we consolidate the duplicated static fields between `NinjaBuildCommand` and `BuildSystemCommand`?  This current implementation means we have to pass pointers to the static fields, causing extra code. However, I'm not sure I understand the build system well enough to know if they should be consolidated into static fields in `AdvancedInterruptSignalHandler`
- we need a way to call back to a `void()` function when the signal is interrupted. I'm probably not doing it right currently, but this general approach saves duplicate implementations of `signalWaitThread` and `checkForInterrupt`. This is also necessary to get the `new std::thread(&AdvancedInterruptSignalHandler::signalWaitThread, this);` code compiling

This refactoring makes it simpler to support Windows too - but I think the LOC diff is skewed by the addition of 2 license headers!

ddunbar:
Yes, I think we should coalesce the static state -- the infrastructure only needs exactly one instance of those fields.

Ideally, what would happen is that the first time anyone used the infrastructure it would allocate the single static instance of the a helper class which spins up the pipe, the background thread, and the signal handler.

hughbe:
> Ideally, what would happen is that the first time anyone used the infrastructure it would allocate the single static instance of the a helper class

I've taken a shot at doing this - I just want to make sure that I'm correctly setting and removing the listener for interrupt signals

ddunbar:
Thanks, I'm a bit under the weather but will take a look soon (same goes for #98)

hughbe:
Get better! No rush

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

include/llbuild/Basic/SignalHandler.h
```diff
@@ -0,0 +1,35 @@
...
+// This file implements a wrapper for handling signals.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLBUILD_BASIC_SIGNALHANDLER_H
+#define LLBUILD_BASIC_SIGNALHANDLER_H
+
+namespace llbuild {
+namespace basic {
+struct SignalHandler {
```
ddunbar: We need some documentation comments here.

What I think we want the API to look like is something more like one method which registers an *interrupt* handler (ideally a lambda) specifically (I don't think we need to worry about any other signals at the moment, and other signals come with other problems).

That should return a token the client can use to unregister it.

That should probably be a static method, since there should only be one instance of the thing managing the handlers.

Internally, the manager for the handlers will simply keep a mapping between tokens and handlers (the token can be as simple as an ID we just increment, or an opaque pointer) and when an interrupt happens it will just call them all in some unspecified order.

lib/Commands/InterruptSignalAwaiter.h
```diff
@@ -0,0 +1,65 @@
...
+#include <thread>
+
+namespace llbuild {
+namespace command {
+struct InterruptSignalAwaiter {
+private:
+  basic::SignalHandler handler{SIGINT, &sigintHandler};
+  static int signalWatchingPipe[2];
+  static std::atomic<bool> wasInterrupted;
+  std::function<void()> interruptHandler;
```
ddunbar: I would probably try and hide all of this in a single implementation file.

ddunbar: I see now that this type is closer to the thing I  had imagined -- I would probably just merge the signal handler method with this code since it should end up being the only thing which interacts with the OS level facility.

lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -479,38 +480,20 @@ class BuildContext {
...
-    char byte{};
-    sys::write(signalWatchingPipe[1], &byte, 1);
-  }
-
+  
   /// Cancel the build in response to an interrupt event.
-  void cancelBuildOnInterrupt() {
-    std::lock_guard<std::mutex> guard(spawnedProcessesMutex);
+  static void interruptedHandler(BuildContext *context) {
+    std::lock_guard<std::mutex> guard(context->spawnedProcessesMutex);
```
ddunbar: I'd probably chain to a non-static method here, but a better option is to just allow a lambda for the handler so that this isn't even necessary.

ddunbar: Actually your current implementation does that because it uses `std::function`, you could register a lambda for the handler already.

lib/Commands/InterruptSignalAwaiter.cpp
```diff
@@ -0,0 +1,75 @@
...
+    perror("pipe");
+  }
+
+  new std::thread(&InterruptSignalAwaiter::signalWaitThread, this);
+}
+
+InterruptSignalAwaiter::~InterruptSignalAwaiter() {
+  // Close the signal watching pipe.
+  basic::sys::close(signalWatchingPipe[1]);
+  signalWatchingPipe[1] = -1;
```
ddunbar: It might be nice for this to join on the signal watching thread, just to ensure a tidy shutdown.

lib/Commands/InterruptSignalAwaiter.h
```diff
@@ -0,0 +1,65 @@
...
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements a wrapper for handling signals and processing arrived
+// signals.
+//
+//===----------------------------------------------------------------------===//
```
ddunbar: FWIW, I don't usually follow the LLVM convention of putting comments at the top of the file, I prefer putting proper documentation comments associated with actual types; those get found by automated tools.

hughbe:
Sorry for the delay - it's bad form to ping you then take 2 weeks responding to your feedback!

I'm getting llbuild working with VS2017 so will get to this within a couple of hours.

lib/Commands/InterruptSignalAwaiter.cpp
hughbe: Could you clarify this comment? I'm not sure I fully understand. Cheers

ddunbar: It doesn't wait for the signal watching thread to complete before finishing the destructor, so the thread associated with the type can outlive it. I was suggesting making the destructor wait for the thread to complete (by using thread::join) before finishing.

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
I've added some unit tests. Let me know if you think of any other scenarios

unittests/Commands/InterruptSignalAwaiterTests.cpp
```diff
@@ -0,0 +1,54 @@
...
+//
+// Copyright (c) 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#include "../lib/Commands/InterruptSignalAwaiter.h"
```
hughbe: this is ugly - I'm not sure the best way to include this file to avoid exposing this as public API

unittests/Commands/InterruptSignalAwaiterTests.cpp
```diff
@@ -0,0 +1,54 @@
...
+TEST(InterruptSignalAwaitertests, setInterruptHandler) {
+  bool calledInterruptHandler = false;
+  InterruptSignalAwaiter::GlobalAwaiter.setInterruptHandler([&] {
+    calledInterruptHandler = true;
+  });
+
+  std::raise(SIGINT);
+  
+  // Wait a bit to make sure the interrupt signal handler was called.
+  std::this_thread::sleep_for(std::chrono::nanoseconds(1));
```
hughbe: If we don't have this, for any length of time (1ns works), then the signal handler gets called after `EXPECT_TRUE(calledInterruptHandler);`

hughbe: Might be worth extending this time, I'm not sure

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
Dan, id like to see this through and continue working on windows support for llbuild. Could you let me know if there's anything you need from my end to make reviewing this a bit easier? Maybe @neonichu could also take a look?

ddunbar:
Hey @hughbe, I'm really sorry for the long delay -- we were very strapped for time trying to get llbuild in shape for use in Xcode 9. I plan to get to your outstanding PRs next week!

lib/Commands/InterruptSignalAwaiter.cpp
```diff
@@ -0,0 +1,88 @@
+//===-- InterruptSignalAwaiter.cpp ----------------------------------------===//
```
ddunbar: I would put this in `Basic` since it has no higher dependencies, and is general purpose infrastructure.

lib/Commands/InterruptSignalAwaiter.cpp
```diff
@@ -0,0 +1,88 @@
...
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
```
ddunbar: 2017 :)

lib/Commands/InterruptSignalAwaiter.h
```diff
@@ -0,0 +1,67 @@
...
+#include <thread>
+
+namespace llbuild {
+namespace command {
+/// A platform-independent abstraction that allows users to provide a
+/// function that is called when a SIGINT signal is received by the program.
+struct InterruptSignalAwaiter {
+ private:
+  void(*previousSignalHandler)(int);
+  static int signalWatchingPipe[2];
```
ddunbar: This shouldn't be `static` since each instance is recreating the pipe. Either the expose API should hide the fact that there is a singleton (and not expose the underlying type), or it needs to "behave" if there were multiple instances created, even if it wouldn't make much sense to do so.

hughbe: This has to be static, because it is used in the signal handler `static void sigintHandler(int);` which has to be static:

```cpp
void InterruptSignalAwaiter::sigintHandler(int) {
  // Set the atomic interrupt flag.
  wasInterrupted = true;

  // Write to wake up the signal monitoring thread.
  char byte{};
  basic::sys::write(signalWatchingPipe[1], &byte, 1);
}
```

Do correct me if I'm wrong though

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
Looks like this is compiling and testing now successfully. Apologies for the previous attempts after the rebase

dmbryson:
Sadly, as this has sat around a bit, some non-trivial conflicts have arisen.  The idea overall seems sound. Mind rebasing it?

PR #90, opened 2017-01-24
Add implementation of vasprintf on Windows

hughbe:
`vsnprintf(NULL, ...)` returns the length of the output string.

Therefore, use this before calling `vsnprintf` again, to figure out the correct length of the string.

Discussion:
`  if (::vsnprintf(buf, sizeof(buf), fmt, ap) < 0) {` can't really fail, as we're guaranteed to have the correct length.
Should we use an assertion here, instead of an if?

### DIFF
--- a/lib/Commands/NinjaBuildCommand.cpp
+++ b/lib/Commands/NinjaBuildCommand.cpp
@@ -61,13 +61,19 @@ static uint64_t getTimeInMicroseconds() {
   return now.msec();
 }
 
-static std::string getFormattedString(const char* fmt, va_list ap) {
-  char* buf;
-  if (::vasprintf(&buf, fmt, ap) < 0) {
+static std::string getFormattedString(const char* fmt, va_list ap1) {
+  va_list ap2;
+  va_copy(ap2, ap1);
+  int count = vsnprintf(NULL, 0, fmt, ap1);
+  if (count <= 0) {
     return "unable to format message";
   }
-  std::string result(buf);
-  ::free(buf);
+
+  std::string result = std::string(count, NULL);
+  if (vsnprintf(const_cast<char *>(result.c_str()), count + 1, fmt, ap2) < 0) {
+    return "unable to format message";
+  }
+
   return result;
 }
 


hughbe:
@swift-ci please smoke test

lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -61,10 +61,16 @@ static uint64_t getTimeInMicroseconds() {
...
 static std::string getFormattedString(const char* fmt, va_list ap) {
-  char* buf;
-  if (::vasprintf(&buf, fmt, ap) < 0) {
+  int count = ::vsnprintf(NULL, 0, fmt, ap);
+  if (count < 0) {
     return "unable to format message";
   }
+
+  char *buf = static_cast<char *>(malloc(count + 1));
+  if (::vsnprintf(buf, sizeof(buf), fmt, ap) < 0) {
```
grp: Jumping in as an outside person interested in llbuild: you could use `std::string` directly as the buffer here, since it's not coming from `asprintf()` any more.

hughbe: Good idea! Also helps us avoid `free` and `malloc` which feels a bit to c-like to me!

hughbe:
@swift-ci please smoke test

hughbe:
could someone trigger the tests for me - it seems I can't after pushing (with and without -f) to any PR @ddunbar @shahmishal FYI

ddunbar:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

hughbe:
Hmm, looks like there's an odd behaviour difference on Windows vs Unix that's causing the builds to fail:

Same code:
```
#include <iostream>
#include <stdarg.h>

static std::string getFormattedString(const char* fmt, va_list ap) {
  int count = vsnprintf(NULL, 0, fmt, ap);
  if (count <= 0) {
    return "unable to format message";
  }

  std::string result = std::string(count, NULL);
  if (vsnprintf(const_cast<char *>(result.c_str()), count + 1, fmt, ap) <
    0) {
    return "unable to format message";
  }

  return result;
}

static std::string getFormattedString(const char* fmt, ...) {
  va_list ap;
  va_start(ap, fmt);
  std::string result = getFormattedString(fmt, ap);
  va_end(ap);
  return result;
}

int main(int argc, char *argv[]) {

  std::cout << getFormattedString("a").c_str() << "\n";
  std::cout << getFormattedString("abcdefghijklmnop").c_str() << "\n";
  std::cout << getFormattedString("1________________________________________________A").c_str() << "\n";
  std::cout << getFormattedString("%d", 12345).c_str() << "\n";
  std::cout << getFormattedString("a%de", 12345).c_str() << "\n";

  return 0;
}
```

Windows: http://rextester.com/WHTP25141
> a
> abcdefghijklmnop
> 1________________________________________________A
> 12345
> a12345e

Unix: http://rextester.com/GJMKYN47723
> a
> abcdefghijklmnop
> 1________________________________________________A
> 0
> a8e

Anyone got some ideas?

hughbe:
The endless battle against platforms should be fixed. Tested on a VM, seems to work now.

@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

tbkka:
Your sample code is missing a `va_copy`.  As I understand it, the `va_list` type is a reference type on some platforms -- on those platforms, your first call to `vsnprintf` actually consumes the `va_list` so the second call to `vsnprintf` will not work as you expect.  You need to `va_copy` the varargs list to avoid this problem.

hughbe:
> Your sample code is missing a va_copy. As I understand it, the va_list type is a reference type on some platforms -- on those platforms, your first call to vsnprintf actually consumes the va_list so the second call to vsnprintf will not work as you expect. You need to va_copy the varargs list to avoid this problem.

That was the problem :D I updated the code about 15 mins ago :)

PR #91, opened 2017-01-24
Use C++ 11 thread APIs instead of usleep

hughbe:


hughbe:
@swift-ci please smoke test

PR #96, opened 2017-01-26
[BuildSystem] Add support for an 'is-mutated' node flag.

ddunbar:
- This can be used to indicate a file which is mutated by subsequent commands,
   and cannot have its stat information (other than existence) meaningfully
   checked. This can be used as with command timestamp nodes as part of a
   primitive mechanism for supporting commands which mutate their inputs.


ddunbar:
@swift-ci please smoke test

PR #97, opened 2017-01-26
[BuildSystem] Switch MkdirCommand to be be an ExternalCommand.

ddunbar:
- The most important impact of this is that it allows this command to use all
   of the general purpose machinery from the ExternalCommand w.r.t. input and
   output handling. In particular, MkdirCommand's can now drive virtual output
   nodes.


ddunbar:
@swift-ci please smoke test

PR #98, opened 2017-01-26
Refactor code for creating a process and redirecting input and output

hughbe:
The code in `NinjaBuildCommand.cpp` and `LaneBasedExecutionQueue.cpp` for creating processes is identical (with a few formatting and comment changes).

This means that we can refactor the code to avoid duplication.

However, refactoring is also necessary to make this more portable to Windows (which will be in a subsequent PR to simplify your work as reviewers).

The process of creating a process and reading input has been split up into multiple sections, based on functionality, and to make it simpler to identify which bits do what. This will also simplify large blocks of platform specific code in the future, and provide an abstraction of behaviour.

In each file, we:
- Open a pipe
- Create a process, and associate the process and the pipe
- Read from the pipe
- Wait until the process closes
- Assess the return result of the process

We also perform related changes such as killing a process. These require us to allow the `RedirectedProcess` abstraction to be hashable and equatable to be contained in `std::unordered_set`. Of course, the hash code and equality code with differ on Windows vs Unix platforms.

As part of aligning the code to be the same in `NinjaBuildCommand` and `LaneBasedExecutionQueue`, **`NinjaBuildCommand` reads into `llvm::SmallString<1024>` rather than `std::vector<char>`

This does incur one behaviour change - NinjaBuildCommand now considers SIGABRT to be a cancellation signal, aligning it with the behaviour of `LaneBasedExecutionQueue.cpp` introduced in #74

@neonichu FYI

Also, there's a TODO in `~RedirectedProcess` that I'd like to call out as I'm not sure what's the right thing to do

### DIFF
--- a/include/llbuild/Basic/RedirectedProcess.h
+++ b/include/llbuild/Basic/RedirectedProcess.h
@@ -1,0 +1,68 @@
+//===- RedirectedProcess.h --------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+//
+// This file contains abstractions for creating a process and redirecting
+// input and output of that process.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLBUILD_BASIC_REDIRECTEDPROCESS_H
+#define LLBUILD_BASIC_REDIRECTEDPROCESS_H
+
+#include "llvm/ADT/SmallString.h"
+
+#include <mutex>
+#include <vector>
+
+namespace llbuild {
+namespace basic {
+namespace sys {
+class RedirectedProcess {
+  bool shouldCaptureOutput;
+  struct ProcessInfo *innerProcessInfo;
+
+public:
+  RedirectedProcess(bool shouldCaptureOutput);
+
+  bool openPipe();
+
+  bool execute(const char *path, bool setGroupFlags, char *const *args,
+               char *const *envp, std::mutex &spawnedProcessMutex);
+
+  bool readPipe(llvm::SmallString<1024> &output);
+
+  bool waitForCompletion(int *exitStatus);
+
+  bool kill(int signal);
+
+  bool operator==(const RedirectedProcess &rhs) const;
+  size_t hash() const;
+
+  ~RedirectedProcess();
+  
+  static int sigkill();
+
+  static bool isProcessCancelledStatus(int status);
+};
+}
+}
+}
+
+namespace std {
+template <> struct hash<llbuild::basic::sys::RedirectedProcess> {
+  size_t operator()(const llbuild::basic::sys::RedirectedProcess &x) const {
+    return x.hash();
+  };
+};
+}
+
+#endif
--- a/lib/Basic/RedirectedProcess.cpp
+++ b/lib/Basic/RedirectedProcess.cpp
@@ -1,0 +1,217 @@
+//===-- RedirectedProcess.cpp ---------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#include "llbuild/Basic/RedirectedProcess.h"
+
+#include <signal.h>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <spawn.h>
+#include <string.h>
+#include <sys/wait.h>
+
+using namespace llbuild;
+using namespace llbuild::basic;
+using namespace llbuild::basic::sys;
+
+struct llbuild::basic::sys::ProcessInfo {
+  pid_t processId;
+  int readHandle;
+  int writeHandle;
+};
+
+RedirectedProcess::RedirectedProcess(bool shouldCaptureOutput) {
+  this->shouldCaptureOutput = shouldCaptureOutput;
+  this->innerProcessInfo = new ProcessInfo();
+}
+
+bool RedirectedProcess::openPipe() {
+  if (!shouldCaptureOutput) {
+    return true;
+  }
+
+  int pipeHandles[2] = {-1, -1};
+  if (::pipe(pipeHandles) < 0) {
+    return false;
+  }
+
+  innerProcessInfo->readHandle = pipeHandles[0];
+  innerProcessInfo->writeHandle = pipeHandles[1];
+
+  return true;
+}
+
+bool RedirectedProcess::execute(const char *path, bool setGroupFlags,
+                                char *const *args, char *const *envp,
+                                std::mutex &spawnedProcessesMutex) {
+  // Initialize the spawn attributes.
+  posix_spawnattr_t attributes;
+  posix_spawnattr_init(&attributes);
+
+  // Unmask all signals
+  sigset_t noSignals;
+  sigemptyset(&noSignals);
+  posix_spawnattr_setsigmask(&attributes, &noSignals);
+
+// Reset all signals to default behavior.
+//
+// On Linux, this can only be used to reset signals that are legal to
+// modify, so we have to take care about the set we use.
+#if defined(__linux__)
+  sigset_t mostSignals;
+  sigemptyset(&mostSignals);
+  for (int i = 1; i < SIGUNUSED; ++i) {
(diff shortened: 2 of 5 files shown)


hughbe:
@swift-ci please smoke test

hughbe:
@ddunbar friendly ping (for this and the other open Windows PRs)!

hughbe:
@swift-ci please smoke test

hughbe:
@swift-ci please smoke test

include/llbuild/Basic/RedirectedProcess.h
```diff
@@ -0,0 +1,68 @@
+//===- RedirectedProcess.h --------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
```
ddunbar: Nit: 2017

include/llbuild/Basic/RedirectedProcess.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include "llvm/ADT/SmallString.h"
+
+#include <mutex>
+#include <vector>
+
+namespace llbuild {
+namespace basic {
+namespace sys {
+class RedirectedProcess {
```
ddunbar: Can we get some documentation comments here on what the class is for and how it behaves?

Some particular questions I would have reading this header w/o the implementation:
1. Is it legal to execute multiple processes on the same object? If not, why have the flags on `execute()` versus the constructor?
2. What are the return values of things like `kill()` presumably false on failure, but always good to call out? I would also call `kill` something more English like `sendSignal`.
3. What does `sigkill()` do?
4. What is the mutex argument for?

This class should probably be non-copyable right? Is there a reason I would want it to be equatable versus working with pointers to it?

include/llbuild/Basic/RedirectedProcess.h
```diff
@@ -0,0 +1,68 @@
...
+  bool shouldCaptureOutput;
+  struct ProcessInfo *innerProcessInfo;
+
+public:
+  RedirectedProcess(bool shouldCaptureOutput);
+
+  bool openPipe();
+
+  bool execute(const char *path, bool setGroupFlags, char *const *args,
+               char *const *envp, std::mutex &spawnedProcessMutex);
```
ddunbar: I'm worried this is a bit too literal of a port of the existing code and doesn't really abstract enough over the underlying implementation details (exposing the mutex, for example), and requiring raw C strings.

For the mutex, its possible the right abstraction would be one which can manage the entire set of spawned subprocesses -- that might also clean up some of the killing and cancellation logic in the execution queue...

I'm not sure about this, and I'm fine with starting with something less than ideal which is still an improvement and helps with Windows portability...

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+  posix_spawnattr_destroy(&attributes);
+
+  innerProcessInfo->processId = pid;
+
+  return true;
+}
+
+bool RedirectedProcess::readPipe(llvm::SmallString<1024> &output) {
+  if (!shouldCaptureOutput) {
+    return true;
```
ddunbar: I would expect this to be a programatic error (i.e. a fatal error or an assert)

include/llbuild/Basic/RedirectedProcess.h
```diff
@@ -0,0 +1,68 @@
...
+
+  bool readPipe(llvm::SmallString<1024> &output);
```
ddunbar: Nit: `type& name` not `type &name` (we use Google style not LLVM style here)

include/llbuild/Basic/RedirectedProcess.h
ddunbar: `setGroupFlags` would be better named something like `createProcessGroup`

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  // Setup the file actions.
+  posix_spawn_file_actions_t fileActions;
+  posix_spawn_file_actions_init(&fileActions);
+
+  // Open /dev/null as stdin.
+  posix_spawn_file_actions_addopen(&fileActions, 0, "/dev/null", O_RDONLY, 0);
+
+  // Create a pipe to use to read the command output, if necessary.
+  if (shouldCaptureOutput) {
```
ddunbar: This should assert `openPipe` has effectively been called (probably by checking the handle IDs are not some sentinel)

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+    posix_spawn_file_actions_adddup2(&fileActions, 1, 1);
+    posix_spawn_file_actions_adddup2(&fileActions, 2, 2);
+  }
+
+  // Spawn the command.
+  pid_t pid;
+  {
+    // We need to hold the spawn processes lock when we spawn, to ensure that
+    // we don't create a process in between when we are cancelled.
+    std::lock_guard<std::mutex> guard(spawnedProcessesMutex);
```
ddunbar: API simplification: If you make this class only able to launch one process *and* move the initialization out *and* make just one method which does the actual spawn, then the client could hold the lock and not expose it through this API.

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  // Close the write end of the output pipe.
+  ::close(innerProcessInfo->writeHandle);
+
+  // Read all the data from the output pipe.
+  while (true) {
+    char buf[4096];
+    ssize_t numBytes = ::read(innerProcessInfo->readHandle, buf, sizeof(buf));
+    if (numBytes < 0) {
+      return false;
```
ddunbar: We should loop here if `errno == EINTR` (again another problem in existing code)

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+bool RedirectedProcess::waitForCompletion(int *exitStatus) {
+  int result = waitpid(innerProcessInfo->processId, exitStatus, 0);
+  while (result == -1 && errno == EINTR)
+    result = waitpid(innerProcessInfo->processId, exitStatus, 0);
+
+  return result != -1;
+}
+
+bool RedirectedProcess::kill(int signal) {
+  int result = ::kill(-(innerProcessInfo->processId), signal);
```
ddunbar: We attempt to kill the process group here -- which is definitely something we should document. This is also probably wrong if we don't create a process group, but that is another existing bug that maybe we shouldn't try and fix here.

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+size_t RedirectedProcess::hash() const {
+  return innerProcessInfo->processId;
+}
+
+RedirectedProcess::~RedirectedProcess() {
+  // TODO: this causes tens of errors at runtime running the tests:
+  // "*** Error in `/mnt/c/Users/hughb/Documents/GitHub/swift-linux/build/bin/llbuild': double free or corruption (fasttop): 0x00007f1450000940 ***"
+  // Surely we allocated innerProcessInfo (this->innerProcessInfo = new ProcessInfo();), so have
+  // to delete it? This may need investigation to make sure we don't leak memory.
+  // Or I'm probably wrong, and should delete this destructor.
```
ddunbar: This is probably related to the type being copied incorrectly.

As written, if the type is copied (using the compiler synthesized copy constructor or assignment operator) then two types will share the allocated process info.

Note that using `std::unique_ptr` to hold the data is usually the ideal choice; it makes the right thing happen *most* of the time.

lib/Basic/RedirectedProcess.cpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+
+int RedirectedProcess::sigkill() {
+  return SIGKILL;
+}
+
+bool RedirectedProcess::isProcessCancelledStatus(int status) {
+  return WIFSIGNALED(status) &&
+    (WTERMSIG(status) == SIGINT || WTERMSIG(status) == SIGKILL);
```
ddunbar: API simplification: rather than have the client pass back in the status which we previously handed them, why not return a more rich status object, or alternately just store all the status here and (a) assert if it isn't present and (b) avoid forcing the client to pass it back in?

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -76,7 +78,7 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
   bool shutdown { false };
   
   /// The set of spawned processes to terminate if we get cancelled.
-  std::unordered_set<pid_t> spawnedProcesses;
+  std::unordered_set<sys::RedirectedProcess> spawnedProcesses;
```
ddunbar: I would probably just make this a map of pid_t to `std::unique_ptr<sys::RedirectedProcess>`

ddunbar:
Done! Sorry for the long delay, thanks for working on this; it will definitely be nice to get this factored out nicely and maybe with some unit tests to back it up.

hughbe:
Yeah, I'm gonna take a longer look at this and address your comments after #89 is merged, as these two conflict :)

hughbe:
Temporarily closing to avoid confusion. Will reopen when #135 is merged

PR #82, opened 2017-01-23
Make detection of number of CPUs platform independent

hughbe:
Use c++ 11's hardware_concurrency:

See http://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency

Also change the error condition: `hardware_concurrency` returns 0 if it fails

PR #100, opened 2017-01-27
Fix another use of `sysconf` instead of platform independent c++ 11

hughbe:
I missed this in #82 as I hadn't got lib/Commands building at that point


hughbe:
@swift-ci please smoke test

PR #101, opened 2017-01-27
Fixup some MSVC warnings

hughbe:
- Implicit conversion of NULL to char
- unknown escape '\e' that I missed in #83


hughbe:
@swift-ci please smoke test

PR #102, opened 2017-01-27
Fix environ uses on Windows/MSVC

hughbe:
On Windows/MSVC, environ is not an` extern "C"` field in the global namespace.
Instead, it's a macro that returns a pointer.

Therefore, we get errors trying to declare `extern char **environ`, and we also get errors trying to refer to `::environ`. The fix is to remove `::`, which we can also do on non-Windows/MSVC platforms.

> warning C4273: '__p__environ': inconsistent dll linkage
> error C2589: '(': illegal token on right side of '::'
> error C2059: syntax error: '::' 

I have introduced `#ifdefs` around the `extern "C"` declaration of `environ` to fix the first issue mentioned. Can move them to a platform specific file if you'd like.

### DIFF
--- a/lib/BuildSystem/LaneBasedExecutionQueue.cpp
+++ b/lib/BuildSystem/LaneBasedExecutionQueue.cpp
@@ -43,9 +43,11 @@
 using namespace llbuild;
 using namespace llbuild::buildsystem;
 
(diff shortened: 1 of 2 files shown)


hughbe:
@swift-ci please smoke test

PR #109, opened 2017-02-10
[BuildSystem] Switch BuildValue to use BinaryCoding.

ddunbar:
- This cleans up some gross code which was trying to use type punning but then
   having to mask off the non-deterministic values

 - Instead, now we just properly encode only the data we wish to, which reduces
   the encoded size of BuildValues significantly in some cases.

 - On average, this reduces the encoded BuildValue size by 50% for all encodings
   done on our test suite.


ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #111, opened 2017-02-11
[BuildSystem] Add support for task-level build system tests.

ddunbar:
- This exposes a new internal API from the BuildSystem which lets clients
   request the results for arbitrary BuildKeys, which in turn lets us unit test
   specific tasks much more effectively, when combined with the previous work to
   allow us to manually build up an in memory build description.


ddunbar:
@swift-ci please smoke test

PR #112, opened 2017-02-11
[BuildSystem] Don't crash decoding empty values.

ddunbar:
- We don't generally use this, but it is convenient to support empty values as
   a valid sentinel.


ddunbar:
@swift-ci please smoke test

PR #113, opened 2017-02-11
[BuildSystem] Add support for directory signatures.

ddunbar:
- This is another new task type which computes a recursive signature of
   a complete directory tree.

 - This isn't yet as efficient as it absolutely could be, because it
   ends up using both a node and directory contents input which
   duplicate some data, but this is expected to be a fairly heavyweight
   task now anyway.

 - This does not yet properly handle symbolic links, and still isn't
   used by anything real yet...

 - The signature is intentionally constructed so that it will always end up
   accumulating dependency edges on any nodes in the graph which produce the
   contents of the directory, for files which exist. It does *not* attempt to
   scan the entire dependency graph to accumulate edges on nodes which are not
   produced, thus clients are currently required to explicitly ensure some other
   ordering exists between this task and any task which produces output in the
   directory.

If you are curious about the performance, here are some ad hoc
benchmarks taken on a large directory subtree with ~75k nodes in
it. Times are taken on a MacBook Pro with an SSD:

* Initial snapshot, w/o database: 2.2s
* Initial snapshot, w/  database: 4.0s
  * Incremental update          : 4.3s
  * Database size               : 57MB
* `find <path> | wc -l`         : 0.5s

We can do better by avoiding the redundant work and taking more care to
ensure all IO is async, but this seems like a reasonable start.

### DIFF
--- a/include/llbuild/Basic/BinaryCoding.h
+++ b/include/llbuild/Basic/BinaryCoding.h
@@ -76,6 +76,11 @@ public:
     write(uint32_t(value >> 32));
   }
 
+  /// Encode a sequence of bytes to the stream.
+  void writeBytes(StringRef bytes) {
+    data.insert(data.end(), bytes.begin(), bytes.end());
+  }
+
   /// Encode a value to the stream.
   template<typename T>
   void write(T value) {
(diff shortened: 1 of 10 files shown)


ddunbar:
@swift-ci please smoke test

PR #117, opened 2017-02-21
[BuildSystem] Directory tree signatures must include path value.

ddunbar:
- This was an oversight -- normally this isn't required because the values for
   each recursively requested node end up serving as a signature for the
   top-level directory. However, when the top-level item *isn't* a directory,
   but is a missing directory or a file, then this fails.

 - <rdar://problem/30638878> Directory based dependencies need to be behave properly when tracking non-directories


ddunbar:
@swift-ci please test

ddunbar:
@swift-ci please smoke test

PR #119, opened 2017-02-22
[BuildSystem] Directory contents should depend on the input dir node.

ddunbar:
- Although the value itself isn't currently used, this is important for
   ensuring a dependency edge between a possible producer of the directory.


ddunbar:
@swift-ci please smoke test

PR #121, opened 2017-02-23
[libllbuild] Add support for build system reuse.

ddunbar:
- This adds llb_buildsystem_initialize() so that the initialize is clearly
   separated from the build and to make it explicit the system can be reused for
   multiple builds.


ddunbar:
@swift-ci please smoke test

PR #124, opened 2017-02-23
[BuildSystem] Cache the ShellCommand signature.

ddunbar:
- Computing shell command signatures can be expensive; this speeds up null
   builds in cases where the build description is being held in memory.


ddunbar:
@swift-ci please smoke test

PR #125, opened 2017-03-01
[Basic] Introduce a FS method for creating directories.

ddunbar:
- This also fixes a problem with the prior use of the LLVM APIs, which don't
   properly delegate to the umask to set the permissions.


ddunbar:
@swift-ci please smoke test

PR #126, opened 2017-03-01
[BuildSystem] Ensure we create directories recursively.

ddunbar:
- Several callsites were left out of previous refactor to use the new
   FileSystem interface, which meant we were only doing this sometimes. This
   fixes all callsites to use the new interface, and also adds test coverage of
   symlink needing to create directories recursively.


ddunbar:
@swift-ci please smoke test

PR #130, opened 2017-03-18
Fix mkdir usage on Windows

hughbe:
Windows has no equivilent of `S_IRWXU | S_IRWXG |  S_IRWXO`

However, it exposes `_mkdir(1)` in` <direct.h>`: https://msdn.microsoft.com/en-us/library/2fkk4dzw.aspx


ddunbar:
@swift-ci please smoke test

PR #131, opened 2017-03-21
[BuildSystem] Fix bug where directory contents were always recomputed.

ddunbar:
- Reapplied with a fix for the directory signature test to esnure that the
   directory is actually modified (as reported by stat()) before the check which
   requires that.

 - This exposes some infragility in caching the directory contents, at least on
   some systems (this test failed regularly on Ubuntu 16.05), but at least for
   now we still rely on that ability for better or worse.

 - <rdar://problem/30640443> Directory contents nodes are always
   rebuilding


ddunbar:
@swift-ci please smoke test

PR #132, opened 2017-03-22
[tests] Add explicit test for synthetic edges on directory tree nodes.

ddunbar:
- Also, add a doc note on this behavior.


ddunbar:
@swift-ci please smoke test

ddunbar:
@shahmishal something odd changed here which caused llbuild to use GCC instead of Clang, did something about the machine environment change?

ddunbar:

PR #136, opened 2017-04-09
[BuildSystem] Fix a missing signature check for symlink commands.

ddunbar:
- <rdar://problem/31522690> Missing rebuild of symlink node if target changes


ddunbar:
@swift-ci please smoke test

PR #137, opened 2017-04-10
[Core] Insert discovered deps at the appropriate point.

ddunbar:
- This was just an inefficient oversigh, the discovered dependencies should
   always be appended to the result deps list.

 - <rdar://problem/31537299> Discovered dependencies aren't in the appropriate order


ddunbar:
@swift-ci please smoke test

PR #138, opened 2017-04-10
[Core] Ensure database maintains the relative order of dependencies.

ddunbar:
- As currently modeled, the engine only scans dependencies in serial order
   because we cannot yet handle the situation where a dependency has been
   inferred to be out-of-date, but it isn't yet at the head of the queue of
   things to actually begin to run in order to bring another node up-to-date.

 - Given that, it is very important that we do the scanning of dependencies in
   the order that they were declared, since there may be an implicit ordering
   relationship between the impact of evaluating those nodes. See the test case
   for an example (which also serves as a test case for the previous commit).

 - <rdar://problem/31490515> Serial dependency scanning is out-of-order

### DIFF
--- a/include/llbuild/Core/BuildDB.h
+++ b/include/llbuild/Core/BuildDB.h
@@ -61,6 +61,9 @@ public:
   /// unique. However, duplicate dependencies have no semantic meaning for the
   /// engine, and the database may elect to discard them from storage.
   ///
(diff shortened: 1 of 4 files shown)


ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #146, opened 2017-05-10
[BuildSystem] Fix a race between build() and cancel()

aciidb0mb3r:
This fixes a race which occurs when delegate is being initialized and
clients call cancel().

This also adds asserts that clients call build() before cancel().

-- <rdar://problem/31951869>

### DIFF
--- a/lib/BuildSystem/BuildSystemFrontend.cpp
+++ b/lib/BuildSystem/BuildSystemFrontend.cpp
@@ -232,6 +232,15 @@ public:
 };
 
 struct BuildSystemFrontendDelegateImpl {
+
+  /// The status of delegate.
+  enum class Status {
+    Uninitialized = 0,
+    Initialized,
+    InitializationFailure,
+    Cancelled,
+  };
+
   llvm::SourceMgr& sourceMgr;
   const BuildSystemInvocation& invocation;
   
@@ -248,6 +257,28 @@ struct BuildSystemFrontendDelegateImpl {
                                   const BuildSystemInvocation& invocation)
       : sourceMgr(sourceMgr), invocation(invocation),
         executionQueueDelegate(*this) {}
+
+private:
+  /// The status of the delegate.
+  std::atomic<Status> status{Status::Uninitialized};
+
+public:
+
+  /// Set the status of delegate to the given value.
+  ///
+  /// It is not possible to update the status once status is set to initialization failure.
+  void setStatus(Status newStatus) {
+    // Disallow changing status if there was an initialization failure.
+    if (status == Status::InitializationFailure) {
+      return;
+    }
+    status = newStatus;
+  }
+
+  /// Returns the current status.
+  Status getStatus() {
+    return status;
+  }
 };
 
 bool BuildSystemFrontendExecutionQueueDelegate::showVerboseOutput() const {
@@ -378,6 +409,11 @@ BuildSystemFrontendDelegate::createExecutionQueue() {
 
 void BuildSystemFrontendDelegate::cancel() {
   auto delegateImpl = static_cast<BuildSystemFrontendDelegateImpl*>(impl);
+  assert(delegateImpl->getStatus() != BuildSystemFrontendDelegateImpl::Status::Uninitialized);
+
+  // Update the status to cancelled.
+  delegateImpl->setStatus(BuildSystemFrontendDelegateImpl::Status::Cancelled);
+
   auto system = delegateImpl->system;
   if (system) {
     system->cancel();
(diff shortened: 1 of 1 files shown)


lib/BuildSystem/BuildSystemFrontend.cpp
```diff
@@ -232,12 +232,21 @@ class BuildSystemFrontendExecutionQueueDelegate
 };
 
 struct BuildSystemFrontendDelegateImpl {
+
+  /// The status of delegate.
+  enum class Status {
+    Uninitialized = 0,
```
ddunbar: This term is a little confusing given `initialize` -- the system can be "initialized" w.r.t. `initialize()`, but the Status be `Uninitialized` which is a little odd.

aciidb0mb3r: What about `Uninitialized`, `Initialized` and `Cancelled`?

ddunbar: Sounds good

ddunbar: Actually, we should also have "FailedToInitialize" or something, so that we can report an error if initialization fails but the client tries to build anyway.

aciidb0mb3r: Don't think we need a state for that, we already fail and return errors in case of failed initialization.
https://github.com/apple/swift-llbuild/blob/master/lib/BuildSystem/BuildSystemFrontend.cpp#L546

ddunbar: What if a client tries to build after the first initialization fails, though?

aciidb0mb3r: debug mode: assertional failure
release mode: not sure, will this still return false? https://github.com/apple/swift-llbuild/blob/master/lib/BuildSystem/BuildSystemFrontend.cpp#L544

ddunbar: No, because of:
https://github.com/apple/swift-llbuild/blob/master/lib/BuildSystem/BuildSystemFrontend.cpp#L496

lib/BuildSystem/BuildSystemFrontend.cpp
```diff
@@ -378,6 +409,11 @@ BuildSystemFrontendDelegate::createExecutionQueue() {
 
 void BuildSystemFrontendDelegate::cancel() {
   auto delegateImpl = static_cast<BuildSystemFrontendDelegateImpl*>(impl);
+  assert(delegateImpl->getStatus() == BuildSystemFrontendDelegateImpl::Status::Initialized);
```
ddunbar: I think it is probably legal to call this with IntializationFailed, too, because client won't necessarily know at the time they call.

aciidb0mb3r: Fixed

lib/BuildSystem/BuildSystemFrontend.cpp
```diff
@@ -388,6 +424,8 @@ void BuildSystemFrontendDelegate::resetForBuild() {
   auto impl = static_cast<BuildSystemFrontendDelegateImpl*>(this->impl);
 
   impl->numFailedCommands = 0;
+  impl->numErrors = 0;
+  impl->setStatus(BuildSystemFrontendDelegateImpl::Status::Uninitialized);
```
ddunbar: I don't think this is right... this will be the wrong value if the status has already been Initialized.

I think all this should do is reset to Initialized if the status is Cancelled.

lib/BuildSystem/BuildSystemFrontend.cpp
```diff
@@ -540,9 +578,27 @@ bool BuildSystemFrontend::initialize() {
 }
 
 bool BuildSystemFrontend::build(StringRef targetToBuild) {
+
+  auto delegateImpl =
+    static_cast<BuildSystemFrontendDelegateImpl*>(delegate.impl);
+
+  // Delegate should be in uninitialized state right now.
+  assert(delegateImpl->getStatus() == BuildSystemFrontendDelegateImpl::Status::Uninitialized);
```
ddunbar: This isn't right, the status could also be Initialized.

ddunbar: (and the one C API test should test that).

ddunbar:
@sw

ddunbar:
@swift-ci please smoke test

aciidb0mb3r:
@swift-ci please smoke test

aciidb0mb3r:

PR #149, opened 2017-05-10
[BuildSystem] Fix yet another cancellation race

aciidb0mb3r:
<rdar://problem/32126388> Race in BuildSystemImpl

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -45,6 +45,7 @@
 #include "llvm/Support/raw_ostream.h"
 
 #include <memory>
+#include <mutex>
 
 #include <unistd.h>
 
@@ -158,6 +159,9 @@ class BuildSystemImpl : public BuildSystemCommandInterface {
 
   /// The build engine.
   BuildEngine buildEngine;
+
+  /// Mutex for access to execution queue.
+  std::mutex executionQueueMutex;
 
   /// The execution queue reference; this is only valid while a build is
   /// actually in progress.
@@ -302,10 +306,19 @@ public:
     buildWasAborted = value;
   }
 
+  void resetForBuild() {
+    std::lock_guard<std::mutex> guard(executionQueueMutex);
+    isCancelled_ = false;
+  }
+
   /// Cancel the running build.
   void cancel() {
+    std::lock_guard<std::mutex> guard(executionQueueMutex);
+
     isCancelled_ = true;
-    getExecutionQueue().cancelAllJobs();
+    // Cancel jobs if we actually have a queue.
+    if (executionQueue.get() != nullptr)
+      getExecutionQueue().cancelAllJobs();
   }
 
   /// Check if the build has been cancelled.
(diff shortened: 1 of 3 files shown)


lib/BuildSystem/BuildSystem.cpp
```diff
@@ -1203,6 +1205,13 @@ BuildSystemImpl::lookupNode(StringRef name, bool isImplicit) {
 }
 
 llvm::Optional<BuildValue> BuildSystemImpl::build(BuildKey key) {
+
+  // If we were cancelled, return.
+  if (isCancelled_) {
+      isCancelled_ = false;
```
aciidb0mb3r: I am not sure if resetting cancel like this is correct.

aciidb0mb3r: Maybe BuildSystem should have its own reset method and frontend delegate should call that if it has a build system?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -305,7 +305,9 @@ class BuildSystemImpl : public BuildSystemCommandInterface {
   /// Cancel the running build.
   void cancel() {
     isCancelled_ = true;
-    getExecutionQueue().cancelAllJobs();
+    // Cancel jobs if we actually have a queue.
+    if (executionQueue.get() != nullptr)
```
aciidb0mb3r: I think it is not thread safe to check like this?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -1203,6 +1205,13 @@ BuildSystemImpl::lookupNode(StringRef name, bool isImplicit) {
...
 llvm::Optional<BuildValue> BuildSystemImpl::build(BuildKey key) {
+
+  // If we were cancelled, return.
+  if (isCancelled_) {
+      isCancelled_ = false;
+      return None;
+  }
+
   // Create the execution queue.
   executionQueue = delegate.createExecutionQueue();
```
aciidb0mb3r: Do we need an atomic variable here to signify that execution queue was successfully created ?

aciidb0mb3r: May be we can combine these and have a status enum here. `{Started, QueueCreated, Aborted, Cancelled}` but not sure what are the exact semantics of BuildSystem.

lib/BuildSystem/BuildSystem.cpp
ddunbar: Correct (it is not thread safe), we need some kind of lock.

lib/BuildSystem/BuildSystem.cpp
ddunbar: I agree, that seems cleanest.

lib/BuildSystem/BuildSystem.cpp
ddunbar: You could start with just a shared lock to protect the cancellation flag and executionQueue.

aciidb0mb3r:
@ddunbar updated

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -302,14 +306,24 @@ class BuildSystemImpl : public BuildSystemCommandInterface {
...
-    getExecutionQueue().cancelAllJobs();
+    // Cancel jobs if we actually have a queue.
+    if (executionQueue.get() != nullptr)
+      getExecutionQueue().cancelAllJobs();
   }
 
   /// Check if the build has been cancelled.
   bool isCancelled() {
+    std::lock_guard<std::mutex> guard(stateMutex);
     return isCancelled_;
```
ddunbar: It would be nice to avoid this lock, since we check this in many places inside the system. I think we can safely do so by making isCancelled_ atomic, we only actually need the lock for the ordering w.r.t. the executionQueue.

aciidb0mb3r: Oh, I thought atomic and lock will be similar in terms of perf.

aciidb0mb3r:
@swift-ci please smoke test

aciidb0mb3r:
@swift-ci please smoke test

PR #150, opened 2017-05-11
Optimize the rule dependency search query.

ddunbar:
- This adds an additional covering index for the rule dependency query, which
   is important for the null build case where our work largely amounts to
   populating and scanning the results.

 - This fixes a regression introduced in
   bcc3d1011eff6450d74f9b0bec913d17bf48efa2 -- the table primary key index was
   previously covering, but now is not because of the required ordinal
   parameter.

 - This change is unfortunate as there is a lot of redundancy in the tables, and
   this increases database size by 25% in one example, but it is also good for a
   ~10% win in null build time on a large graph.

 - <rdar://problem/32127527> Investigate optimizing rule dependency w/ ordinal lookup

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -193,8 +193,21 @@ public:
         // Create an index to be used for efficiently looking up rule
         // information from a key.
         result = sqlite3_exec(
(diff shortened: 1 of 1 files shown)


ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #152, opened 2017-05-11
Intern build description strings.

ddunbar:
- It is common to have many duplicated strings in a loaded build description
   file (typically from command line arguments), so this patch optimizes the
   memory use of the build description by interning these strings while loading.

 - While we pay some cost for the interning itself, on the other hand this
   allows avoiding what would otherwise be additional allocations in many cases,
   so the interning ends up paying for itself while reducing memory footprint &
   improving cache locality. On the balance my tests indicate this is a wash to
   small improvement in terms of CPU use.

 - <rdar://problem/32128433> Intern build description strings

### DIFF
--- a/include/llbuild/BuildSystem/BuildFile.h
+++ b/include/llbuild/BuildSystem/BuildFile.h
@@ -65,6 +65,8 @@ struct ConfigureContext {
   BuildFileToken at;
 
 public:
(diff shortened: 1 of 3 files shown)


ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #154, opened 2017-05-11
Use a StringMap for BuildDescription data.

ddunbar:


ddunbar:
@swift-ci please smoke test

PR #161, opened 2017-05-13
[Core/DB] Denormalize rule dependencies.

ddunbar:
- Now that the rule dependencies are stored directly by key ID, we can
   denormalize the `rule_result` table to directly include the results (as a
   binary blob). This results in a substantially more compact database and
   efficient access to the rule result (given that we always rewrite rules
   wholesale).

 - This also enables future optimizations to avoid the result re-insert.

 - <rdar://problem/32180158> Denormalize rule_result table to include dependencies


ddunbar:
@swift-ci please smoke test

PR #168, opened 2017-05-18
[BuildSystem] Symlink commands should expose target file information.

ddunbar:
- When a node was depending on a symlink, we were ending up depending on 
   the symlink command because it is the producer of that node. However, that
   is broken because the command is typically depending on the *contents* of
   the file, and the producer was only tracking the lstat() information as
   part of its build value.

 - Instead, change the symlink command to vend the target file information (i.e.
   stat()) as its build result, and only track the lstat() information internally
   for the purposes of knowing when the command needs to rerun.

 - <rdar://problem/31455060> Missing command execution when link target changes

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -1936,6 +1936,27 @@ public:
 
 #pragma mark - SymlinkTool implementation
 
+/// Implementation of the symlink command.
+///
+/// This command uses a non-orthodox encoding of its data. In order to
+/// accurately model the semantics of symbolic links, we need to track both the
+/// stat() and lstat() information of the output path:
+///
+/// * We need to track the stat() information because this command will be the
+///   producer of the file output, and commands which depend on that file should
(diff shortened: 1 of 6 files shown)


ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2109,6 +2161,21 @@ class SymlinkCommand : public Command {
       return BuildValue::makeFailedCommand();
     }
 
+    // Check if we actually needed to run, or just update the result value.
+    fprintf(stderr, "rerunning symlink task, has prior value? (%s)\n",
```
m-i-r-z-a: Did you mean to leave this in?

ddunbar: Hah, thanks definitely not

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #177, opened 2017-05-25
[BuildSystem] Add an experimental feature to expose both views of a s…

ddunbar:
…ymlink.

 - There was previously no way to safely see the `stat()`-level consistency of a
   symlink target, this adds an experimental (albeit very clunky) way to do so
   by declaring a separate argument to define the link output path from the
   declared output node.


ddunbar:
@swift-ci please smoke test

PR #184, opened 2017-06-15
Fix Windows implementation of symlink

hughbe:
`_symlink` doesn't actually exist on Windows, so this fails compilation

Also, we need to pass a different flag for files vs directories

@neonichu


hughbe:
@swift-ci please smoke test

PR #185, opened 2017-06-17
[LLVM] Add LLVM import for Windows.

ddunbar:


ddunbar:
This is just an import of the .inc files circa when the original LLVM import was done. We still need to clean up the import process...

ddunbar:
@swift-ci please smoke test

ddunbar:
@hughbe Well, we *should* update and come up with a well defined process for managing this -- obviously doing a manual custom import of a small slice of the code is not at all ideal. It's been hard to justify spending the time on the cleanup though, given that it doesn't really cause us harm at the moment.

PR #190, opened 2017-10-26
Export LLBUILD_TASK_ID to subprocesses.

ddunbar:
[BuildSystem] Export LLBUILD_TASK_ID to subprocesses.
 - This will eventually be used to propagate a unique ID which subprocesses can
   use for reassociation, but for now it is just the lane ID to use for tracing
   purposes.

 - <rdar://problem/35179142> Pass LLBUILD_TASK_ID environment variable to Clang for tracing purposes


ddunbar:
@swift-ci please smoke test

PR #193, opened 2017-11-05
[Core] Fix a race condition in task cancellation.

ddunbar:
- This also renames the cancellation method to be more obvious about
   when it is used.

 - When task cancellation was done as part of an error handling path,
   there was a potential race with any outstanding work being completed.
   We solve this by forcing (engine-level) cancellation to wait for any
   outstanding work before completing (to avoid needing to introduce
   any additional synchronization logic).

 - This is currently very improbable, since errors aren’t ever expected,
   but that will change once we expose engine cancellation to clients.


ddunbar:
@swift-ci please smoke test

PR #196, opened 2017-11-07
[BuildSystem] Don't propagate cancel state through phony commands.

ddunbar:
- When using phony commands to enforce ordering constraints, the cancelled
   state would incorrectly propagate through the phony command, thus causing
   unnecessary rebuilds of all downstream tasks. Instead, we

 - A more fundamental solution to this problem will be to adopt support for
   engine cancellation, and to ultimately allow individual commands to fail
   without reporting a result value (and then triggering engine termination).

 - <rdar://problem/35319783> Unnecessary rebuilds after an error


ddunbar:
@swift-ci please smoke test

PR #197, opened 2017-11-09
[BuildSystem] Fix symlink output value if the output is ever missing.

ddunbar:
- This shouldn't normally be possible, since the command would have failed, but
   it can happen in corner cases when trying to use a simulated filesystem.

 - <rdar://problem/35403456> Crash in llbuild::buildsystem::BuildValue::makeExistingInput()


ddunbar:
@swift-ci please smoke test

ddunbar:

PR #199, opened 2017-11-17
[Linux] Use SIGSYS over SIGUNUSED.

ddunbar:
- SIGUNUSED is removed in newer versions of glibc.

 - https://bugs.swift.org/browse/SR-6409


ddunbar:
@swift-ci please smoke test

PR #205, opened 2017-12-08
[CMake] install llbuild executable

theodelrieu:
Hello,

From my understanding of the README, one should be able to use `llbuild` after using CMake to build and install the project.

However, there was no `install` command for the `llbuild` executable.
If it is not meant to be installed, feel free to close this PR!

### DIFF
--- a/products/llbuild/CMakeLists.txt
+++ b/products/llbuild/CMakeLists.txt
@@ -13,3 +13,14 @@ target_link_libraries(llbuild
 if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
   target_link_libraries(llbuild curses)
 endif()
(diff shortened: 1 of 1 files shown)


theodelrieu:
@swift-ci please smoke test

ddunbar:
llbuild (the project) is primarily intended to be used as a library, and `llbuild` (the tool) is primarily intended to be used for own tests.

It is useful as an alternative to `ninja`, but I'm not sure this is valuable enough to motivate installing it.

Unless you have a strong reason to want the tool, I'd recommend we continue not installing it.

theodelrieu:
No problem, I was fooling around to make Ninja work with Swift, but went another direction instead.

PR #206, opened 2017-12-08
[tests/Ninja/Loader] Make shell-quote test more robust

rdwampler:
The correct check for the "depfile" line should only contain single quotes
around the variable portion. I.e., 'target12 foo'.d not 'target12 foo.d'. To
avoid future fragility, change the `CHECK-NOT` to an explicit `CHECK`.

Sorry for the noise!


ddunbar:
@swift-ci please smoke test

PR #208, opened 2018-01-07
[BuildSystem] Add support for a directory structure tracking node.

ddunbar:
- This is like the existing node which tracks a directory (recursively), but it
   only considers the structural elements of the directory (the files and their
   modes).

### DIFF
--- a/tests/BuildSystem/Build/directory-tree-structure-signatures.llbuild
+++ b/tests/BuildSystem/Build/directory-tree-structure-signatures.llbuild
@@ -1,4 +1,4 @@
-# Check the handling of directory tree signatures.
+# Check the handling of directory tree structure signatures.
 #
 # We check the node used against three kinds of directories: missing, a file, an
 # actual directory.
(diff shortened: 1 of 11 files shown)


ddunbar:
@swift-ci please smoke test

ddunbar:
@aciidb0mb3r this is the node we talked about...

To use it, create a node like this:
```
nodes:
 "foo/":
    is-directory-structure: true
```
and then create a command with that as an input. It should run every time the structure changes.

Note that the current implementation actually does traverse symbolic links; I view this as a bug (it should be configurable, and off by default), so I would recommend still looking at putting in the SwiftPM logic to add explicit dependencies on the targets of any symbolic links. However, it isn't strictly necessary given the current implementation (and honestly I expect this to be totally sufficient, it is just a shame it will blow up if someone creates a symbolic link to some giant directory).

tests/BuildSystem/Build/directory-tree-structure-signatures.llbuild
```diff
@@ -0,0 +1,100 @@
...
+targets:
+  "": ["<all>"]
+
+nodes:
+  "missing-dir/":
+    is-directory-structure: true
+  "file/":
+    is-directory-structure: true
+  "dir/":
+    is-directory-structure: true
```
aciidb0mb3r: Random question: What if I also need `dir/` as a node with is-directory-structure = false?

aciidb0mb3r:
Awesome, thanks!

tests/BuildSystem/Build/directory-tree-structure-signatures.llbuild
ddunbar: That may not work out currently... we don't yet have a way to override the target path independent of the name, so unfortunately it has to be mangled this way. I could probably make it work to use "dir//" as a workaround.

PR #216, opened 2018-01-23
[BuildSystem] Validate target name before building it

aciidb0mb3r:
<rdar://problem/31675430> swift-build-tool crashes when wrong target name is passed


aciidb0mb3r:
@swift-ci please smoke test

PR #220, opened 2018-01-25
`posix_spawn` does not set `errno`

neonichu:
Instead we should get possible errors from the return value.


neonichu:
@swift-ci please smoke test

PR #224, opened 2018-02-10
[Swift Bindings] Re-add CMake support for building Swift bindings.

aciidb0mb3r:
In order to use llbuild's Swift API in SwiftPM, we need to build the
bindings as part of Swift's build script. This is the first step in that
direction.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,12 @@ else()
   set(cmake_3_2_USES_TERMINAL USES_TERMINAL)
 endif()
 
+# Configure the default set of bindings to build.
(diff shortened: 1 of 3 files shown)


ddunbar:
I guess I had a wild dream we could use the Swift package support in llbuild itself. I guess that is a little tricky to pull off if we are trying to bootstrap?

I really wish we had moved a model where we always built starting from a stable toolchain, but that is probably too much work to try and tackle as part of this effort.

ddunbar:
@swift-ci please smoke test

aciidb0mb3r:
Yah, maybe someday when the package manager can handle more complex scenarios.

PR #225, opened 2018-02-11
Install libllbuild swift

aciidb0mb3r:


aciidb0mb3r:
Depends on https://github.com/apple/swift-llbuild/pull/224

products/llbuildSwift/CMakeLists.txt
```diff
@@ -5,8 +5,38 @@ set(SOURCES
...
 list(APPEND additional_args -I ${CMAKE_CURRENT_SOURCE_DIR}/../libllbuild/include -lllbuild)
+list(APPEND additional_args -target x86_64-apple-macosx10.10)
 
 # Add swift bindings target if swift compiler is present.
 if (SWIFTC_FOUND)
   add_swift_module(libllbuildSwift llbuildSwift libllbuild "${SOURCES}" "${additional_args}")
+
+  # Install libllbuild.
+  install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/../libllbuild/include
+    DESTINATION lib/swift/pm/llbuild
```
ddunbar: This is ok, but in an ideal world would be a little more decoupled from Swift, and we would just expose the option and a build setting which defines the install location, and the Swift build script would configure it appropriately.

I don't feel strongly enough about it to advocate fixing it for this effort though.

aciidb0mb3r: Yeah, maybe I can rename the target name to `install-swift-toolchain-libllbuildSwift`.

ddunbar: meh, I'd rather keep it as is and make the path configurable later.

aciidb0mb3r:
@swift-ci please smoke test

aciidb0mb3r:
@swift-ci please smoke test

PR #228, opened 2018-02-16
Added guard so including macOS target is limited to an Apple platform

tachoknight:
Including the x86_64-apple-macosx10.10 target causes a build failure on Linux. I added a guard around the line so it only gets included when building on an Apple (i.e. macOS) platform.


aciidb0mb3r:
Oops, thanks!

aciidb0mb3r:
@swift-ci please smoke test

aciidb0mb3r:
We should get the bindings building on Linux CI too..

PR #237, opened 2018-03-08
Report extended results in commandProcessFinished

dmbryson:
Collect user/system time and max RSS values for subprocesses and report
them via a new CommamdExtendedResults struct in the
commandProcessFinished callback.

rdar://problem/37889042

### DIFF
--- a/examples/c-api/buildsystem/main.c
+++ b/examples/c-api/buildsystem/main.c
@@ -179,8 +179,7 @@ static void command_process_had_output(void* context,
 static void command_process_finished(void* context,
                                      llb_buildsystem_command_t* command,
                                      llb_buildsystem_process_t* process,
-                                     llb_buildsystem_command_result_t result,
-                                     int exit_status) {
+                                     const llb_buildsystem_command_extended_result_t* result) {
 }
 
 int main(int argc, char **argv) {
--- a/include/llbuild/BuildSystem/CommandResult.h
+++ b/include/llbuild/BuildSystem/CommandResult.h
@@ -13,6 +13,8 @@
 #ifndef LLBUILD_BUILDSYSTEM_COMMAND_RESULT_H
 #define LLBUILD_BUILDSYSTEM_COMMAND_RESULT_H
 
+#include <inttypes.h>
+
 namespace llbuild {
 namespace buildsystem {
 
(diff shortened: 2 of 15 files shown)


dmbryson:
@swift-ci please smoke test

include/llbuild/BuildSystem/CommandResult.h
```diff
@@ -24,6 +26,32 @@ enum class CommandResult {
...
+  uint64_t maxrss;      /// Max RSS (in bytes)
+
+
+  CommandExtendedResult(CommandResult result, int exitStatus, uint64_t utime = 0,
+                uint64_t stime = 0, uint64_t maxrss = 0)
+    : result(result), exitStatus(exitStatus)
+    , utime(utime), stime(stime), maxrss(maxrss)
+  {}
+
+  static CommandExtendedResult failed(int exitStatus = -1) {
```
ddunbar: We have a "mini convention" in LLVM of calling struct factory functions something like `makeFailed` which I kinda prefer, unless you have an objection

examples/c-api/buildsystem/main.c
```diff
@@ -179,8 +179,7 @@ static void command_process_had_output(void* context,
 static void command_process_finished(void* context,
                                      llb_buildsystem_command_t* command,
                                      llb_buildsystem_process_t* process,
-                                     llb_buildsystem_command_result_t result,
-                                     int exit_status) {
+                                     llb_buildsystem_command_extended_result_t result) {
```
ddunbar: I think we want to pass this as a pointer. It might be more efficient, but more importantly it means we can extend the struct without changing the ABI.

llbuild.xcodeproj/xcshareddata/xcschemes/llbuild-all.xcscheme
```diff
@@ -0,0 +1,187 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ddunbar: Should we just roll this into the regular llbuild scheme? I'm not sure there is a downside. It might be worth landing that separately, though?

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -220,6 +220,23 @@ public enum CommandResult {
   }
 }
 
+/// Result of a command execution.
+public struct CommandExtendedResult {
+  public let result : CommandResult /// The result of a command execution
```
ddunbar: By convention, Swift has no `:` after the variable name, so `let result: CommandResult`

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -220,6 +220,23 @@ public enum CommandResult {
...
+  public let exitStatus : Int32     /// The exit code
+  public let utime : UInt64         /// User time (in us)
+  public let stime : UInt64         /// Sys time (in us)
+  public let maxrss : UInt64        /// Max RSS (in bytes)
```
ddunbar: I would call this `maxRSS` in more pedantic Swift convention style.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #243, opened 2018-03-14
Fix Release build preprocessor definitions

dmbryson:


dmbryson:
@swift-ci please smoke test

ddunbar:
Can you explain, I can’t tell the issue just from the diff

dmbryson:
Ah, yes. Release.xcconfig sets GCC_PREPROCESSOR_DEFINITIONS to include $(COMMON_PREPROCESSOR_DEFINITIONS) and NDEBUG=1.  The project level and several targets were overriding these without also using $(inherited).   This change ensures that all targets are inheriting COMMON_PREPROCESSOR_DEFINITIONS as expected.

The bug caused by the lack of these definitions is that the PerfTest bot fails to build libllbuild.

ddunbar:
Got it, thanks!

PR #245, opened 2018-03-15
SR-2100 - Add support for multi-lib systems

capent:
[SR-2100 - Add support for multi-lib systems](https://bugs.swift.org/browse/SR-2100)

Use a newly create cmake variable SWIFT_LIBDIR_SUFFIX to create a suffix for the /lib dir.
If the variable is empty or not set, the /lib will be created without a suffix.

It's the same suffix system used by LLVM project with LLVM_LIBDIR_SUFFIX.


CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ set(LLBUILD_OBJ_DIR ${CMAKE_CURRENT_BINARY_DIR})
 
 # Shared output directories for executables and libraries.
 set(LLBUILD_EXECUTABLE_OUTPUT_INTDIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/bin)
-set(LLBUILD_LIBRARY_OUTPUT_INTDIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/lib)
+set(LLBUILD_LIBRARY_OUTPUT_INTDIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/lib${SWIFT_LIBDIR_SUFFIX})
```
aciidb0mb3r: I think this should be `LLBUILD_LIBDIR_SUFFIX`.

capent: Ok, just did it.

CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ set(LLBUILD_OBJ_DIR ${CMAKE_CURRENT_BINARY_DIR})
 
 # Shared output directories for executables and libraries.
 set(LLBUILD_EXECUTABLE_OUTPUT_INTDIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/bin)
-set(LLBUILD_LIBRARY_OUTPUT_INTDIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/lib)
+set(LLBUILD_LIBRARY_OUTPUT_INTDIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/lib${LLBUILD_LIBDIR_SUFFIX})
```
aciidb0mb3r: Do we need a default value? What happens when this variable is not set?

capent: On cmake, a variable reference is replaced by the value of the variable, or by the empty string if the variable is not set. So i don't think that we need a default value since the default will be empty (no suffix).

I planned this value to be defined during cmake invocation if needed to be set, using the "-D" argument, as in -DLLBUILD_LIBDIR_SUFFIX:STRING="64", since this is the way that [build-script-impl](https://github.com/capent/swift/blob/master/utils/build-script-impl) script in swift repository deal with cmake arguments (like on line 1922 onward).

ddunbar:
This LGTM, but can you rebase to clean up the merge commits?

CMakeLists.txt
aciidb0mb3r: Sounds good!

capent:
Ok, i rebased it.

CMakeLists.txt
dmbryson: It would be nice to have LLBUILD_LIBDIR_SUFFIX in a documented form. That way it is more discoverable and not a magic variable that just needs to be known.

`set(LLBUILD_LIBDIR_SUFFIX "" CACHE STRING "... explanation for what and why this might be set ...")`

dmbryson:
@swift-ci please smoke test

capent:
@dmbryson Agreed. I will document the LLBUILD_LIBDIR_SUFFIX variable.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #246, opened 2018-03-15
Allow Generate-version.h to work with paths that contain spaces

dmbryson:
rdar://problem/38480932


dmbryson:
@swift-ci please smoke test

PR #247, opened 2018-03-15
[llbuildSwift] Update bindings

aciidb0mb3r:


products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -301,8 +301,28 @@ public struct BuildKey {
     }
 }
 
+public protocol FileInfo {
```
ddunbar: Can we get documentation comments on these protocols?

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -456,6 +476,11 @@ public final class BuildSystem {
         // Construct the system delegate.
         var _delegate = llb_buildsystem_delegate_t()
         _delegate.context = Unmanaged.passUnretained(self).toOpaque()
+        if delegate.fs != nil {
+            _delegate.fs_get_file_contents = { BuildSystem.toSystem($0!).fsGetFileContents(String(cString: $1!), $2!) }
+            _delegate.fs_get_file_info = { BuildSystem.toSystem($0!).fsGetFileInfo(String(cString: $1!), $2!) }
+            _delegate.fs_get_link_info = { BuildSystem.toSystem($0!).fsGetFileInfo(String(cString: $1!), $2!) }
```
ddunbar: This should be a separate callback, not shared with getFileInfo (or get FileInfo should take a parameter)

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -488,12 +513,20 @@ public final class BuildSystem {
...
-    public func build() -> Bool {
-        var data = llb_data_t(length: 0, data: nil)
-        return llb_buildsystem_build(_system, &data)
+    ///
+    /// - parameter node: Optional path to a single node to build.
+    /// - returns: True if the build was successful, false otherwise.
+    public func build(node: String? = nil) -> Bool {
+        if let node = node {
+            var data = copiedDataFromBytes([UInt8](node.utf8))
+            return llb_buildsystem_build_node(_system, &data)
```
ddunbar: Can this be simplified to `copiedDataFromBytes(node.map{ [UInt8]($0.utf8) } ?? [])` and eliminate the if?

aciidb0mb3r: Don't think so. Its calling a different method based on presence of a node.

products/llbuildSwift/BuildSystemBindings.swift
aciidb0mb3r: Added as FIXME for now.

products/llbuildSwift/BuildSystemBindings.swift
aciidb0mb3r: Done.

aciidb0mb3r:
Merging without CI because we don't really build bindings on CI right now.

PR #250, opened 2018-03-15
Fix build breakage introduced by af8a897

jakepetroules:


ddunbar:
Bypassing swift-ci here, since it doesn't test the Xcode project.

PR #252, opened 2018-03-15
Make Command::shouldShowStatus available in C and Swift

hartbit:
Can someone help me understand how to build `llbuildSwift` with CMake? I know I can built `llbuildSwift` with SwiftPM, but it isn't working currently because of the recent changes to version.h

### DIFF
--- a/products/libllbuild/BuildSystem-C-API.cpp
+++ b/products/libllbuild/BuildSystem-C-API.cpp
@@ -676,6 +676,12 @@ void llb_buildsystem_command_get_name(llb_buildsystem_command_t* command_p,
   key_out->data = (const uint8_t*) name.data();
 }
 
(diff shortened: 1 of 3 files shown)


aciidb0mb3r:
I use:
`cmake -G Ninja -DCMAKE_BUILD_TYPE:STRING=Debug -DLLBUILD_SUPPORT_BINDINGS:=Swift ..`

aciidb0mb3r:
We can never too many build systems!

hartbit:
Ok, I got it building, if someone wants to review :-)

hartbit:
@swift-ci please smoke test

ddunbar:
This seems reasonable, although we might want to eventually remove this logic from the core and make it all managed by the client

PR #254, opened 2018-03-20
Generate version.h in place

dmbryson:
Simplifies usage of generated header with SwiftPM

### DIFF
--- a/utils/generate-version-h.sh
+++ b/utils/generate-version-h.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# This script generates the version.h file.  It exists as a separate build
+# phase so that both the framework build and tests that import it
+# 'library-style' can depend on it existing
+
+set -e
+
+# If not in Xcode, infer SRCROOT relative to this script
+if [ -z ${SRCROOT+x} ]; then
+    scriptroot=$(dirname ${BASH_SOURCE[0]})
+    SRCROOT="${scriptroot}/.."
+fi;
+
+# Read the version out of the xcconfig file
+if [ -z ${LLBUILD_C_API_VERSION+x} ]; then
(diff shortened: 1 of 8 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:

utils/generate-version-h.sh
```diff
@@ -0,0 +1,24 @@
...
+# Read the version out of the xcconfig file
+if [ -z ${LLBUILD_C_API_VERSION+x} ]; then
+    LLBUILD_C_API_VERSION=$(sed -n -e "s/.*LLBUILD_C_API_VERSION = \([0-9][0-9]*\).*/\1/p" "${SRCROOT}/Xcode/Configs/Version.xcconfig")
+fi
+
+# Write out version.h from the template
+sed -e "s/\${LLBUILD_C_API_VERSION}/${LLBUILD_C_API_VERSION}/" "${SRCROOT}/products/libllbuild/include/llbuild/version.h.in" > "${SRCROOT}/products/libllbuild/include/llbuild/version.h"
+
+# For use with SwiftPM, output the conditional compilation flag to stdout
+echo "-Xswiftc -DLLBUILD_C_API_VERSION_${LLBUILD_C_API_VERSION}"
```
ddunbar: Technically, this -D is only needed by downstream clients, right? I think we can dispense with that, since they can use the package manager features to manage which version of llbuild they pick up.

Or do we use this in our own code?

dmbryson: We have the version guard code in our own Swift bindings at the moment.   Arguably those should be version matched to the underlying libllbuild library/code and built/tested by CI to verify that is the case.  We have a little bit of work to make that true, though, so in the meantime this has the benefit of being a safety net that can catch divergences.

PR #257, opened 2018-03-27
Report extended results in commandProcessFinished

dmbryson:
Collect user/system time and max RSS values for subprocesses and report
them via a new CommamdExtendedResults struct in the
commandProcessFinished callback.

rdar://problem/37889042


dmbryson:
@swift-ci please smoke test

dmbryson:
This is just re-landing PR #237 now that downstream projects are ready for it.

PR #259, opened 2018-03-27
Prevent spurious rebuilds with SwiftPM builds of llbuild

jakepetroules:
The generate-version-h.sh script was writing a new version.h file (thus
updating the timestamp, thus triggering a change detection by the build
system) every time it was run, regardless of whether the content had
actually changed. Only update the file if doing so will actually change
the content, in order to make null builds actually work again.


dmbryson:
@swift-ci please smoke test

PR #261, opened 2018-03-28
Add some macros to easily allow users to hook into the tracing system

jakepetroules:
Because these are simple RAII structs intended to be used directly, they
can't be expected to be subclassed, and users would end up having to
redefine their implementation in a modified version of this header.

### DIFF
--- a/include/llbuild/Basic/Tracing.h
+++ b/include/llbuild/Basic/Tracing.h
@@ -15,6 +15,31 @@
 
 #include "llvm/ADT/StringRef.h"
 
+// Allow vendored llbuild installations to provide custom tracing entry points
+#if __has_include("TracingMacros.h")
+#include "TracingMacros.h"
+#endif
+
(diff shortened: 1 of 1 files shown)


include/llbuild/Basic/Tracing.h
```diff
@@ -15,6 +15,30 @@
 
 #include "llvm/ADT/StringRef.h"
 
+#if __has_include("TracingMacros.h")
```
ddunbar: These seems a little convoluted, but I agree it might make it easier to compose custom tracing endpoints.

Can you at least include a comment here about why this is phrased the way this is (basically, so that vendored llbuild installations can provide custom tracing entry points).

jakepetroules: Done!

ddunbar:
@swift-ci please smoke test

PR #268, opened 2018-04-04
Set tab width to 2 for C/C++ files in llbuild's Xcode project

jakepetroules:
llbuild uses LLVM coding conventions, so make it convenient to actually
edit the source code within Xcode. An attempt has been made to set tab
widths "intelligently" based on group contents in order to minimize the
resulting diff.


dmbryson:
@swift-ci please smoke test

PR #271, opened 2018-04-05
[BuildSystem] Support disabling SIGINT on cancellation.

ddunbar:
- This adds support for a new field in the manifest which allows a client to
   disable use of SIGINT when cancelling certain tasks.

 - Some tools which run during a build, like `codesign`, can leave stray
   temporary files which will inhibit correct operation on the next build. Since
   the build system doesn't know about these files, it is useful to try and
   avoid interrupting the tool in the hopes that it will finish before our
   SIGKILL timeout expires and we force the tool to cancel.

 - <rdar://problem/38773503> Support for not killing codesign with SIGINT (llbuild work)


ddunbar:
@swift-ci please smoke test

PR #272, opened 2018-04-05
Rudimentary cycle resolution support

dmbryson:
Although preliminary and disabled by default, if a delegate opts in with the 'shouldResolveCycle' callback, the engine will attempt to resolve build cycles.  This is done in two ways, currently.  First, it will find nodes in the cycle that are pending scan and marked the as 'needs to run'.  Second, it will try to locate an item in the cycle that has a prior value and supply it to the node requesting it in the cycle.


dmbryson:
@swift-ci please smoke test

include/llbuild/Core/BuildEngine.h
```diff
@@ -203,6 +203,16 @@ class BuildEngineDelegate {
...
   virtual Rule lookupRule(const KeyType& key) = 0;
 
+  /// Called when a cycle is detected by the build engine to check if it should
+  /// attempt to resolve the cycle and continue
+  ///
+  /// \param items The ordered list of items comprising the cycle, starting from
+  /// the node which was requested to build and ending with the first node in
+  /// the cycle (i.e., the node participating in the cycle will appear twice).
+  /// \returns True if the engine should attempt to resolve the cycle, false
+  /// otherwise.
```
ddunbar: Probably good to have a little warning here about the consequences of returning true (or, alternately, we need that warning/discussion *somewhere*)

lib/Core/BuildEngine.cpp
```diff
@@ -528,7 +535,10 @@ class BuildEngineImpl {
   void processRuleScanRequest(RuleScanRequest request) {
     auto& ruleInfo = *request.ruleInfo;
 
-    assert(ruleInfo.isScanning());
+    // With forced builds in cycle breaking, we may end up being asked to scan
+    // something that has already been 'scanned'
+    if (!ruleInfo.isScanning())
+      return;
```
ddunbar: Is there a way we could preserve this assert so it checks that either the rule is scanning, or it is being forced?

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
     std::lock_guard<std::mutex> guard1(taskInfosMutex);
     std::lock_guard<std::mutex> guard2(finishedTaskInfosMutex);
 
+    std::vector<Rule*> cycleList = findCycle(buildKey);
+    assert(!cycleList.empty());
+
+    if (delegate.shouldResolveCycle(cycleList)) {
+      TracingInterval i(EngineQueueItemKind::BreakingCycle);
+
+      for (auto ruleIt = cycleList.rbegin(); ruleIt != cycleList.rend(); ruleIt++) {
```
ddunbar: This function might be easier to read if this was factored out into a separate `breakCycle` function + conditional return, which would make the fall through more obvious.

ddunbar: It would be nice to have a comment here (or somewhere) explaining the basic strategy... for example how/why we start with the node we start with.

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+              finishedInputRequests.insert(finishedInputRequests.end(), *it);
+
+              // remove this request from the task info
+              taskInfo->requestedBy.erase(it);
+              return true;
+            }
+          }
+
+        }
+      }
```
ddunbar: Do we have test coverage for both of these ifs failing to match?

The current client API is a little awkward in that the client can't know whether or not to present a diagnostic about a cycle, because it doesn't know if it is going to get broken, or if instead this is going to fall through and report a cycle diagnostic. I think we need to make sure the client can distinguish these situations (and make sure we have test coverage showing the difference)

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+          return true;
+        }
+
+        // Check if this rule has a (potentially) valid previous result and if so
+        // try to provide it to the node requesting it to break the cycle.
+        if (ruleInfo.isInProgressWaiting() && ruleInfo.result.builtAt != 0) {
+          auto* taskInfo = ruleInfo.getPendingTaskInfo();
+
+          // find downstream node in the cycle that wants this input
+          auto& nextRuleInfo = getRuleInfoForKey((*std::next(ruleIt))->key);
```
ddunbar: This needs to handle wraparound, doesn't it?

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+                trace->cycleSupplyPriorValue(&ruleInfo.rule, it->taskInfo->task.get());
+              }
+              it->forcePriorValue = true;
+              finishedInputRequests.insert(finishedInputRequests.end(), *it);
+
+              // remove this request from the task info
+              taskInfo->requestedBy.erase(it);
+              return true;
+            }
+          }
```
ddunbar: It would be a serious error if this loop fell through, right?

I would suggest phrasing this as a search for the matching node (maybe via a helper function), which must succeed, then an assert, then the logic to break that node.

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+        auto& ruleInfo = getRuleInfoForKey((*ruleIt)->key);
+
+        // If this rule is scanning, try to force a rebuild to break the cycle
+        if (ruleInfo.isScanning()) {
+          if (trace) {
+            trace->cycleForceRuleNeedsToRun(&ruleInfo.rule);
+          }
+          for (auto it = ruleInfosToScan.begin(); it != ruleInfosToScan.end(); it++) {
+            if (it->ruleInfo->keyID == ruleInfo.keyID) {
+              ruleInfosToScan.erase(it);
```
ddunbar: This should break, I think (I think it can actually crash if it doesn't)?

Actually, maybe we should probably find the node, then assert we found it, then erase (same comment below)

ddunbar: I think you can also just use `it->ruleInfo == ruleInfo` and avoid a dereference.

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+        // try to provide it to the node requesting it to break the cycle.
+        if (ruleInfo.isInProgressWaiting() && ruleInfo.result.builtAt != 0) {
+          auto* taskInfo = ruleInfo.getPendingTaskInfo();
+
+          // find downstream node in the cycle that wants this input
+          auto& nextRuleInfo = getRuleInfoForKey((*std::next(ruleIt))->key);
+
+          // find the corresponding request on this task
+          for (auto it = taskInfo->requestedBy.begin(); it != taskInfo->requestedBy.end(); it++) {
+            if (it->taskInfo->forRuleInfo->keyID == nextRuleInfo.keyID) {
```
ddunbar: I think you can use `it->taskInfo->forRuleInfo == nextRuleInfo` and avoid a dereference.

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+            if (it->ruleInfo->keyID == ruleInfo.keyID) {
+              ruleInfosToScan.erase(it);
+            }
+          }
+          finishScanRequest(ruleInfo, RuleInfo::StateKind::NeedsToRun);
+          return true;
+        }
+
+        // Check if this rule has a (potentially) valid previous result and if so
+        // try to provide it to the node requesting it to break the cycle.
```
ddunbar: I'd probably put a comment here explaining why we don't complete this node completely.

lib/Core/BuildEngine.cpp
```diff
@@ -931,31 +942,97 @@ class BuildEngineImpl {
...
+
+          // find the corresponding request on this task
+          for (auto it = taskInfo->requestedBy.begin(); it != taskInfo->requestedBy.end(); it++) {
+            if (it->taskInfo->forRuleInfo->keyID == nextRuleInfo.keyID) {
+              // supply the prior value to the node
+              if (trace) {
+                trace->cycleSupplyPriorValue(&ruleInfo.rule, it->taskInfo->task.get());
+              }
+              it->forcePriorValue = true;
+              finishedInputRequests.insert(finishedInputRequests.end(), *it);
```
ddunbar: I think it is possible that this change exposes the following very subtle race:
1. I *think* it is true that we don't protect accesses to `ruleInfo->result`, because no task should ever try and read them before they have been written.
2. However, with this change, the processing of the finished input request will in fact try to read the result.
3. If the task meanwhile gets spun up and finishes, it could be writing that.

I think we might end up needing a separate queue to hold the "forced input requests" (since either copying the result or adding a lock to the hot path feels annoying for something which will rarely trigger).

Can you confirm my reasoning here? I may have missed something.

```

File: lib/Core/BuildEngine.cpp (hunk 2 above)
Comment (ddunbar) [reply lib/Core/BuildEngine.cpp]: I'm not sure I understand when this happens, can you explain?

Comment (dmbryson) [reply lib/Core/BuildEngine.cpp]: I believe the pathway that leads here is that a node gets placed on the deferredScanRequests list. When we force a node to build it has explicitly been moved past the scanning phase.  However, the node that deferred the scan may not be in the cycle we are currently evaluating.

File: lib/Core/BuildEngine.cpp (hunk 3 above)
Comment (dmbryson) [reply lib/Core/BuildEngine.cpp]: Thought had cross my mind.  Sounds good.

File: lib/Core/BuildEngine.cpp (hunk 10 above)
Comment (dmbryson) [reply lib/Core/BuildEngine.cpp]: Indeed, this thought crossed my mind.  I convinced myself it was okay, but perhaps being explicit about it would be safer, or at least a comment explaining the logic. Basically, if we have reached this point, the engine has stalled all other work.  We are explicitly unblocking only a single item with this prior value at this point, thus it should be the first task executed in the main task loop above.

**Comment by dmbryson** [comment]
@swift-ci please smoke test

**Comment by dmbryson** [comment]
@swift-ci please smoke test

**Comment by dmbryson** [comment]
This should be good to go for another review.

File: lib/Core/BuildEngine.cpp [hunk 11]
```diff
@@ -1075,10 +1112,104 @@ class BuildEngineImpl {
...
+        auto* taskInfo = ruleInfo.getPendingTaskInfo();
+
+        // find downstream node in the cycle that wants this input
+        auto& nextRuleInfo = getRuleInfoForKey((*std::next(ruleIt))->key);
+
+        // find the corresponding request on this task
+        auto it = findTaskInputRequestForRule(taskInfo->requestedBy, &nextRuleInfo);
+        if (it == taskInfo->requestedBy.end()) {
+          // this rule has not generated an input request yet, so we cannot
+          // provide a value to it
```
ddunbar: When does this happen?

lib/Core/BuildEngine.cpp
```diff
@@ -1124,6 +1255,13 @@ class BuildEngineImpl {
       ruleInfo->setCancelled();
     }
 
+    for (auto& it: ruleInfos) {
+      // Cancel outstanding activity on rules
+      if (it.second.isScanning()) {
+        it.second.setCancelled();
+      }
```
ddunbar: Can you explain why this is needed now? Was it not needed before, or it was an oversight?

ddunbar: Also, this is really expensive because of the full scan... isn't there a cheaper way to simply iterate over the rules which are being scanned (`ruleInfosToScan`)?

ddunbar:
Have you tried running this test suite and checking our code coverage for this change? It is subtle enough that I think we should try and have code coverage for all of the cycle handling, if we don't yet. (doesn't necessarily need to come in this particular PR)

lib/Core/BuildEngine.cpp
dmbryson: This situation occurs right now in the unit test the second time through when we are trying to break the cycle, after we have forced the rule to run in the previous attempt.  It was implicitly handled in this manner in my previous code organization, but is more explicit this way.  That said, I will re-read the code path to confirm my understanding of the sequence of events.

lib/Core/BuildEngine.cpp
dmbryson: I believe it was an oversight.  It was exposed when I added a third pass to the cycle detection unit test, even before actually breaking the cycle.  i.e. trying to perform a build after cancellation caused asserts to fire.

lib/Core/BuildEngine.cpp
dmbryson: Ah yes, I see how this happens.  If the node that will ultimately want this input (nextRuleInfo) has not been scanned yet, it won't have an input request.  How the breaking is working right now, we force the node to build, then we supply it the prior value.   There is likely a good bit of room for improvement here.

dmbryson:
@swift-ci please smoke test

PR #275, opened 2018-04-09
Fix static analyzer warnings about zero-allocations

dmbryson:
Although doing this is technically legal, as the C++ Standard states: "The effect of dereferencing a pointer returned as a request for zero size is undefined."

So it's not very useful. And the static analyzer doesn't like it anyways. So always allocate at least one byte and null-terminate the entire thing. I specifically didn't just set p/contents to nullptr if size == 0 because that would simply introduce a different analyzer warning.


dmbryson:
@swift-ci please smoke test

PR #279, opened 2018-04-16
Fix findCycle when called following a prior forced build action

dmbryson:


dmbryson:
@swift-ci please smoke test

ddunbar:
Can we get a test case for this?

dmbryson:
Yeah, I'll look into it.

PR #281, opened 2018-04-18
Allow tracing points to be conditionally enabled and disabled

dmbryson:
- Exposes tracing point toggle via the C API


dmbryson:
@swift-ci please smoke test

PR #282, opened 2018-04-20
Refactor Engine/DB Key/KeyID handling

dmbryson:
- Separate database key IDs from engine keys to reduce database queries
while holding the engine key table lock
- Add support for directly accessing the byte array of BinaryEncoding

### DIFF
--- a/include/llbuild/Core/BuildDB.h
+++ b/include/llbuild/Core/BuildDB.h
@@ -28,22 +28,34 @@ namespace core {
 struct Result;
 class Rule;
 
-class BuildDB {
+/// Delegate interface for use with the build database
+class BuildDBDelegate {
 public:
-  virtual ~BuildDB();
+  virtual ~BuildDBDelegate();
 
   /// Get a unique ID for the given key.
   ///
   /// This method is thread safe.
   ///
   /// \param key [out] The key whose unique ID is being returned.
-  /// \param error_out [out] Error string if return value is 0.
-  virtual KeyID getKeyID(const KeyType& key, std::string *error_out) = 0;
+  virtual KeyID getKeyID(const KeyType& key) = 0;
 
   /// Get the key corresponding to a key ID.
   ///
   /// This method is thread safe, and cannot fail.
   virtual KeyType getKeyForID(KeyID key) = 0;
+};
+
+
+class BuildDB {
+public:
+  virtual ~BuildDB();
+
+  /// Set the delegate for key/id mapping
+  ///
+  /// \param delegate The delegate pointer (does not take ownership)
+  virtual void attachDelegate(BuildDBDelegate* delegate) = 0;
+
   
   /// Get the current build iteration.
   ///
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -50,7 +50,7 @@ bool BuildEngineDelegate::shouldResolveCycle(const std::vector<Rule*>& items,
 
 namespace {
 
-class BuildEngineImpl {
+class BuildEngineImpl : public BuildDBDelegate {
   struct RuleInfo;
   struct TaskInfo;
 
@@ -322,6 +322,8 @@ class BuildEngineImpl {
   /// This variable is used to signal when additional work is added to the
   /// FinishedTaskInfos queue, which the engine may need to wait on.
   std::condition_variable finishedTaskInfosCondition;
+
+
 
 private:
   /// @name RuleScanRecord Allocation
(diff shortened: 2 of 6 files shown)


dmbryson:
In the synthetic llvm ninja build benchmark this is not a clear win.  Indeed, it seems to make the test more variable, rather than always good or always bad.

llbuild.xcodeproj/xcshareddata/xcbaselines/E1C404AB1A0308F3003392BA.xcbaseline/B0B8BCF6-6DB5-43A2-9D39-B420784A2097.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dmbryson: This was not supposed to be added.

include/llbuild/Core/BuildDB.h
```diff
@@ -28,22 +28,34 @@ namespace core {
...
+  virtual ~BuildDBDelegate();
 
   /// Get a unique ID for the given key.
   ///
   /// This method is thread safe.
   ///
   /// \param key [out] The key whose unique ID is being returned.
-  /// \param error_out [out] Error string if return value is 0.
-  virtual KeyID getKeyID(const KeyType& key, std::string *error_out) = 0;
+  virtual KeyID getKeyID(const KeyType& key) = 0;
```
ddunbar: I don't think this method is actually used on the delegate.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -600,6 +571,70 @@ class SQLiteBuildDB : public BuildDB {
   }
 
   /// @}
+
+  
+private:
+  /// Inserts a key if not present and always returns keyID
+  /// Sometimes key will be inserted for a lookup operation
+  /// but that is okay because it'll be added at somepoint anyway
+  KeyID getKeyID(const KeyType& key, std::string *error_out) {
```
ddunbar: We should probably introduce a new private `DBKeyID` type to prevent this accidentally being mixed

dmbryson: 👍 Indeed, during development of this I had a bug where that was the case.

include/llbuild/Core/BuildDB.h
dmbryson: lookupRuleResult uses it

ddunbar: That doesn't need to be exposed via the delegate though, that is internal to the DB?

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

include/llbuild/Core/BuildDB.h
dmbryson: It does, in that it is part of the bidirectional mapping.  Since a rule result object has a vector of engine KeyIDs, we need to ask the delegate what the engine key ID is for each key after reading it out of the database.

ddunbar: Oh, right, I'm sorry... of course.

More ☕️ needed

dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
```diff
@@ -1343,15 +1331,10 @@ class BuildEngineImpl {
...
-    if (db) {
-      key = db->getKeyForID(keyID);
-    } else {
-      // Note that we don't need to lock `keyTable` here because the key entries
-      // themselves don't change once created.
-      key = llvm::StringMapEntry<KeyID>::GetStringMapEntryFromKeyData(
+    // Note that we don't need to lock `keyTable` here because the key entries
+    // themselves don't change once created.
+    KeyType key = llvm::StringMapEntry<KeyID>::GetStringMapEntryFromKeyData(
           (const char*)(uintptr_t)keyID).getKey();
```
ddunbar: Should we just change this to call `getKeyForID`?

lib/Core/BuildEngine.cpp
```diff
@@ -1292,37 +1294,23 @@ class BuildEngineImpl {
...
-    // Otherwise, use our builtin key table.
-    //
     // The RHS of the mapping is actually ignored, we use the StringMap's ptr
     // identity because it allows us to efficiently map back to the key string
     // in `getRuleInfoForKey`.
     auto it = keyTable.insert(std::make_pair(key, 0)).first;
     return (KeyID)(uintptr_t)it->getKey().data();
   }
 
+  virtual KeyType getKeyForID(KeyID key) override {
```
ddunbar: Should we make this return a const reference instead? That should be a little more efficient, since we are probably copying these keys a bunch right now.

ddunbar: Actually, ignore that suggestion... that doesn't work. We would need to return a `StringRef` and that breaks the abstraction of "KeyType".

It's probably worth us having a follow on bug to improve this at some later date though.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -41,11 +41,18 @@ class SQLiteBuildDB : public BuildDB {
   /// * 4: Pre-history
   static const int currentSchemaVersion = 7;
 
+  /// Internal key ID type used for linking rule dependencies. Related to, but
+  /// distinct from the build engine's KeyID type.
+  typedef uint64_t DBKeyID;
```
ddunbar: This isn't actually going to give us any type safety, is it (because of C++)? I think for us to get type safety we would have to make a wrapper type.

dmbryson: True enough.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -600,6 +577,70 @@ class SQLiteBuildDB : public BuildDB {
   }
 
   /// @}
+
+  
+private:
+  /// Inserts a key if not present and always returns keyID
+  /// Sometimes key will be inserted for a lookup operation
+  /// but that is okay because it'll be added at somepoint anyway
```
ddunbar: This second part of the comment isn't true anymore, I guess?

dmbryson: Well, it is still technically true for this method. We no longer do it for 'lookupRuleResult', but it could conceivably happen for discovered dependencies.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

lib/Core/SQLiteBuildDB.cpp
ddunbar: 👍

PR #283, opened 2018-04-23
[BuildSystem] Set QOS on launched processes.

ddunbar:
- On Apple platforms, we want to default everything to running in the "utility"
   QOS class.

 - <rdar://problem/24986348> [llbuild] Set QOS on spawned tasks


ddunbar:
@swift-ci please smoke test

PR #284, opened 2018-04-24
Install llbuild binary

saschpe:
Otherwise needs to be copied manually into the install-dir of a
successful CMake build.

### DIFF
--- a/products/llbuild/CMakeLists.txt
+++ b/products/llbuild/CMakeLists.txt
@@ -13,3 +13,7 @@ target_link_libraries(llbuild
 if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
   target_link_libraries(llbuild curses)
 endif()
+
+install(TARGETS llbuild
+        COMPONENT llbuild
+        DESTINATION bin)


dmbryson:
@swift-ci please smoke test

ddunbar:
@saschpe why do you want to install `llbuild`?

The goal of `llbuild` is really to provide libraries. The command line tool is more oriented currently towards being a testing tool for `llbuild` internally, and not something we necessary want people to depend on.

dmbryson:
Ah, true.  Seemed innocuous to install the binary, but does 'break the 4th wall' I suppose.

saschpe:
So I started an attempt to package Swift and dependencies for Fedora. llbuild is part of that. I might be understanding the purpose of the binary but I wanted to put it into a llbuild-devel package. maybe llbuild-tools is more appropriate though. This is all in-flux at the moment as it's quite a challenge to bisect the Swift build-process into individual / independent pieces that can be packaged properly.

ddunbar:
@saschpe ah, ok. You shouldn't need to install `llbuild` to package Swift -- the only thing that is needed is `swift-build-tool` which is used by the Swift package manager. I'd recommend diving into the Swift `build-script`, which is what produces the toolchains on https://swift.org; that is the canonical definition of what is needed for a Swift release.

Given that context, I'm going to close this out because I'd prefer not start installing this unless it meets a concrete need.

ddunbar:
Here is a link to the build script: https://github.com/apple/swift/blob/master/utils/build-script

PR #285, opened 2018-04-24
Adhere to CMake LIB_SUFFIX if set

saschpe:
CMake cache variables need to be changed manually in CMakeCache.txt,
which is annoying at best. Instead, default to LIB_SUFFIX, which is
commonly used in many other projects and in Linux packaging macros.


dmbryson:
@swift-ci please smoke test

dmbryson:
I do wonder if we should be using LIB_SUFFIX in general, rather than replicating it with LLBUILD_LIBDIR_SUFFIX, but this change seems reasonable to me as it is.

PR #286, opened 2018-04-24
Add a cache for DB <-> Engine key ID mapping in the db layer

dmbryson:
- This yields a 75% increase in performance for testPseudoLLVMNullBuild

rdar://problem/39697555

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -16,6 +16,7 @@
 #include "llbuild/Basic/PlatformUtility.h"
 #include "llbuild/Core/BuildEngine.h"
 
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/raw_ostream.h"
 
@@ -410,13 +411,9 @@ public:
       DBKeyID dbKeyID;
       decoder.read(dbKeyID.value);
 
-      // Map the database key ID into an engine key ID
-      //
-      // FIXME: This is roundtripping through the key table for every ID with
-      // no caching, thus likely fairly expensive. Should consider a db layer
-      // cache of key/id mappings.
-      auto key = getKeyForID(dbKeyID);
-      result_out->dependencies[i] = delegate->getKeyID(key);
+      // Map the database key ID into an engine key ID (note that we already
+      // hold the dbMutex at this point as required by getKeyIDforID())
+      result_out->dependencies[i] = getKeyIDForID(dbKeyID);
     }
 
     return true;
@@ -585,6 +582,7 @@ public:
 
   
 private:
+
   /// Inserts a key if not present and always returns keyID
   /// Sometimes key will be inserted for a lookup operation
   /// but that is okay because it'll be added at somepoint anyway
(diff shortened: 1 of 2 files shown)


dmbryson:
@swift-ci please smoke test

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -23,6 +23,7 @@
 #include <cerrno>
 #include <cstring>
 #include <mutex>
+#include <unordered_map>
```
ddunbar: Can we use `llvm:DenseMap` rather than an `std` data structure here? The LLVM ones are generally faster.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -626,25 +623,42 @@ class SQLiteBuildDB : public BuildDB {
...
+  std::unordered_map<uint64_t, KeyID> engineKeyIDs;
 
-    // Seach for the key.
-    result = sqlite3_reset(findKeyNameForKeyIDStmt);
-    assert(result == SQLITE_OK);
-    result = sqlite3_clear_bindings(findKeyNameForKeyIDStmt);
-    assert(result == SQLITE_OK);
-    result = sqlite3_bind_int64(findKeyNameForKeyIDStmt, /*index=*/1, keyID.value);
-    assert(result == SQLITE_OK);
+  /// Maps a DBKeyID into an engine KeyID
```
ddunbar: We should probably have a comment here about what expectations are around holding a lock around this call? I think we are effectively using `dbMutex` for this.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -626,25 +623,42 @@ class SQLiteBuildDB : public BuildDB {
...
+  KeyID getKeyIDForID(DBKeyID keyID) {
+    // Search local db <-> engine mapping cache
+    auto it = engineKeyIDs.find(keyID.value);
+    if (it == engineKeyIDs.end()) {
```
ddunbar: I would invert the sense of this, so that the early exit is the fast return and most of the code can be de-dented.

lib/Core/SQLiteBuildDB.cpp
ddunbar: Also, should we document somewhere clearly that the engine <-> db interface must keep KeyIDs stable across multiple builds? Otherwise we would need to reset that map on every build.

ddunbar:
FYI, it looks like with this PR, and your previous one, a command line null build of "FileCheck" is now faster than before! Nice work! (The full build is insanely faster, I'm too embarrassed to even show the number)

BEFORE:
```
$ runN 10 llbo ninja build FileCheck
name	  avg  	  min  	  med  	  max  	  SD   	  SD% 	 total 
user	 0.3770	 0.3643	 0.3735	 0.4008	 0.0111	   2.95	 3.7699
 sys	 0.3078	 0.2899	 0.3034	 0.3247	 0.0122	   3.96	 3.0776
wall	 0.6980	 0.6647	 0.6879	 0.7395	 0.0225	   3.22	 6.9796
```

AFTER:
```
ddunbar@ozzy-5:llvm.build$ runN 10 llbo ninja build FileCheck
name	  avg  	  min  	  med  	  max  	  SD   	  SD% 	 total 
user	 0.2837	 0.2752	 0.2849	 0.2923	 0.0053	   1.86	 2.8375
 sys	 0.1413	 0.1429	 0.1440	 0.1481	 0.0060	   4.22	 1.4134
wall	 0.4370	 0.4284	 0.4410	 0.4538	 0.0103	   2.36	 4.3701
```

**Actually**, on this particular test case it looks like we are now faster than `ninja` itself (which historically we lagged by a bit):

Ninja
```
$ runN 10 $(xcrun --find ninja) FileCheck
name	  avg  	  min  	  med  	  max  	  SD   	  SD% 	 total 
user	 0.4394	 0.4087	 0.4358	 0.4827	 0.0223	   5.07	 4.3938
 sys	 0.0556	 0.0520	 0.0537	 0.0542	 0.0038	   6.89	 0.5557
wall	 0.4988	 0.4628	 0.4934	 0.5430	 0.0242	   4.85	 4.9885
```

🎉

PR #287, opened 2018-04-25
Add fast path lookupRuleResult query

dmbryson:
- Cache engine KeyID to DBKeyID mapping
- When ID mapping is known, skip table join in lookupRuleResult
- Cache ID mapping when reading out previously unknown rules

Yields a ~17% improvement in the testPseudoLLVMNullBuild test


ddunbar:
@swift-ci please smoke test

PR #288, opened 2018-04-25
Refactor setRuleResult to use cache and update with single query

dmbryson:
This does not seem to make much of a performance improvement for the perf tests (~1-2% reduction), but may help more real world situations.  Maximizes our chances of caching IDs and simplifies the query code for storing results.

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -32,25 +32,38 @@ using namespace llbuild::core;
 
 // SQLite BuildDB Implementation
 
+/// Internal key ID type used for linking rule dependencies. Related to, but
+/// distinct from the build engine's KeyID type.
+struct DBKeyID {
+  uint64_t value = 0;
+
+  DBKeyID() { ; }
+  DBKeyID(const DBKeyID&) = default;
+  explicit DBKeyID(uint64_t value) : value(value) { ; }
+};
+
+// Provide DenseMapInfo for DBKeyID.
+template<> struct ::llvm::DenseMapInfo<DBKeyID> {
+  static inline DBKeyID getEmptyKey() { return DBKeyID(~0ULL); }
+  static inline DBKeyID getTombstoneKey() { return DBKeyID(~0ULL - 1ULL); }
+  static unsigned getHashValue(const DBKeyID& Val) {
+    return (unsigned)(Val.value * 37ULL);
+  }
+  static bool isEqual(const DBKeyID& LHS, const DBKeyID& RHS) {
+    return LHS.value == RHS.value;
+  }
+};
+
 namespace {
 
 class SQLiteBuildDB : public BuildDB {
   /// Version History:
+  /// * 8: Remove ID from rule results
   /// * 7: De-normalized the rule result dependencies.
   /// * 6: Added `ordinal` field for dependencies.
   /// * 5: Switched to using `WITHOUT ROWID` for dependencies.
   /// * 4: Pre-history
-  static const int currentSchemaVersion = 7;
-
-  /// Internal key ID type used for linking rule dependencies. Related to, but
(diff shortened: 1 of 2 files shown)


dmbryson:
@swift-ci please smoke test

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -192,8 +193,7 @@ class SQLiteBuildDB : public BuildDB {
       if (result == SQLITE_OK) {
         result = sqlite3_exec(
           db, ("CREATE TABLE rule_results ("
-               "id INTEGER PRIMARY KEY, "
-               "key_id INTEGER, "
+               "key_id INTEGER PRIMARY KEY, "
```
ddunbar: Heh, of course. Wonder why I didn't do this in the first place! 🤔

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -435,24 +428,23 @@ class SQLiteBuildDB : public BuildDB {
       }
 
       // Otherwise, read the result contents from the row.
-      assert(sqlite3_column_count(findRuleResultStmt) == 6);
+      assert(sqlite3_column_count(findRuleResultStmt) == 5);
       ruleID = sqlite3_column_int64(findRuleResultStmt, 0);
```
ddunbar: Can we call this `dbKeyID`? I think it would be a little more readable below, when we put it into the mapping table.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -640,14 +581,27 @@ class SQLiteBuildDB : public BuildDB {
 
   
 private:
+  /// Local cache of database DBKeyID (values) to engine KeyIDs
+  llvm::DenseMap<uint64_t, KeyID> engineKeyIDs;
```
ddunbar: FWIW, DenseMap should offer some kind of traits you can implement to allow this to be a DBKeyID -> KeyID map.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -640,14 +581,27 @@ class SQLiteBuildDB : public BuildDB {
...
+
+  /// Local cache of database engine KeyIDs to DBKeyIDs
+  llvm::DenseMap<KeyID, DBKeyID> dbKeyIDs;
+
+  /// Lookup or create a DBKeyID for a given engine KeyID
+  ///
+  /// This method is not thread-safe. The caller must protect access via the
+  /// dbMutex.
+  DBKeyID getKeyID(KeyID keyID, std::string *error_out) {
```
ddunbar: I think this function would be a little more readable if you split out a function that actually did the database lookup or insert operation, and the helper that then wrote the result into the caches, if we didn't have a cache entry.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -477,19 +469,10 @@ class SQLiteBuildDB : public BuildDB {
...
   }
 
-  static constexpr const char *findIDForKeyInRuleResultsStmtSQL =
-    "SELECT id FROM rule_results "
-    "WHERE key_id == ? LIMIT 1;";
-  sqlite3_stmt* findIDForKeyInRuleResultsStmt = nullptr;
-
   static constexpr const char *insertIntoRuleResultsStmtSQL =
-    "INSERT INTO rule_results VALUES (NULL, ?, ?, ?, ?, ?);";
+    "INSERT OR REPLACE INTO rule_results VALUES (?, ?, ?, ?, ?);";
```
ddunbar: Nice! I'm surprised we don't see a win on this PR, but maybe our unit tests don't properly exercise these scenarios.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #295, opened 2018-04-29
[BuildSystem] Disable defaulting to UTILITY QOS for tasks.

ddunbar:
- This turns out to have a pretty substantial impact on some build scenarios,
   so I am going to reland this but exposed as an API so that clients can try to
   choose more intelligently what class is appropriate.

 - <rdar://problem/39825229> llbuild should use default QOS class for subprocess, by default


ddunbar:
@swift-ci please smoke test

PR #300, opened 2018-05-01
Xcode: add a couple inputs/outputs whose absence was preventing null builds

jakepetroules:
…builds


dmbryson:
@swift-ci please smoke test

PR #301, opened 2018-05-02
<rdar://35326105> Rethink the args-for-signature manifest property

jakepetroules:
Instead of adding a domain-specific "args-for-signature" property, we
instead now have "additional-signature-data" which allows to add
arbitrary string data that will contribute to the task signature, and
ignore-args-in-signature, that will allow to remove existing args data
from the task signature. This is a more flexible design and also allows
us to ADD arbitrary data to the signature as well.

### DIFF
--- a/docs/buildsystem.rst
+++ b/docs/buildsystem.rst
@@ -499,18 +499,11 @@ attributes on commands, and not at the tool level.
      - A string or string list indicating the command line to be executed. If a
        single string is provided, it will be executed using ``/bin/sh -c``.
 
-   * - args-for-signature
-     - A string or string list indicating the canonical form of the command line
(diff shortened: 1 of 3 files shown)


ddunbar:
I thought we had decided to go with an even more generic `signature` property, which will completely define the signature for the task (and the task must include the other bits).

I think I prefer that approach, since it results in an even more compact manifest, and is more extensible... even if the clients have to take on some of the signature computation (but that is something I expected they would have to do anyway in the long term)

jakepetroules:
@ddunbar Right, uploaded the updated version.

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #302, opened 2018-05-03
Add support for configuring the build system scheduler

dmbryson:
- Add methods for changing the scheduler algorithm. Supports:
-- commandNamePriority: priority queue ordered by command name [default]
-- fifo: first in/first out queue - previous behavior
- Add methods for setting the lane width of the scheduler
- Add command line option for llbuild debug tool for selecting the scheduler

### DIFF
--- a/include/llbuild/BuildSystem/BuildSystem.h
+++ b/include/llbuild/BuildSystem/BuildSystem.h
@@ -295,6 +295,22 @@ enum class QualityOfService {
 QualityOfService getDefaultQualityOfService();
 void setDefaultQualityOfService(QualityOfService level);
 
+// MARK: Execution Queue Scheduler Control
+
+enum class SchedulerAlgorithm {
+  /// Command name priority queue based scheduling [default]
+  commandNamePriority = 0,
+
+  /// First in, first out
+  fifo = 1
+};
+
+SchedulerAlgorithm getSchedulerAlgorithm();
+void setSchedulerAlgorithm(SchedulerAlgorithm algorithm);
+
+uint32_t getSchedulerLaneWidth();
+void setSchedulerLaneWidth(uint32_t width);
+
 }
 }
 
--- a/lib/BuildSystem/LaneBasedExecutionQueue.cpp
+++ b/lib/BuildSystem/LaneBasedExecutionQueue.cpp
@@ -56,14 +56,6 @@
 using namespace llbuild;
 using namespace llbuild::buildsystem;
 
-// Enable the priority based scheduler by default, based on measured performance
-// improvements for large projects.
-//
-// FIXME: We should turn this into an option that can be selected at
-// construction time so that clients may select the algorithm at run time via
-// the API.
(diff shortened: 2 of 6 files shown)


dmbryson:
@swift-ci please smoke test

include/llbuild/BuildSystem/BuildSystem.h
```diff
@@ -295,6 +295,22 @@ enum class QualityOfService {
...
 
+// MARK: Execution Queue Scheduler Control
+
+enum class SchedulerAlgorithm {
+  /// Command name priority queue based scheduling [default]
+  commandNamePriority = 0,
+
+  /// First in, first out
+  fifo = 1
+};
```
ddunbar: Unlike Swift, we have been using initial-uppercase for `enum class` constants... 🤷‍♂️

products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -811,7 +811,7 @@ char* llb_buildsystem_command_get_verbose_description(
   return strdup(result.c_str());
 }
 
-llb_quality_of_service_t llb_get_quality_of_servicellb_enable_tracing() {
+llb_quality_of_service_t llb_get_quality_of_service() {
```
ddunbar: Thanks! 😊

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -607,3 +662,18 @@ void llbuild::buildsystem::setDefaultQualityOfService(QualityOfService level) {
...
+void llbuild::buildsystem::setSchedulerAlgorithm(SchedulerAlgorithm algorithm) {
+  defaultAlgorithm = algorithm;
+}
+
+uint32_t llbuild::buildsystem::getSchedulerLaneWidth() {
+  return defaultLaneWidth;
+}
+
+void llbuild::buildsystem::setSchedulerLaneWidth(uint32_t width) {
+  defaultLaneWidth = width;
```
ddunbar: I wonder if we should impose a limit here? Unlike systems like say Ninja which uses a select-based mechanism, our lane based approach isn't very compatible with setting a super high lane count... OTOH, its hard to pick a sane limit, so maybe we should just say caveat emptor.

include/llbuild/BuildSystem/BuildSystem.h
dmbryson: Ah, I matched the style of QualityOfService right above.  😉

Probably worth cleaning up at some point, but I'll save it for later <rdar://problem/39958068>

lib/BuildSystem/LaneBasedExecutionQueue.cpp
dmbryson: Caveat emptor it is

include/llbuild/BuildSystem/BuildSystem.h
ddunbar: Well played

PR #304, opened 2018-05-10
Added process identifier to command_extended_result

BenchR267:
The command extended result struct contains information about the spawned subprocesses. It should also contain the process identifier of the spawned child process.
I chose to take int32 as the type and use -1 for the case that the process creation has failed.

rdar://39798231

### DIFF
--- a/include/llbuild/Basic/CrossPlatformCompatibility.h
+++ b/include/llbuild/Basic/CrossPlatformCompatibility.h
@@ -1,0 +1,33 @@
+//===- PlatformUtility.h ----------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+//
+// This file defines cross platform definitions.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CrossPlatformCompatibility_h
+#define CrossPlatformCompatibility_h
+
+#ifdef _MSC_VER
+typedef int llbuild_pid_t;
+#else
+
+#if defined(__linux__) || defined(__GNU__)
+#include <termios.h>
+#else
+#include <sys/types.h>
+#endif
+
+typedef pid_t llbuild_pid_t;
+#endif
+
+#endif /* CrossPlatformCompatibility_h */
--- a/include/llbuild/BuildSystem/CommandResult.h
+++ b/include/llbuild/BuildSystem/CommandResult.h
@@ -13,6 +13,7 @@
 #ifndef LLBUILD_BUILDSYSTEM_COMMAND_RESULT_H
 #define LLBUILD_BUILDSYSTEM_COMMAND_RESULT_H
 
+#include "llbuild/Basic/CrossPlatformCompatibility.h"
 #include <inttypes.h>
 
 namespace llbuild {
@@ -31,23 +32,23 @@ struct CommandExtendedResult {
   CommandResult result; /// The final status of the command
   int exitStatus;       /// The exit code
 
-  uint64_t utime;       /// User time (in us)
-  uint64_t stime;       /// Sys time (in us)
-  uint64_t maxrss;      /// Max RSS (in bytes)
+  uint64_t utime;     /// User time (in us)
+  uint64_t stime;     /// Sys time (in us)
+  uint64_t maxrss;    /// Max RSS (in bytes)
+  llbuild_pid_t pid;  /// Process identifier (can be -1 for failure reasons)
 
-
-  CommandExtendedResult(CommandResult result, int exitStatus, uint64_t utime = 0,
+  CommandExtendedResult(CommandResult result, int exitStatus, llbuild_pid_t pid, uint64_t utime = 0,
                 uint64_t stime = 0, uint64_t maxrss = 0)
-    : result(result), exitStatus(exitStatus)
-    , utime(utime), stime(stime), maxrss(maxrss)
+    : result(result), exitStatus(exitStatus), utime(utime)
+    , stime(stime), maxrss(maxrss), pid(pid)
   {}
 
   static CommandExtendedResult makeFailed(int exitStatus = -1) {
-    return CommandExtendedResult(CommandResult::Failed, exitStatus);
+    return CommandExtendedResult(CommandResult::Failed, exitStatus, -1);
   }
 
   static CommandExtendedResult makeCancelled(int exitStatus = -1) {
-    return CommandExtendedResult(CommandResult::Cancelled, exitStatus);
+    return CommandExtendedResult(CommandResult::Cancelled, exitStatus, -1);
   }
 
 };
--- a/products/libllbuild/include/llbuild/buildsystem.h
+++ b/products/libllbuild/include/llbuild/buildsystem.h
@@ -24,6 +24,19 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#ifdef _MSC_VER
+typedef int llbuild_pid_t;
+#else
+
+#if defined(__linux__) || defined(__GNU__)
+#include <termios.h>
+#else
+#include <sys/types.h>
+#endif
+
+typedef pid_t llbuild_pid_t;
+#endif
+
 /// @name File System Behaviors
 /// @{
 
@@ -102,6 +115,7 @@ typedef struct llb_buildsystem_command_extended_result_t_ {
   uint64_t utime;                           /// User time (in us)
   uint64_t stime;                           /// Sys time (in us)
   uint64_t maxrss;                          /// Max RSS (in bytes)
+  llbuild_pid_t pid;                        /// The process identifier (-1 if process creation failed)
 } llb_buildsystem_command_extended_result_t;
 
 /// Status change event kinds.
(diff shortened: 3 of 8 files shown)


BenchR267:
@swift-ci please smoke test

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -251,11 +251,13 @@ public struct CommandMetrics {
 public struct CommandExtendedResult {
     public let result: CommandResult    /// The result of a command execution
     public let exitStatus: Int32        /// The exit code
+    public let pid: Int32               /// The process identifier (-1 if failed)
```
dmbryson: Given that we are bridging to swift here, I'd say make this an optional that is nil when the process ID doesn't exist.  Also, though not 100% certain ("I found it on the internet"), I think we can also use pid_t here in swift.

include/llbuild/BuildSystem/CommandResult.h
```diff
@@ -34,20 +34,21 @@ struct CommandExtendedResult {
...
+  CommandExtendedResult(CommandResult result, int exitStatus, int32_t pid, uint64_t utime = 0,
                 uint64_t stime = 0, uint64_t maxrss = 0)
-    : result(result), exitStatus(exitStatus)
+    : result(result), exitStatus(exitStatus), pid(pid)
     , utime(utime), stime(stime), maxrss(maxrss)
   {}
 
-  static CommandExtendedResult makeFailed(int exitStatus = -1) {
-    return CommandExtendedResult(CommandResult::Failed, exitStatus);
+  static CommandExtendedResult makeFailed(int32_t pid, int exitStatus = -1) {
```
dmbryson: Might read a bit cleaner if we put the value for 'no pid' here as the default value, that way callers don't need to be aware of what that value needs to be.

products/libllbuild/include/llbuild/buildsystem.h
```diff
@@ -99,6 +99,7 @@ typedef enum {
 typedef struct llb_buildsystem_command_extended_result_t_ {
   llb_buildsystem_command_result_t result;  /// Result of command execution
   int exit_status;                          /// The exit code
+  int32_t pid;                              /// The process identifier (-1 if process creation failed)
```
dmbryson: Although it shouldn't be a problem, for safety I'd prefer to have this field added to the bottom of the struct after maxrss.

include/llbuild/BuildSystem/CommandResult.h
```diff
@@ -34,20 +34,21 @@ struct CommandExtendedResult {
   uint64_t utime;       /// User time (in us)
   uint64_t stime;       /// Sys time (in us)
   uint64_t maxrss;      /// Max RSS (in bytes)
+  int32_t pid;          /// Process identifier (can be -1 for failure reasons)
```
dmbryson: For portability, internally here, we should store this as pid_t.

products/llbuildSwift/BuildSystemBindings.swift
BenchR267: pid_t is an typealias for Int32, but you're right - it's better to be explicit here! :)

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

include/llbuild/BuildSystem/CommandResult.h
```diff
@@ -34,20 +35,20 @@ struct CommandExtendedResult {
   uint64_t utime;       /// User time (in us)
   uint64_t stime;       /// Sys time (in us)
   uint64_t maxrss;      /// Max RSS (in bytes)
+  pid_t pid;            /// Process identifier (can be -1 for failure reasons)
 
-
-  CommandExtendedResult(CommandResult result, int exitStatus, uint64_t utime = 0,
+  CommandExtendedResult(CommandResult result, int exitStatus, int32_t pid, uint64_t utime = 0,
```
dmbryson: s/int32_t/pid_t/

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -264,9 +270,10 @@ public struct CommandExtendedResult {
         }
     }
 
-    public init(result: CommandResult, exitStatus: Int32) {
+    public init(result: CommandResult, exitStatus: Int32, pid: Int32) {
```
dmbryson: If we were to keep this constructor, that pid arg should probably be a pid_t?.   However, I think this is left over from a transitionary step and is no longer used.  Probably cleaner to delete it rather than change its signature.

BenchR267: You are right, we should delete it. The command extended result should not be initialized with information not coming from an `llb_buildsystem_command_extended_result_t` instance.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

products/llbuildSwift/BuildSystemBindings.swift
BenchR267: Sorry, I had to add this back. It does make sense to open up this initializer since the client could define its own `ExternalCommand`s which should be able to produce extended results on them own. I also opened up the init of CommandMetrics since external commands should also be able to provide metrics over this mechanism.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -251,24 +251,32 @@ public struct CommandMetrics {
...
         switch self.result {
         case .succeeded, .failed:
             self.metrics = CommandMetrics(utime: result.pointee.utime, stime: result.pointee.stime, maxRSS: result.pointee.maxrss)
         default:
             self.metrics = nil
         }
     }
-
-    public init(result: CommandResult, exitStatus: Int32) {
+
```
jakepetroules: I'd recommend turning on "trim trailing whitespace" in your editor ;)

include/llbuild/BuildSystem/CommandResult.h
```diff
@@ -15,6 +15,12 @@
 
 #include <inttypes.h>
 
+#if defined(__linux__) || defined(__GNU__)
+#include <termios.h>
+#else
+#include <unistd.h>
+#endif
```
ddunbar: Why is termios needed?

BenchR267: I had a build error on Linux which suggeted to import termios for pid_t. If it can be imported with `sys/types.h` on macOS and Linux, we should use this 👍

ddunbar: Sounds like a bogus clang diagnostic to me... I don't think of termios.h as the canonical location for pid_t. There should be a cross platform header for this.

My general preference w.r.t. cross platform compatibility is that we avoid using platform-dependent types in the public headers, preferring to hide them in wrapper types with shims inside of Basic/, or making use of the LLVM shims we already have imported (where appropriate).

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

include/llbuild/BuildSystem/CommandResult.h
BenchR267: I just added a new header in Basic/ with a definition of llbuild_pid_t which is cross platform in 95bc9a70. Is there any way I can test this on Windows without installing Windows? :)

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

include/llbuild/Basic/CrossPlatformCompatibility.h
```diff
@@ -0,0 +1,28 @@
...
+// This file defines cross platform wrappers around non-cross platform types.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CrossPlatformCompatibility_h
+#define CrossPlatformCompatibility_h
+
+#if defined(_WIN32)
+#include <cstdint>
+typedef int32_t llbuild_pid_t;
```
jakepetroules: Actually we might want to use `ifdef _MSC_VER` here, since this typedef is only needed for MSVC. MinGW provides sys/types.h which already defines pid_t to an appropriate value.

include/llvm/Support/Program.h
```diff
@@ -33,7 +34,7 @@ namespace sys {
 /// @brief This struct encapsulates information about a process.
 struct ProcessInfo {
 #if defined(LLVM_ON_UNIX)
-  typedef pid_t ProcessId;
+  typedef llbuild_pid_t ProcessId;
 #elif defined(LLVM_ON_WIN32)
   typedef unsigned long ProcessId; // Must match the type of DWORD on Windows.
```
jakepetroules: Actually, noticing this now (sorry), why not just re-use the existing ProcessId typedef from here?

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

include/llvm/Support/Program.h
BenchR267: It is nested in a struct, so it's not visible from Swift :/

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
I didn't came up with a satisfying solution for the llbuild_pid_t typedef but at least with a working one ;) It's now defined twice: in CrossPlatformCompatibility.h and in the C-API-Header file as well which is used for bridging to Swift.

PR #305, opened 2018-05-10
[Core] Add safety checks to demandRule, guard taskInfos, and move all cancellation to executeTasks

dmbryson:


dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
```diff
@@ -1359,10 +1359,12 @@ class BuildEngineImpl : public BuildDBDelegate {
   RuleInfo& addRule(KeyID keyID, Rule&& rule) {
     auto result = ruleInfos.emplace(keyID, RuleInfo(keyID, std::move(rule)));
     if (!result.second) {
-      // FIXME: Error handling.
-      std::cerr << "error: attempt to register duplicate rule \""
-                << rule.key << "\"\n";
-      exit(1);
+      delegate.error("error: attempt to register duplicate rule \"" + rule.key + "\"\n");
```
BenchR267: I think the `error: ` prefix here is not needed since the delegate knows it's an error when receiving the method call. It should then decide on its own how to present this error.

lib/Core/BuildEngine.cpp
```diff
@@ -492,10 +492,17 @@ class BuildEngineImpl : public BuildDBDelegate {
...
 
     // Find the task info for this task.
     auto taskInfo = getTaskInfo(task);
     assert(taskInfo && "rule action returned an unregistered task");
+
+    // Same as the above assert, but for catching this in production as a
+    // clearer cause for demandRule crashes (maybe?)
+    if (!taskInfo)
+      abort();
+
```
ddunbar: I would drop this... I think we will be able to tell this is true from the exact crash anyway...

lib/Core/BuildEngine.cpp
dmbryson: I had dropped it, though all of the error messages below in the source file sent to the delegate also had the prefix, so I put it back.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #308, opened 2018-05-18
Add -Wimplicit-fallthrough to most C++ targets

dmbryson:
- Explicitly not used in llvmSupport, which has intentional use of
implicit fallthrough

### DIFF
--- a/lib/llvm/Support/Path.cpp
+++ b/lib/llvm/Support/Path.cpp
@@ -1019,6 +1019,7 @@ file_magic identify_magic(StringRef Magic) {
     case 0xc4: // ARMNT Windows
       if (Magic[1] == 0x01)
         return file_magic::coff_object;
+      [[clang::fallthrough]];
 
     case 0x90: // PA-RISC Windows
     case 0x68: // mc68K Windows
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,7 @@ else ()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnewline-eof -Wdeprecated-declarations")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Winvalid-offsetof")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnon-virtual-dtor")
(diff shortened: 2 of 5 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:
Yeah. I'll have to look up how to set per target flags in CMake...

dmbryson:
@swift-ci please smoke test

dmbryson:
Changed to xcconfig, added to CMakeLists.txt, and fixed up llvm files with attribute labels/pragmas where appropriate.

dmbryson:
@swift-ci please smoke test

lib/llvm/Support/Path.cpp
```diff
@@ -898,6 +898,9 @@ void directory_entry::replace_filename(const Twine &filename, file_status st) {
   Status = st;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wimplicit-fallthrough"
```
jakepetroules: Please use the fallthrough attribute here as well.

dmbryson: My thought here with the pragma was that this was imported code that is making rather extensive use of the fallthrough in a way that readability will not be improved with the attributes.  We'll be diverging from the imported source fairly extensively.

I don't disagree strongly enough not to make the changes, but I did select this purposely.

dmbryson: Alright, I am a bit confused at how clang processes this code. (previous comment retracted).  There is extensive use of fallthrough here, but it only warns on a single line that is exactly the same as other lines in this code.  I am rather dubious of the attribute use and warning processing here.

dmbryson: Ugh, actually, it appears to be non-deterministic in its registration of the warnings?

dmbryson: Yeah, there is really only 1 fall through. The multiple case labels should be referring to 1 block of code, which is not actually fall through. (update my understanding of how the code is processed). The oddity with the warning on my system notwithstanding, I'll post the change in a moment.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@jakepetroules BTW, thanks for the comments on this. I think I made a much more solid change overall. 👍

PR #309, opened 2018-05-18
xcconfig: add an optional include for locally overriding build settings

jakepetroules:


jakepetroules:
@swift-ci please smoke test

PR #310, opened 2018-05-18
De-confuse SKIP_INSTALL settings

jakepetroules:
We want to opt individual targets into installation, and of the
libraries, we want to install llbuild.framework, not llbuild.dylib.


jakepetroules:
@swift-ci please smoke test

PR #311, opened 2018-05-19
Remove a bunch of redundant settings from the project file

jakepetroules:
...and move a few key ones into xcconfig.


jakepetroules:
We should probably move even more of them around, but let's start with these redundant ones which don't change any behavior.

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please smoke test

dmbryson:
FWIW, I checkout out and verified these changes locally.  We unfortunately still don't have automated CI on the Xcode project (its been requested, but still in queue).

PR #313, opened 2018-05-23
Fix flaky phony.llbuild test

dmbryson:
This test is sensitive to scheduler ordering due to the way
lit/FileCheck tests work. This changes the test back to using the
fifo scheduler for greater consistency in CI (same was done for the
missing-inputs.llbuild test previously)

rdar://problem/40387374


dmbryson:
@swift-ci please smoke test

PR #314, opened 2018-05-24
[BuildSystem] Produced nodes should build when prior value was missing

dmbryson:
In situations where a node was previously a missing input, that fact may
be recorded in the build database. If subsequent builds add a rule that
may now produce that node, the system could get into a situation such
that 'missing input' errors are triggered even though we can
successfully build it. This changes produced nodes to cause their rule
to be built.

rdar://problem/40376404

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -709,6 +709,12 @@ public:
     if (value.isFailedInput())
       return false;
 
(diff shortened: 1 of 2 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
Rebase merges are fine with me.

PR #316, opened 2018-05-25
[perftests] Find ninja with xcrun when not in llbuild-test-tools

dmbryson:


dmbryson:
@swift-ci please smoke test

perftests/Xcode/PerfTests/NinjaPerfTests.mm
```diff
@@ -30,6 +30,15 @@ static void ExecuteShellCommand(const char *String) {
 
 @implementation NinjaPerfTests
 
++ (NSString*) findNinja {
+    NSString* overrideNinjaPath = [@(SRCROOT)
+                                   stringByAppendingPathComponent:@"llbuild-test-tools/utils/Xcode/ninja"];
```
jakepetroules: Is this even really needed?

dmbryson: I expect that the perf test bots are set up this way.  My goal here was to not change the existing behavior for those that have it set up as such, while adding a sane default.  You could conceivably want to override what version of ninja you want to compare against.

dmbryson:
Ping

PR #317, opened 2018-05-31
Address multiple static analysis warnings

dmbryson:
rdar://problem/40484309&40484331&40484338


dmbryson:
@swift-ci please smoke test

PR #318, opened 2018-06-04
[Tracing] Introduce os_signpost/os_log

BenchR267:
This commit introduces the usage of os_signpost and os_log to implement tracing in llbuild. os_signpost is part of macOS 10.14 and gives the ability to add more and structured information to tracing intervals.
We adopt this in llbuild to get a better understanding about the build system if tracing as a feature has been enabled. If tracing is disabled this should result in no-ops to not influence the performance of the system.

rdar://problem/40310911

### DIFF
--- a/include/llbuild/Basic/Tracing.h
+++ b/include/llbuild/Basic/Tracing.h
@@ -15,57 +15,99 @@
 
 #include "llvm/ADT/StringRef.h"
 
-// Allow vendored llbuild installations to provide custom tracing entry points
-#if __has_include("TracingMacros.h")
-#include "TracingMacros.h"
-#endif
+// os_signpost is included in mac OS 10.14, if that header is not available, we don't trace at all.
+#if __has_include(<os/signpost.h>)
+#include <os/signpost.h>
 
-#ifndef LLBUILD_TRACE_POINT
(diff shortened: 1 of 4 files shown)


ddunbar:
@swift-ci please smoke test

include/llbuild/Basic/Tracing.h
```diff
@@ -15,57 +15,99 @@
...
+  return generalLog;
+}
 
-#ifndef LLBUILD_TRACE_POINT
-#define LLBUILD_TRACE_POINT(kind, arg1, arg2, arg3, arg4) \
-    { (void)(kind); (void)(arg1); (void)(arg2); (void)(arg3); (void)(arg4); }
-#endif
+/// Returns a singleton id to use for all signpost calls during the lifetime of the program.
+API_AVAILABLE(macos(10.14), ios(12.0), tvos(12.0), watchos(5.0))
+static os_signpost_id_t signpost_id() {
```
achivetta: Why not use `OS_SIGNPOST_ID_EXCLUSIVE` here?

BenchR267: We can change that in the future, it doesn't make a difference semantically since the only requirement is uniqueness. But it would be better though!

PR #319, opened 2018-06-05
Use a better fitting name for the log subsystem

BenchR267:
llbuild is a generic low level build system and should therefore not be specific about any clients.


BenchR267:
@swift-ci please smoke test

PR #321, opened 2018-06-11
[BuildSystem] Guard killAfterTimeoutThread in LaneBasedExecutionQueue

dmbryson:
TSAN tests show a potential data race in cancellation. Add a guard
around the killAfterTimeoutThread variable to protect against it.

rdar://problem/40707314


dmbryson:
@swift-ci please smoke test

PR #326, opened 2018-06-15
Make llbuild.framework get the right content in its Info.plist

jakepetroules:
It should pick up the default CURRENT_PROJECT_VERSION from the project,
not be hardcoded to 1 at the target level. Also, make other appropriate
plist tweaks.


jakepetroules:
@swift-ci please smoke test

PR #327, opened 2018-06-15
Use the new build system for llbuild by default

jakepetroules:


jakepetroules:
@swift-ci please smoke test

PR #330, opened 2018-06-17
[UnitTests] Add the libllbuild header search path to the new target for C API unit tests

abertelrud:
This is necessary in order for them to find the C API headers, in order to be able to run the unit tests from the Xcode project.


dmbryson:
@swift-ci please smoke test

PR #331, opened 2018-06-18
[BuildSystem] Don't pass -num-threads for non-wmo invocations

aciidb0mb3r:
<rdar://problem/40178238> llbuild should not pass -num-threads to swiftc for non-wmo invocations


aciidb0mb3r:
@swift-ci please smoke test

PR #328, opened 2018-06-15
<rdar://41142590> Temporarily revert part of PR #320

jakepetroules:
This PR causes tasks to execute in the wrong order under certain
circumstances. Revert the offending code until we have a proper fix.

PR #334, opened 2018-06-19
[BuildSystem] Fix tests and comments due to #328 and #329

dmbryson:
Reverting directory contents behavior broke tests expecting the new
behavior. This change adds a test that directly exposes the bug that was
fixed in #328, and fixes the tests that were XFAIL'd in #329. It also
adds commentary explaining the problem in BuildSystem.cpp, as the
full solution is non-trivial.

rdar://problem/41182818


dmbryson:
@swift-ci please smoke test

PR #336, opened 2018-06-22
Alter directory-contents test to be more reliable in CI

dmbryson:
The second modification test failed in internal CI. The likely
explanation I can think of is that the system must still be HFS, thus
two close together touch commands did not alter the stat info. Change to
a file content change instead.

rdar://problem/41352819


dmbryson:
@swift-ci please smoke test

PR #337, opened 2018-06-22
Fix initializer order warning

dmbryson:
- Also, enable -Wall in Xcode so that this warning is triggered there


dmbryson:
@swift-ci please smoke test

PR #347, opened 2018-07-11
[CMake] Enable autolinking when compiling llbuildSwift

aciidb0mb3r:
This will let clients avoid passing -l flag when they import
llbuildSwift.


aciidb0mb3r:
@swift-ci please smoke test

PR #349, opened 2018-07-17
Set APPLICATION_EXTENSION_API_ONLY = YES

jakepetroules:
We don't use any APIs that are not app-extension-safe, so this is just
general best practice.


jakepetroules:
@swift-ci please smoke test

PR #350, opened 2018-07-17
[Core] Lazily open and automatically close SQLite DB connections

dmbryson:
Keeping the DB open grabs exclusive locks that prevents other processes
from accessing the DB, even when we are otherwise not using it. This
changes the driver to close the connection whenever a build has been
completed.

rdar://problem/42247751

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -73,6 +73,9 @@ class SQLiteBuildDB : public BuildDB {
   /// * 4: Pre-history
   static const int currentSchemaVersion = 8;
 
+  std::string path;
+  uint32_t clientSchemaVersion;
+
   sqlite3 *db = nullptr;
 
   /// The mutex to protect all access to the database and statements.
@@ -98,16 +101,10 @@ class SQLiteBuildDB : public BuildDB {
     return out;
   }
 
-public:
-  virtual ~SQLiteBuildDB() {
-    if (db)
-      close();
-  }
-
-  bool open(StringRef path, uint32_t clientSchemaVersion,
-            std::string *error_out) {
-    std::lock_guard<std::mutex> guard(dbMutex);
-    assert(!db);
+  bool open(std::string *error_out) {
+    // The db is opened lazily whenever an operation on it occurs. Thus if it is
+    // already open, we don't need to do any further work.
+    if (db) return true;
 
     // Configure SQLite3 on first use.
     //
@@ -124,7 +121,7 @@ public:
         }
     }
 
-    int result = sqlite3_open(path.str().c_str(), &db);
+    int result = sqlite3_open(path.c_str(), &db);
     if (result != SQLITE_OK) {
       *error_out = "unable to open database: " + std::string(
           sqlite3_errstr(result));
@@ -166,7 +163,7 @@ public:
       sqlite3_close(db);
 
       // Always recreate the database from scratch when the schema changes.
-      result = basic::sys::unlink(path.str().c_str());
+      result = basic::sys::unlink(path.c_str());
       if (result == -1) {
         if (errno != ENOENT) {
           *error_out = std::string("unable to unlink existing database: ") +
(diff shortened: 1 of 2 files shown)


dmbryson:
Checked with a root and confirmed that this resolves the problem in the referenced radar.

dmbryson:
@swift-ci please smoke test

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -98,16 +101,10 @@ class SQLiteBuildDB : public BuildDB {
...
-    if (db)
-      close();
-  }
-
-  bool open(StringRef path, uint32_t clientSchemaVersion,
-            std::string *error_out) {
-    std::lock_guard<std::mutex> guard(dbMutex);
-    assert(!db);
+  bool open(std::string *error_out) {
+    // The db is opened lazily whenever an operation on it occurs. Thus if it is
```
ddunbar: Don't we still want to grab the dbMutex here?

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -285,22 +282,38 @@ class SQLiteBuildDB : public BuildDB {
   }
 
   void close() {
-    std::lock_guard<std::mutex> guard(dbMutex);
-    assert(db);
+    if (!db) return;
```
ddunbar: Same comment here about the lock, although maybe it was never necessary?

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -285,22 +282,38 @@ class SQLiteBuildDB : public BuildDB {
...
     sqlite3_close(db);
     db = nullptr;
   }
 
+public:
+  SQLiteBuildDB(StringRef path, uint32_t clientSchemaVersion)
+    : path(path), clientSchemaVersion(clientSchemaVersion) { }
+
+  virtual ~SQLiteBuildDB() {
+    std::lock_guard<std::mutex> guard(dbMutex);
```
ddunbar: This is a case where I think you actually don't need the lock -- if the db is being destructed while something else is operating on it that is already a bug?

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -310,7 +323,11 @@ class SQLiteBuildDB : public BuildDB {
 
   virtual uint64_t getCurrentIteration(bool* success_out, std::string *error_out) override {
     std::lock_guard<std::mutex> guard(dbMutex);
-    assert(db);
+
+    if (!open(error_out)) {
+      *success_out = false;
+      return 0;
+    }
```
ddunbar: This works for now as a minimal fix, but it feels unfortunately we have to have the lazy behavior versus a more explicit open/close that could be propagated up to the `BuildEngine` -- did you explore how much work it would be to make this more explicit in the BuildDB API?

ddunbar: For example, if the engine ever decided to call this but never call build, then the db wouldn't be closed.

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -740,10 +777,7 @@ if (result != SQLITE_OK) { \
 std::unique_ptr<BuildDB> core::createSQLiteBuildDB(StringRef path,
                                                    uint32_t clientSchemaVersion,
                                                    std::string* error_out) {
-  auto db = llvm::make_unique<SQLiteBuildDB>();
-  if (!db->open(path, clientSchemaVersion, error_out))
-    return nullptr;
-
+  auto db = llvm::make_unique<SQLiteBuildDB>(path, clientSchemaVersion);
   return std::move(db);
```
ddunbar: Can this just be `return llvm::...` and avoid the need for the move?

lib/Core/SQLiteBuildDB.cpp
dmbryson: As discussed offline, yeah, this is a bit awkward and could definitely be helped with a more explicit API, such as a build context.  This is indeed intended to be a minimal fix for the time being.  Tracking the refactor in rdar://problem/42310556

PR #351, opened 2018-07-18
Add a Makefile to build llbuild using CMake

aciidb0mb3r:
This provides a convenient way to build and test llbuild using CMake.
Since it is possible to build llbuild using multiple buildsystems, we
can extend the Makefile as needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+all: cmake-build
+
+MAKE_FILE_PATH := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+SRCROOT := $(abspath $(MAKE_FILE_PATH))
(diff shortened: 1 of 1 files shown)


ddunbar:
(put another way, this feels very specific to a particular workflow... extending it to support other workflows feels like it would be tantamount to just reinventing the commands themselves). Is there a reason why this needs to be part of the project versus just a script of your own?

aciidb0mb3r:
The idea was to provide a convenient way for developers to build and test llbuild on desktops, similar to SwiftPM's Utilities/bootstrap script or Swift's utils/build-script. I don't have strong arguments for keeping it in llbuild (other than convenience).

PR #355, opened 2018-07-24
[CMake] Add dependency on the Swift compiler version

aciidb0mb3r:
<rdar://problem/42543712> Rebuild Swift bindings when Swift compiler changes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,6 +233,17 @@ if(LLBUILD_ENABLE_ASSERTIONS)
   string(REPLACE "-DNDEBUG" "" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
 endif()
 
+# Add an always out-of-date target to get the Swift compiler version.
+if (SWIFTC_FOUND)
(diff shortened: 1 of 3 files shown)


aciidb0mb3r:
@swift-ci please smoke test

jrose-apple:
Any reason to not just depend on the path to swiftc itself?

aciidb0mb3r:
@jrose-apple path doesn't change on the CI

jakepetroules:
> Any reason to not just depend on the path to swiftc itself?

In theory that's less reliable since the entirety of the compiler's implementation is not in the driver.

aciidb0mb3r:
@jrose-apple I am not sure if that works as the Swift compiler is not an output of the build process. This is what we do for in llbuild for Swift packages. I guess we can remove this once we have ABI

aciidb0mb3r:
https://github.com/apple/swift-llbuild/blob/master/lib/BuildSystem/BuildSystem.cpp#L2704

jrose-apple:
This isn't about ABI stability; it's module stability.

aciidb0mb3r:
2empag

aciidb0mb3r:
I don't think so. The Xcode build system should take care of that.

PR #356, opened 2018-07-30
[BuildSystem] Move Swift compiler's aux files next to the object file

aciidb0mb3r:
This is important for performing correct builds when there are two source files
with the same name in a Swift target.

<rdar://problem/41060279> Nest all swift auxiliary files similar to the object files


aciidb0mb3r:
@swift-ci smoke test

aciidb0mb3r:
Not sure if we can create a future proof test as the "failure" depends on the type of supplementary file.

jrose-apple:
I guess it would be hardcoding knowledge of the file map format, but you could check that no paths are duplicated throughout the file. (Honestly, the compiler could probably check that too, so maybe you should just file a JIRA bug and leave it at that.)

PR #358, opened 2018-08-02
Propose fix two typos

jeis2497052:
fix two typos


dmbryson:
@swift-ci please smoke test

PR #362, opened 2018-09-21
[BuildSystem] Allow processes to release lanes

dmbryson:
The general approach here is that LaneBasedExecutionQueue::executeProcess()
exposes a pipe file descriptor to the launched process. If that process
sends the value of LLBUILD_TASK_ID over that pipe, the process will
release the lane on which it is executing and asynchronously wait for
completion.

This initial implementation spawns a new thread for each process that
triggers the release. Longer term, more sophisticated handling of async
IO and process wait may be preferred.

Since any process may opt in to use this functionality, the chain of
methods that lead into the executeProcess call have all been modified
to take completion/result reporting callbacks.

rdar://problem/44683736

### DIFF
--- a/lib/BuildSystem/BuildExecutionQueue.cpp
+++ b/lib/BuildSystem/BuildExecutionQueue.cpp
@@ -17,6 +17,7 @@
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/StringRef.h"
 
+#include <future>
 #include <vector>
 
 using namespace llbuild;
@@ -32,7 +33,16 @@ BuildExecutionQueue::~BuildExecutionQueue() {
 
 CommandResult BuildExecutionQueue::executeProcess(
     QueueJobContext* context, ArrayRef<StringRef> commandLine) {
-  return executeProcess(context, commandLine, {});
+  // Promises are move constructible only, thus cannot be put into std::function
+  // objects that themselves get copied around. So we must create a shared_ptr
+  // here to allow it to go along with the labmda.
+  std::shared_ptr<std::promise<CommandResult>> p{new std::promise<CommandResult>};
+  auto result = p->get_future();
+  executeProcess(context, commandLine, {}, true, true,
+                 {[p](CommandResult result) mutable {
+    p->set_value(result);
+  }});
+  return result.get();
 }
 
 bool BuildExecutionQueue::executeShellCommand(QueueJobContext* context,
--- a/lib/BuildSystem/LaneBasedExecutionQueue.cpp
+++ b/lib/BuildSystem/LaneBasedExecutionQueue.cpp
@@ -33,16 +33,19 @@
 
 #include <atomic>
 #include <condition_variable>
+#include <cstdlib>
 #include <deque>
 #include <memory>
 #include <mutex>
 #include <queue>
+#include <random>
 #include <string>
 #include <thread>
 #include <unordered_map>
 #include <vector>
 
 #include <fcntl.h>
+#include <poll.h>
 #include <pthread.h>
 #include <unistd.h>
 #include <signal.h>
@@ -93,7 +96,8 @@ qos_class_t _getDarwinQOSClass(QualityOfService level) {
 #endif
 
 struct LaneBasedExecutionQueueJobContext {
-  uint32_t laneNumber;
+  uint64_t jobID;
+  uint64_t laneNumber;
   
   QueueJob& job;
 };
@@ -120,6 +124,9 @@ public:
 //
 // FIXME: Consider trying to share this with the Ninja implementation.
 class LaneBasedExecutionQueue : public BuildExecutionQueue {
+  /// (Random) build identifier
+  uint32_t buildID;
+
   /// The number of lanes the queue was configured with.
   unsigned numLanes;
 
@@ -147,7 +154,7 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
   /// The base environment.
   const char* const* environment;
   
-  void executeLane(unsigned laneNumber) {
+  void executeLane(uint32_t buildID, uint32_t laneNumber) {
     // Set the thread name, if available.
 #if defined(__APPLE__)
     pthread_setname_np(
@@ -165,7 +172,14 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
     pthread_set_qos_class_self_np(
         _getDarwinQOSClass(defaultQualityOfService), 0);
 #endif
-    
+
+    // Lane ID, used in creating reasonably unique task IDs, stores the buildID
+    // in the top 32 bits.  The laneID is stored in bits 16:31, and the job
+    // count is placed in the lower order 16 bits. These are not strictly
+    // guaranteed to be unique, but should be close enough for common use cases.
+    uint32_t jobCount = 0;
+    uint64_t laneID = (((uint64_t)buildID & 0xFFFF) << 32) + (((uint64_t)laneNumber & 0xFFFF) << 16);
+
     // Execute items from the queue until shutdown.
     while (true) {
       // Take a job from the ready queue.
@@ -191,7 +205,9 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
         break;
 
       // Process the job.
-      LaneBasedExecutionQueueJobContext context{ laneNumber, job };
+      jobCount++;
+      uint64_t jobID = laneID + jobCount;
+      LaneBasedExecutionQueueJobContext context{ jobID, laneNumber, job };
       {
         TracingExecutionQueueDepth(readyJobsCount);
 
@@ -240,13 +256,13 @@ public:
   LaneBasedExecutionQueue(BuildExecutionQueueDelegate& delegate,
                           unsigned numLanes,
                           const char* const* environment)
-      : BuildExecutionQueue(delegate), numLanes(numLanes),
+  : BuildExecutionQueue(delegate), buildID(std::random_device()()), numLanes(numLanes),
         readyJobs(Scheduler::make()), environment(environment)
   {
     for (unsigned i = 0; i != numLanes; ++i) {
       lanes.push_back(std::unique_ptr<std::thread>(
                           new std::thread(
-                              &LaneBasedExecutionQueue::executeLane, this, i)));
+                              &LaneBasedExecutionQueue::executeLane, this, buildID, i)));
     }
   }
   
@@ -303,25 +319,30 @@ public:
     }
   }
 
-  virtual CommandResult
-  executeProcess(QueueJobContext* opaqueContext,
-                 ArrayRef<StringRef> commandLine,
-                 ArrayRef<std::pair<StringRef,
-                                    StringRef>> environment,
-                 bool inheritEnvironment,
-                 bool canSafelyInterrupt) override {
+  virtual void executeProcess(
+      QueueJobContext* opaqueContext,
+      ArrayRef<StringRef> commandLine,
+      ArrayRef<std::pair<StringRef, StringRef>> environment,
+      bool inheritEnvironment,
+      bool canSafelyInterrupt,
+      llvm::Optional<std::function<void(CommandResult)>> completionFn) override {
+
     LaneBasedExecutionQueueJobContext& context =
       *reinterpret_cast<LaneBasedExecutionQueueJobContext*>(opaqueContext);
 
     llvm::SmallString<64> description;
     context.job.getForCommand()->getShortDescription(description);
-    TracingExecutionQueueSubprocess subprocessInterval(context.laneNumber, description.str());
+    TracingExecutionQueueSubprocess subprocessInterval(context.laneNumber,
+                                                       description.str());
 
     {
       std::unique_lock<std::mutex> lock(readyJobsMutex);
       // Do not execute new processes anymore after cancellation.
       if (cancelled) {
-        return CommandResult::Cancelled;
+        completionFn.unwrapIn([](CommandCompletionFn fn){
+          fn(CommandResult::Cancelled);
+        });
+        return;
       }
     }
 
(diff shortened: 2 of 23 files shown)


dmbryson:
@swift-ci please smoke test

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -303,25 +315,31 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
...
+
     LaneBasedExecutionQueueJobContext& context =
       *reinterpret_cast<LaneBasedExecutionQueueJobContext*>(opaqueContext);
 
     llvm::SmallString<64> description;
     context.job.getForCommand()->getShortDescription(description);
-    TracingExecutionQueueSubprocess subprocessInterval(context.laneNumber, description.str());
+    std::unique_ptr<TracingExecutionQueueSubprocess> subprocessInterval(
+         new TracingExecutionQueueSubprocess(context.laneNumber,
+                                             description.str()));
```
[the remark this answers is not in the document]
ddunbar: Can't we just copy this? It doesn't have interesting state, does it?

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -393,8 +423,27 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
     posix_spawn_file_actions_init(&fileActions);
 
     // Open /dev/null as stdin.
-    posix_spawn_file_actions_addopen(
-        &fileActions, 0, "/dev/null", O_RDONLY, 0);
+//    posix_spawn_file_actions_addopen(
+//        &fileActions, 0, "/dev/null", O_RDONLY, 0);
```
ddunbar: Probably want to hold off merge until we can fix this. :)

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -437,11 +490,11 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
...
     // Export a task ID to subprocesses.
-    //
-    // We currently only export the lane ID, but eventually will export a unique
-    // task ID for SR-6053.
-    posixEnv.setIfMissing("LLBUILD_TASK_ID", Twine(context.laneNumber).str());
-                          
+    auto taskID = Twine::utohexstr(context.jobID);
+    posixEnv.setIfMissing("LLBUILD_LANE_ID", Twine(context.laneNumber).str());
+    posixEnv.setIfMissing("LLBUILD_TASK_ID", taskID.str());
+    posixEnv.setIfMissing("LLBUILD_LANE_RELEASE_FD", Twine(releasePipe[1]).str());
```
ddunbar: Can we pick a different name, indicative of the idea that we would potentially use this as the general purpose communication channel (SR-6053)?

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -499,29 +552,58 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
...
-      return wasCancelled ? CommandResult::Cancelled : CommandResult::Failed;
+      auto result = wasCancelled ? CommandResult::Cancelled : CommandResult::Failed;
+      completionFn.unwrapIn([result](CommandCompletionFn fn){fn(result);});
+      return;
     }
 
+    // Set up our select() structures
+    pollfd readfds[] = {
+      { releasePipe[0], POLLIN, NULL },
+      { outputPipe[0], POLLIN, NULL }
```
ddunbar: Just for clarity, I would prefer we only set readfds[1].0 if shouldCaptureOuput is true.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -534,10 +616,98 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
             StringRef(buf, numBytes));
       }
 
-      // Close the read end of the pipe.
+      if (readfds[0].revents & POLLIN) {
+        ssize_t numBytes = read(releasePipe[0], buf, sizeof(buf));
```
ddunbar: Should we care about the case where we didn't get the full read? Also, I think it would be get to require the id be newline terminated.

Let's also make sure we document the protocol somewhere as part of this PR.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -534,10 +616,98 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
...
+
+        if (numBytes == 0) {
+          readfds[0].events = NULL;
+          if (shouldCaptureOutput) {
+            continue;
+          }
+          break;
+        }
+
+        auto msg = StringRef(buf, numBytes).trim();
```
ddunbar: I would prefer we not trim, but check for a very precise & strict protocol.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -534,10 +616,98 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
...
+        SmallString<16> Vec;
+        if (taskID.toStringRef(Vec).equals_lower(msg)) {
+          // Close the read end of the release pipe.
+          ::close(releasePipe[0]);
+
+          std::thread([
+              this, pid, handle, command=context.job.getForCommand(), outputPipe{outputPipe[0]},
+              completionFn, subprocessInterval{std::move(subprocessInterval)}
+            ]() mutable {
+            if (outputPipe != -1)
```
ddunbar: Would prefer to use `shouldCaptureOutput`. Esp. since outputPipe is an array 😛

tests/BuildSystem/Build/file-system-opts.llbuild
```diff
@@ -13,7 +13,7 @@
 # CHECK-NOT: PATH=
 # CHECK-NEXT: /usr/bin/env
 # CHECK: ENV_KEY=ENV_VALUE_2
-# CHECK: LLBUILD_TASK_ID=0
+# CHECK: LLBUILD_TASK_ID={{.*}}
```
ddunbar: [0-9]+?

lib/BuildSystem/LaneBasedExecutionQueue.cpp
dmbryson: Well, in this case the lambda has just the fd.  Should rename it for clarity.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
dmbryson: The trim was to allow for easy of use with things such as "echo ${LLBUILD_TASK_ID} >&${LLBUILD_LANE_RELEASE_FD}".   I could explicitly check to newline terminated values, though.

lib/BuildSystem/BuildExecutionQueue.cpp
```diff
@@ -32,7 +33,16 @@ BuildExecutionQueue::~BuildExecutionQueue() {
 
 CommandResult BuildExecutionQueue::executeProcess(
     QueueJobContext* context, ArrayRef<StringRef> commandLine) {
-  return executeProcess(context, commandLine, {});
+  // Promises are move constructible only, thus cannot be put into std::function
+  // objects that themselves get copied around. So we must create a shared_ptr
+  // here to allow it to go along with the labmda.
+  std::shared_ptr<std::promise<CommandResult>> p{new std::promise<CommandResult>};
```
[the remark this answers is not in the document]
dmbryson: You can std::move it in to the lambda, but of course you then need to std::move the lambda around too.  I'm sure this could be done, though it complicates the common case.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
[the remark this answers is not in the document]
dmbryson: Hmm, very true about the allocation.   The problem here is that the destructor is what triggers the completion of the trace event.  Maybe implement a move constructor and a 'is active copy' bool.  That way we could std::move it without allocating it on the heap.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
dmbryson: Oops.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
dmbryson: Sure.  LLBUILD_CONTROL_FD?

lib/BuildSystem/LaneBasedExecutionQueue.cpp
dmbryson: In the spirit of generality (re: SR-6053), should this also be something like "release ${LLBUILD_TASK_ID}\n"?

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -240,13 +252,13 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
   LaneBasedExecutionQueue(BuildExecutionQueueDelegate& delegate,
                           unsigned numLanes,
                           const char* const* environment)
-      : BuildExecutionQueue(delegate), numLanes(numLanes),
+  : BuildExecutionQueue(delegate), baseID(((uint64_t)std::random_device()() & 0xFFFF) << 48), numLanes(numLanes),
```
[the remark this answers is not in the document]
dmbryson: Could do. Perhaps rename it 'buildID' or 'instanceID' and then the task ID is then a combination of the buildID, laneID, and task count.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
ddunbar: I suspect we will want to change the binary protocol, so I'd probably recommend something like
`v1\n${LLBUILD_TASK_ID}\n`. Maybe `llbuild.v1` if we want it to be less ambiguous. I don't think the overhead of the extra bytes will matter, but it will be easier to rev in the future.

Also, you can use printf(1) for more precise output in a shell script.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
```diff
@@ -393,8 +423,27 @@ class LaneBasedExecutionQueue : public BuildExecutionQueue {
...
+          Twine("unable to open lane release pipe (") + strerror(errno) + ")");
+      getDelegate().commandProcessFinished(context.job.getForCommand(),
+                                           handle,
+                                           CommandExtendedResult::makeFailed());
+      completionFn.unwrapIn([](CommandCompletionFn fn){
+        fn(CommandResult::Failed);
+      });
+      return;
+    }
+    posix_spawn_file_actions_addinherit_np(&fileActions, releasePipe[1]);
```
dmbryson: Heh, and now I know why the others are using _adddup2.  Linux doesn't support addinherit_np.

dmbryson:
Cool, the swift compiler crashed when building on OS X.

lib/BuildSystem/LaneBasedExecutionQueue.cpp
ddunbar: np stands for non portable. :)

dmbryson: Ohhhh.  🤯

ddunbar:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

dmbryson:

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

dmbryson:

dmbryson:

dmbryson:

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
*Finally*

jrose-apple:
:-( Looks like this broke Ubuntu 14.04: https://ci.swift.org/view/Dashboard/job/oss-swift-incremental-RA-linux-ubuntu-14_04/5965/. I'm going to revert, since it's late on a Friday.

PR #366, opened 2018-09-25
[libllbuildSwift] Remove version protection from Swift bindings

aciidb0mb3r:
The versioning is not really very useful as there is not really
a revlock issue. The Swift bindings are built together with the
C bindings and we don't have any intention of using a different version
of libllbuild with libllbuildSwift. This commit hardcodes the library
version in llbuild.h and removes the generation scripts and targets.


ddunbar:
@swift-ci smoke test macOS

ddunbar:
@dmbryson can you review? I'm not sure I completely understand all the moving parts involved.

aciidb0mb3r:
Great, thanks!!!

PR #368, opened 2018-09-25
[BuildSystem] Split out filtered directory nodes

dmbryson:
This change moves filtered directory support into a parallel tree of
nodes. It also introduces a new 'stat' node that produces the stat
information for a file system object.  The combination of these changes
allows filtered directories to work as intended, depending upon the
filter contents only, and not the stat info of the containing directory.

Supports rdar://problem/30640904

### DIFF
--- a/include/llbuild/BuildSystem/BuildNode.h
+++ b/include/llbuild/BuildSystem/BuildNode.h
@@ -103,6 +103,20 @@ public:
   basic::FileInfo getLinkInfo(basic::FileSystem&) const;
 };
 
+
+class StatNode {
+  std::string name;
+
+public:
+  explicit StatNode(StringRef name) : name(name) {}
+
+  const std::string& getName() { return name; }
+
+  basic::FileInfo getFileInfo(basic::FileSystem&) const;
+  basic::FileInfo getLinkInfo(basic::FileSystem&) const;
+};
+
+
 }
 }
 
--- a/include/llbuild/BuildSystem/BuildValue.h
+++ b/include/llbuild/BuildSystem/BuildValue.h
@@ -90,6 +90,10 @@ class BuildValue {
 
     /// Sentinel value representing the result of "building" a top-level target.
     Target,
+
+    /// The filtered contents of a directory.
+    FilteredDirectoryContents,
+
   };
   static StringRef stringForKind(Kind);
 
@@ -121,12 +125,12 @@ class BuildValue {
   basic::StringList stringValues;
 
   bool kindHasCommandSignature() const {
-    return isSuccessfulCommand() || isDirectoryTreeSignature() ||
-      isDirectoryTreeStructureSignature();
+    return isSuccessfulCommand() ||
+        isDirectoryTreeSignature() || isDirectoryTreeStructureSignature();
   }
 
   bool kindHasStringList() const {
-    return isDirectoryContents() || isStaleFileRemoval();
+    return isDirectoryContents() || isFilteredDirectoryContents() || isStaleFileRemoval();
   }
 
   bool kindHasOutputInfo() const {
@@ -172,7 +176,6 @@ private:
     assert(kindHasStringList());
   }
 
-  
   std::vector<StringRef> getStringListValues() const {
     assert(kindHasStringList());
     return stringValues.getValues();
@@ -287,6 +290,9 @@ public:
   static BuildValue makeStaleFileRemoval(ArrayRef<std::string> values) {
     return BuildValue(Kind::StaleFileRemoval, values);
   }
+  static BuildValue makeFilteredDirectoryContents(ArrayRef<std::string> values) {
+    return BuildValue(Kind::FilteredDirectoryContents, values);
+  }
 
   /// @}
 
@@ -317,9 +323,12 @@ public:
   bool isCancelledCommand() const { return kind == Kind::CancelledCommand; }
   bool isSkippedCommand() const { return kind == Kind::SkippedCommand; }
   bool isTarget() const { return kind == Kind::Target; }
+  bool isFilteredDirectoryContents() const {
+    return kind == Kind::FilteredDirectoryContents;
+  }
 
   std::vector<StringRef> getDirectoryContents() const {
-    assert(isDirectoryContents() && "invalid call for value kind");
+    assert((isDirectoryContents() || isFilteredDirectoryContents()) && "invalid call for value kind");
     return getStringListValues();
   }
 
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -124,6 +124,9 @@ class BuildSystemEngineDelegate : public BuildEngineDelegate {
   // FIXME: This is an inefficent map, the string is duplicated.
   std::unordered_map<std::string, std::unique_ptr<BuildNode>> dynamicNodes;
 
+  // FIXME: This is an inefficent map, the string is duplicated.
+  std::unordered_map<std::string, std::unique_ptr<StatNode>> dynamicStatNodes;
+
   /// The custom tasks which are owned by the build system.
   std::vector<std::unique_ptr<Command>> customTasks;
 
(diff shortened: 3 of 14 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please test macos

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

include/llbuild/BuildSystem/BuildDescription.h
```diff
@@ -149,6 +149,34 @@ class Node {
       ArrayRef<std::pair<StringRef, StringRef>> values) = 0;
 };
 
+/// Definition for a file system stat used by the build file.
+class Stat {
+  // DO NOT COPY
+  Stat(const Stat&) LLBUILD_DELETED_FUNCTION;
```
jakepetroules: We really should just have a macro for `delete`ing these functions, like `LLBUILD_DISABLE_COPY(typename)`

include/llbuild/BuildSystem/BuildNode.h
```diff
@@ -103,6 +103,24 @@ class BuildNode : public Node {
   basic::FileInfo getLinkInfo(basic::FileSystem&) const;
 };
 
+
+class BuildStat : public Stat {
+public:
+  explicit BuildStat(StringRef name) : Stat(name) {}
+
+  virtual bool configureAttribute(const ConfigureContext& ctx, StringRef name,
```
jakepetroules: `virtual` is redundant for these

include/llbuild/BuildSystem/BuildDescription.h
```diff
@@ -149,6 +149,34 @@ class Node {
       ArrayRef<std::pair<StringRef, StringRef>> values) = 0;
 };
 
+/// Definition for a file system stat used by the build file.
+class Stat {
```
ddunbar: I'm not sure why we would need this in the build description layer. I was expecting this to just be a node created transparently by the filtering logic, but not exposed in the BuildDescription.

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -172,7 +180,14 @@ class BuildValue {
     assert(kindHasStringList());
   }
 
-  
+  BuildValue(Kind kind, basic::CommandSignature filters, ArrayRef<std::string> values)
```
ddunbar: Shouldn't `filter` be called signature here?

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -287,6 +302,13 @@ class BuildValue {
   static BuildValue makeStaleFileRemoval(ArrayRef<std::string> values) {
     return BuildValue(Kind::StaleFileRemoval, values);
   }
+  static BuildValue makeFilteredDirectoryContents(basic::CommandSignature filters,
```
ddunbar: Same comment about the name of the signature.

include/llbuild/BuildSystem/BuildDescription.h
ddunbar: To be more explicit, it would have a custom task implementation, and need a BuildKey entry, but it shouldn't surface in the BuildDescription.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -730,41 +779,13 @@ class ProducedNodeTask : public Task {
...
-    // changed. This causes unnecessary rebuilds. See rdar://problem/30640904
-    //
-    // * The 'final state' of a directory is also an thorny patch of toxic land
-    // mines. We really want directory contents to weakly depend upon anything
-    // that is currently and/or may be altered within it. i.e. if one rule
-    // creates the directory and another rule writes a file into it, we want to
-    // defer scanning until both of them have been scanned and possibly run.
-    // Having a 'must scan after' would help with the first rule (mkdir), but
-    // not the second, in particular if rules are added in subsequent builds.
-    // Related rdar://problem/30638921
```
ddunbar: Are you sure it is worth deleting this entire comment? Some parts of it remain true.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -885,6 +887,123 @@ class DirectoryContentsTask : public Task {
...
+
+  /// The value for the input directory.
+  BuildValue directoryValue;
+
+  virtual void start(BuildEngine& engine) override {
+    // Depend on the node so that it can be connected up to potential producers.
+    engine.taskNeedsInput(
+        this, BuildKey::makeNode(path).toData(), /*inputID=*/0);
+    // Depend on the raw stat information, because we want to know if we need
+    // to rescan items under this if anything has changed.
```
ddunbar: I'd comment here about the subtle dependency on linear scanning

ddunbar: Actually, interestingly enough... in this particular case we could make it "non-subtle" by moving the additional input request into provideValue, although I don't think I'd actually recommend that at the moment.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -885,6 +887,123 @@ class DirectoryContentsTask : public Task {
...
+        this, BuildKey::makeStat(path).toData(), /*inputID=*/1);
+  }
+
+  virtual void providePriorValue(BuildEngine&,
+                                 const ValueType& value) override {
+  }
+
+  virtual void provideValue(BuildEngine&, uintptr_t inputID,
+                            const ValueType& value) override {
+    // If our value is uninitialized, take whatever build value came in
```
ddunbar: Hmmm, I'm not sure about this. Why wouldn't we just always take the directoryValue from `inputID == 1`?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -885,6 +887,123 @@ class DirectoryContentsTask : public Task {
...
+
+    auto& info = directoryValue.getOutputInfo();
+
+    // Non-directory things are just plain-ol' inputs
+    if (!info.isDirectory()) {
+      engine.taskIsComplete(this, BuildValue::makeExistingInput(info).toData());
+      return;
+    }
+
+    // Collec the filtered contents
```
ddunbar: Sic collect

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -885,6 +887,123 @@ class DirectoryContentsTask : public Task {
...
+    std::vector<std::string> filenames;
+    getFilteredContents(path, filters, filenames);
+
+    // Compute the filters signature
+    using llvm::hash_combine;
+    llvm::hash_code code = hash_value(path);
+    for (const auto& filter: filters.getValues()) {
+      code = hash_combine(code,
+                          hash_combine_range(filter.begin(), filter.end()));
+    }
```
ddunbar: I don't think we actually need a signature for the filters. Why would we? The filters are a part of the build key?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -1186,8 +1296,153 @@ class DirectoryTreeStructureSignatureTask : public Task {
...
 
 public:
-  DirectoryTreeStructureSignatureTask(StringRef path, StringList&& filters)
-    : path(path), filters(std::move(filters)) {}
+  DirectoryTreeStructureSignatureTask(StringRef path) : path(path) {}
+};
+
+/// This is the task to "build" a directory node which will encapsulate (via a
+/// signature) a filtered view of the contents of the directory, recursively.
+class FilteredDirectoryTreeSignatureTask : public Task {
```
ddunbar: I do wonder if we couldn't have kept a single task implementation, and just had the constructor take the filter list and then switch on which kind of nodes to request?

tests/BuildSystem/Build/directory-contents-filtered.llbuild
```diff
@@ -17,18 +17,17 @@
 # RUN: %{llbuild} buildsystem build --serial --trace %t.rebuild.trace --chdir %t.build
 # RUN: %{FileCheck} --check-prefix=CHECK-REBUILD --input-file=%t.rebuild.trace %s
 #
-# CHECK-REBUILD: { "new-rule", "[[RULE_NAME:R[0-9]+]]", "Ddirexcludedfile" },
+# CHECK-REBUILD: { "new-rule", "[[RULE_NAME:R[0-9]+]]", "Fdirexcludedfile" },
 # CHECK-REBUILD: { "rule-does-not-need-to-run", "[[RULE_NAME]]" },
 
-# A rebuild unfortunately rebuilds with even with excluded files.
-# rdar://41142590
+# A rebuild shouldn't rebuild the directory for excluded files.
```
ddunbar: 👏

lib/BuildSystem/BuildSystem.cpp
dmbryson: Ah, fair point.

lib/BuildSystem/BuildSystem.cpp
dmbryson: Yeah, this was the point I was most interested in. I think, based on our discussion of the serial nature of the dependencies, that inputID == 1 should be sufficient.

lib/BuildSystem/BuildSystem.cpp
dmbryson: Yeah, in the end the differences are not as great as I originally thought at the actual tree signature level.   It is probably worth combining the back together.

include/llbuild/BuildSystem/BuildDescription.h
dmbryson: (though out of scope for this change, indeed this specific code will be removed in a moment)

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

dmbryson:

PR #372, opened 2018-09-29
[tests] Remove 'cat' debug from tests, as it is causing spurious failures in CI

dmbryson:


dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please test

aciidb0mb3r:
I think there is a problem with Python installation in some of the nodes. /cc @shahmishal

dmbryson:
Yeah, some of the nodes have an installation of lit that appears to be missing the “builtin” commands. I do wonder if something has broken in the lit python package, as one of my local VMs has the same problem (but others are fine). 

There is already a radar about it, but since these were just debug statements I thought it safe to remove them in the meantime.

PR #375, opened 2018-10-02
[Core] Bump SQLite DB schema version

dmbryson:
Filtered directory contents added and changed some key defintions such
that older versions are incompatible.

rdar://problem/44872576


dmbryson:
@swift-ci please smoke test

ddunbar:
Ooops!

We really need to get some tests in place that do some work and validate that either the DB is the same, or the schema changed, or something.

PR #377, opened 2018-10-02
[llbuildSwift] Fix TaskWrapper Memory Leak

briantkelley:
Add a destroy_context callback to the C API's task delegate, which, like the build engine's delegate, is called from the destructor. Add the appropriate implementation in the Swift core bindings to release the TaskWrapper associated with the C API's task delegate so it no longer leaks.

### DIFF
--- a/products/libllbuild/Core-C-API.cpp
+++ b/products/libllbuild/Core-C-API.cpp
@@ -118,6 +118,12 @@ public:
              "missing task inputs_available function");
   }
 
+  virtual ~CAPITask() {
+    if (cAPIDelegate.destroy_context) {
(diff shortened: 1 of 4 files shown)


ddunbar:
@swift-ci please smoke test

dmbryson:
I think our known downstream clients now use the Swift bindings we build here, so it is not strictly necessary.   That said, I'd recommend bumping it anyway for clarity and adding the comment to the header describing the change.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #379, opened 2018-10-09
[Basic] Move control release check outside poll loop

dmbryson:
When a complete control message arrives in one read, the poll loop
early exits. As such, lane release may not happen until a subsequent
read occurs on the console pipe. Move the check out of the loop so that
we properly release after the current reads have been finished.

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -523,29 +523,26 @@ void llbuild::basic::spawnProcess(
           readfds[i].events = 0;
           continue;
         }
-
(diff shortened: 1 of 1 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:
It is likely related to the fact that the test does an echo before releasing the lane.  In my local testing with builds using this feature, writing something to the console pipe before the lane release increased the success rate. The two writes are commonly delivered to llbuild in the same poll() response.  Some small changes to that test may more reliably catch that, but should probably be investigated before doing so (rdar://problem/45162480).

PR #381, opened 2018-10-11
[Basic] Hold lane release pipe open until process termination

dmbryson:
Clients (especially bash on Linux) may receive SIGPIPE when we
close the lane release pipe immediately. Although we could require
clients to implement their own SIGPIPE traps, it is simpler to hold the
pipe open until we are sure the client is done.

rdar://problem/45196629

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -21,7 +21,6 @@
 #include "llvm/Support/Program.h"
 
 #include <atomic>
-#include <iostream>
 #include <thread>
 
 #include <fcntl.h>
@@ -209,6 +208,8 @@ static void captureExecutedProcessOutput(
     delegate.processHadOutput(ctx, handle, StringRef(buf, numBytes));
   }
 
+  // We have receieved the zero byte read that indicates an EOF. Go ahead and
+  // close the pipe.
   ::close(outputPipe);
 }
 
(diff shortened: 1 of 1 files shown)


dmbryson:
@swift-ci please smoke test

ddunbar:
Yeah, this makes sense to me.

lib/Basic/Subprocess.cpp
```diff
@@ -256,6 +261,7 @@ static void cleanUpExecutedProcess(
   ProcessStatus processStatus = cancelled ? ProcessStatus::Cancelled : (exitCode == 0) ? ProcessStatus::Succeeded : ProcessStatus::Failed;
   ProcessResult processResult(processStatus, exitCode, pid, utime, stime,
                               usage.ru_maxrss);
+  std::cout << "process finished: " << std::dec << exitCode << std::endl;
```
ddunbar: Stray debug log?

lib/Basic/Subprocess.cpp
```diff
@@ -21,6 +21,7 @@
 #include "llvm/Support/Program.h"
 
 #include <atomic>
+#include <iostream>
```
ddunbar: Stray include (I despise iostream, it adds tons of bloat).

lib/Basic/Subprocess.cpp
```diff
@@ -528,19 +534,17 @@ void llbuild::basic::spawnProcess(
     }
 
     if (control.shouldRelease()) {
-      // Close the read end of the release pipe.
-      ::close(releasePipe[0]);
-
       releaseFn([
```
ddunbar: Might be worth either putting a comment here, or at the other close site, explaining why we close when we do.

lib/Basic/Subprocess.cpp
dmbryson: definitely, that was also stray debug

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #382, opened 2018-10-14
Implement i686 Linux (32-bit x86) Support

Kaiede:
i686 mostly works already, minus a couple issues that need to be fixed for a build to complete successfully.

1) Subprocess.h introduced a uintptr_t property of ProcessHandle as uint64_t instead, which is correctly caught by the compiler on 32-bit platforms.
2) i686 uses libatomic instead of built-ins, and this needs to be correctly detected. The implementation behind this was somewhat shamelessly borrowed from LLVM, as I wasn't able to find a cleaner way to simply ask LLVM itself about the state of libatomic.

### DIFF
--- a/cmake/modules/CheckAtomic.cmake
+++ b/cmake/modules/CheckAtomic.cmake
@@ -1,0 +1,106 @@
+# atomic builtins are required for threading support.
+
+INCLUDE(CheckCXXSourceCompiles)
+INCLUDE(CheckLibraryExists)
+
+# Sometimes linking against libatomic is required for atomic ops, if
+# the platform doesn't support lock-free atomics.
+
+function(check_working_cxx_atomics varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++11")
+  CHECK_CXX_SOURCE_COMPILES("
+#include <atomic>
+std::atomic<int> x;
+int main() {
+  return x;
+}
+" ${varname})
+  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+endfunction(check_working_cxx_atomics)
+
+function(check_working_cxx_atomics64 varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "-std=c++11 ${CMAKE_REQUIRED_FLAGS}")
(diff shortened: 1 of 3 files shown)


ddunbar:
Awesome, thanks!

ddunbar:
@swift-ci please smoke test

ddunbar:
Shamelessly borrowing from LLVM is fine, its at least in keeping with what we do so far. :)

Kaiede:
> Shamelessly borrowing from LLVM is fine, its at least in keeping with what we do so far. :)

Yeah, it duplicates some scripting, but it's at least using the same cmake file. So it should be possible to figure out how to unify the two copies of CheckAtomic.cmake down the road, if that's even desirable here.

ddunbar:
@swift-ci please smoke test

Kaiede:
@jakepetroules Fun fact, Debian Stretch on 32-bit x86 uses i686 as the triple, which is what this patch is mostly aimed at enabling. I figured I'd call out the triple explicitly, rather than call it 'i386', and confuse someone who might be on a distro that does use i386 as the triple.

ddunbar:
@dmbryson this is failing in a lane release test -- it seems unlikely this patch would have regressed this; is it possible this is still flaky? Can you take a look?

dmbryson:
Yeah, started to take a look yesterday and indeed nothing in the change caused a detectable regression locally.  Of course, that makes this test all the more annoying to debug, as I literally ran 10000 iterations of the test before committing the latest change and it STILL has transient failures in Swift-CI... sigh.

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

jakepetroules:
Shall we try to move this forward?

dmbryson:
I did pull this down and investigate it a bit.  At the time it seemed to legitimately break the tests, but I did not have the chance to investigate it further. It remains on my to do list.

Kaiede:
Ugh, if it does break the tests, is the semaphore relying on something that libatomic implements? If so, I wonder if macOS is getting the libatomic implementation instead of the built-in somehow?

ddunbar:
Is anyone planning on picking this up? If not, I’d like to close it out until someone is actually going to give it attention.

Kaiede:
I'm fine closing it out for now. i686 is secondary to ARM 32 for me, and uraimo picked up the important part in a different PR.

atommclain:
I noticed a change to the accompanying [CheckAtomic.cmake](https://github.com/llvm/llvm-project/commit/a348de480d34257ffe394ab51b880c9aef243d80) file in the llvm repo. It looks like the change was related to the logic of atomics in MSVC.

Is it possible that this is what was causing the failures in the Swift-CI?

PR #386, opened 2018-10-23
[BuildSystem] Fix filtered directory path references

dmbryson:


dmbryson:
@swift-ci please smoke test

PR #387, opened 2018-10-26
Avoid escaping array-to-pointer conversions

hamishknight:
Such conversions are only guaranteed to be valid for the duration of the call they're passed to (in this case `UnsafePointer`'s initialiser). Instead, allocate a buffer pointer which we can get the base address from.

(This is currently blocking https://github.com/apple/swift/pull/20070 from being able to build).


dmbryson:
@swift-ci please smoke test

hamishknight:
@dmbryson Thanks for the quick review! Is this good to merge?

ddunbar:
Thanks @hamishknight!

PR #394, opened 2018-11-09
[BuildSystem] Add support for setting external process working directory

dmbryson:
This currently is limited to Darwin platforms, as it is implemented via a
non-portable pthread API.

rdar://problem/39941219

### DIFF
--- a/include/llbuild/Basic/Subprocess.h
+++ b/include/llbuild/Basic/Subprocess.h
@@ -224,6 +224,19 @@ namespace llbuild {
                                    const ProcessResult& result) = 0;
     };
 
+
+    struct ProcessAttributes {
+      /// If true, whether it is safe to attempt to SIGINT the process to cancel
+      /// it. If false, the process won't be interrupted during cancellation and
+      /// will be given a chance to complete (if it fails to complete it will
+      /// ultimately be sent a SIGKILL).
+      bool canSafelyInterrupt;
+
+      /// If set, the working directory to change into before spawning (only
+      /// supported on macOS)
+      StringRef workingDir = {};
+    };
+
     /// Execute the given command line.
     ///
     /// This will launch and execute the given command line and wait for it to
@@ -241,10 +254,7 @@ namespace llbuild {
     ///
     /// \param environment The environment to launch with.
     ///
-    /// \param canSafelyInterrupt If true, whether it is safe to attempt to SIGINT
-    /// the process to cancel it. If false, the process won't be interrupted
-    /// during cancellation and will be given a chance to complete (if it fails to
-    /// complete it will ultimately be sent a SIGKILL).
+    /// \param attributes Additional attributes for the process to be spawned.
     ///
     /// \param releaseFn Functional called when a process wishes to release its
     /// exclusive access to build system resources (namely an execution lane).
@@ -264,7 +274,7 @@ namespace llbuild {
                       ProcessHandle handle,
                       ArrayRef<StringRef> commandLine,
                       POSIXEnvironment environment,
-                      bool canSafelyInterrupt,
+                      ProcessAttributes attributes,
                       ProcessReleaseFn&& releaseFn,
                       ProcessCompletionFn&& completionFn);
 
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -33,6 +33,27 @@
 
 #ifdef __APPLE__
 #include <pthread/spawn.h>
+
+extern "C" {
+  // Provided by System.framework's libsystem_kernel interface
+  extern int __pthread_chdir(const char *path);
+  extern int __pthread_fchdir(int fd);
+}
+
+/// Set the thread specific working directory to the given path.
+int pthread_chdir_np(const char *path)
+{
+  return __pthread_chdir(path);
+}
+
+/// Set the thread specific working directory to that of the given file
+/// descriptor. Passing -1 clears the thread specific working directory,
+/// returning it to the process level working directory.
+int pthread_fchdir_np(int fd)
+{
+  return __pthread_fchdir(fd);
+}
+
 #endif
 
 using namespace llbuild;
@@ -278,7 +299,7 @@ void llbuild::basic::spawnProcess(
     ProcessHandle handle,
     ArrayRef<StringRef> commandLine,
     POSIXEnvironment environment,
-    bool canSafelyInterrupt,
+    ProcessAttributes attr,
     ProcessReleaseFn&& releaseFn,
     ProcessCompletionFn&& completionFn
 ) {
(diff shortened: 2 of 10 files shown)


lib/Basic/Subprocess.cpp
```diff
@@ -436,17 +452,47 @@ void llbuild::basic::spawnProcess(
...
+
+#ifdef __APPLE__
+      int oldCWD = -1;
+      int chdirErr = 0;
+      if (attr.workingDir.size()) {
+        oldCWD = open(".", O_RDONLY);
+        if (pthread_fchdir_np(-1) == -1 && oldCWD != -1) {
+          // no existing per-thread cwd
+          close(oldCWD); oldCWD = -1;
+        }
```
ddunbar: I think this is unnecessary -- we own the thread, so we should be able to keep track directly of what the latest per-thread WD was. I'd like it if we could avoid any extraneous sys calls in the common case where almost all commands will run in the same directory.

lib/Basic/Subprocess.cpp
```diff
@@ -436,17 +452,47 @@ void llbuild::basic::spawnProcess(
...
     if (!wasCancelled) {
-      int result = posix_spawn(&pid, args[0], /*file_actions=*/&fileActions,
-                               /*attrp=*/&attributes, const_cast<char**>(args.data()),
-                               const_cast<char* const*>(environment.getEnvp()));
+      int result = 0;
+
+#ifdef __APPLE__
+      int oldCWD = -1;
+      int chdirErr = 0;
+      if (attr.workingDir.size()) {
```
ddunbar: We should report an error here for non-__APPLE__, if workingDir is set.

lib/Basic/Subprocess.cpp
```diff
@@ -436,17 +452,47 @@ void llbuild::basic::spawnProcess(
...
+      if (result == 0) {
+        result = posix_spawn(&pid, args[0], /*file_actions=*/&fileActions,
+                             /*attrp=*/&attributes, const_cast<char**>(args.data()),
+                             const_cast<char* const*>(environment.getEnvp()));
+      }
+
+#ifdef __APPLE__
+      if (attr.workingDir.size()) {
+        pthread_fchdir_np(oldCWD);
+        if (oldCWD != -1) close(oldCWD);
```
ddunbar: presumably doesn't make sense to try to chdir to -1.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2170,6 +2173,13 @@ class ShellCommand : public ExternalCommand {
         return false;
       }
       inheritEnv = value == "true";
+    } else if (name == "working-directory") {
+#if __APPLE__
+      workingDirectory = value;
+#else
+      ctx.error("working-directory unsupported on this platform");
```
ddunbar: Ah, I see you put this check here. In that case, I'd just add an assert for non-Apple platforms.

tests/BuildSystem/Build/working-directory.llbuild
```diff
@@ -0,0 +1,38 @@
+# Check that, by default, directories are automatically created for output files.
```
ddunbar: This comment is out of date.

tests/BuildSystem/Build/working-directory.llbuild
```diff
@@ -0,0 +1,38 @@
...
+    tool: shell
+    description: TOUCH
+    args: ["touch", "file"]
+    working-directory: "subdir/subsubdir"
+    outputs: ["subdir/subsubdir/file"]
+  C.symlink:
+    tool: symlink
+    description: SYMLINK
+    outputs: ["subdir-2/subsubdir/symlink"]
+    contents: ignored
```
ddunbar: This test feels like it could safely be simplified?

lib/Basic/Subprocess.cpp
dmbryson: Per the “docs” setting -1 clears the per thread working dir, restoring it to the process level.

tests/BuildSystem/Build/working-directory.llbuild
dmbryson: I’m sure.  I (clearly) just copied and tweaked an existing test.  I do like that it covers changing between working directories, but I didn’t evaluate it finely.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

lib/Basic/Subprocess.cpp
```diff
@@ -436,17 +457,44 @@ void llbuild::basic::spawnProcess(
 
     // If we have been cancelled since we started, do nothing.
     if (!wasCancelled) {
-      int result = posix_spawn(&pid, args[0], /*file_actions=*/&fileActions,
-                               /*attrp=*/&attributes, const_cast<char**>(args.data()),
-                               const_cast<char* const*>(environment.getEnvp()));
+      int result = 0;
+
+#ifdef __APPLE__
+      thread_local std::string threadWorkingDir;
```
ddunbar: I'm not sure we should use thread_local, would it be better to store this in the LaneBasedExecutionQueueJobContext?

dmbryson: We could store it in the ProcessContext, but right now that is an opaque type that is merely passed through the process code and back to the queue.  In the lane queue, that context is created per task, thus doesn't stay resident for the lifetime of the thread.   Both can be changed, of course, but it makes using 'spawnProcess' a bit more fragile.   I think the overhead of the thread_local is minimal in this case, given that it is function scoped.

ddunbar: Ok, works for me!

include/llbuild/Basic/Subprocess.h
```diff
@@ -224,6 +224,19 @@ namespace llbuild {
...
+
+    struct ProcessAttributes {
+      /// If true, whether it is safe to attempt to SIGINT the process to cancel
+      /// it. If false, the process won't be interrupted during cancellation and
+      /// will be given a chance to complete (if it fails to complete it will
+      /// ultimately be sent a SIGKILL).
+      bool canSafelyInterrupt;
+
+      /// If set, the working directory to change into before spawning (only
+      /// supported on macOS)
```
jakepetroules: Has anyone investigated whether other platforms (Windows, Linux) have any such API?

At least, [CreateProcess](https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessw) on Win32 appears to accept a working directory parameter (`lpCurrentDirectory`) which may do what we need. In `lib/llvm/Support/Windows/Program.inc` we're just passing NULL right now, which uses the calling process' working directory.

ddunbar: Linux doesn't, as far as my research has shown. We can either fork, or we can use a "zygote" approach.

We don't really support Windows, so that is kind of a tangent.

PR #396, opened 2018-11-14
Pass NULL-terminated buffer to pthread_chdir_np

jakepetroules:


dmbryson:
Works for me, though shouldn’t be necessary in C++11.  data() and c_str() are aliases.

dmbryson:
Ah, but I suppose it is possible our StringRef here may actually point to a non-std::string object.

ddunbar:
Yes, StringRef isn’t guaranteed to be null terminated.

PR #398, opened 2018-11-27
[BuildSystem] Allow control channel to be disabled

dmbryson:
Allows 'shell' tool commands to optionally disable the export of the
control file descriptor and accompanying environment variable.

rdar://problem/46152060

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -1949,6 +1949,9 @@ class ShellCommand : public ExternalCommand {
   /// Working directory in which to spawn the external command
   std::string workingDirectory;
 
+  /// Whether the control pipe is enabled for this command
+  bool controlEnabled = true;
+
   /// The cached signature, once computed -- 0 is used as a sentinel value.
   std::atomic<CommandSignature> cachedSignature{ };
   
@@ -2180,6 +2183,13 @@ public:
       ctx.error("working-directory unsupported on this platform");
       return false;
 #endif
+    } else if (name == "control-enabled") {
+      if (value != "true" && value != "false") {
+        ctx.error("invalid value: '" + value + "' for attribute '" +
+                  name + "'");
+        return false;
+      }
+      controlEnabled = value == "true";
     } else {
       return ExternalCommand::configureAttribute(ctx, name, value);
     }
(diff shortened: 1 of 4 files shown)


dmbryson:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2180,6 +2183,13 @@ class ShellCommand : public ExternalCommand {
       ctx.error("working-directory unsupported on this platform");
       return false;
 #endif
+    } else if (name == "control-enabled") {
+      if (value != "true" && value != "false") {
+        ctx.error("invalid value: '" + value + "' for attribute '" +
+                  name + "'");
+        return false;
+      }
+      controlEnabled = value == "true";
```
ddunbar: Is there a good reason to make this per command? I am concerned about the size impact on the manifest file. Could this be a global setting instead, at least for now while we are still experimenting with it?

dmbryson: Hmm, my thought was that a client could selectively disable the control channel for particular commands that are non-conforming.  That said, it was also a minimal resistance path to passing the argument in.  I can definitely see adding a client level default value for it as well, though it will be a little more work to get that value into the externalCommand objects where it is needed.

jakepetroules: @ddunbar If we do this per-command, some sample stats would be: between 2000-2500 occurrences of this property in a manifest file for a project with around 15k source code files.

dmbryson: Hmm, trying to set this at the client level has some annoying plumbing.  But, perhaps a 'tool' level default makes sense.  So the "shell" tool could support this attribute as well and set the default value for all such commands.

jakepetroules: 👍 tool level

ddunbar: Tool level sounds perfect

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2282,14 +2284,26 @@ class ShellCommand : public ExternalCommand {
 };
 
 class ShellTool : public Tool {
+private:
+  bool controlEnabled = true;
```
ddunbar: Should the default be false?

ddunbar: Okay, sounds good.

PR #399, opened 2018-11-28
[llbuildSwift] Expose error() delegate callback in Core bindings

dmbryson:
rdar://problem/45728573

### DIFF
--- a/products/llbuildSwift/CoreBindings.swift
+++ b/products/llbuildSwift/CoreBindings.swift
@@ -237,6 +237,17 @@ public protocol BuildEngineDelegate {
     /// requested Key cannot be supplied, the delegate should provide a dummy rule
     /// that the client can translate into an error.
     func lookupRule(_ key: Key) -> Rule
+
+    /// Handle error message reported by the engine
+    ///
+    /// The delegate should handle this, but a default implementation is
+    /// is provided that does nothing for backward compatiblity.
+    func error(_ message: String)
+}
+
+public extension BuildEngineDelegate {
+    func error(_ message: String) {
+    }
 }
 
 /// Wrapper to allow passing an opaque pointer to a protocol type.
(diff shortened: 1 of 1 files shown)


dmbryson:
@swift-ci please smoke test

products/llbuildSwift/CoreBindings.swift
```diff
@@ -237,6 +237,18 @@ public protocol BuildEngineDelegate {
...
+
+    /// Handle error message reported by the engine
+    ///
+    /// The delegate should handle this, but a default implementation is
+    /// is provided that does nothing for backward compatiblity.
+    func error(_ message: String)
+}
+
+extension BuildEngineDelegate {
+    func error(_ message: String) {
```
aciidb0mb3r: Shouldn't this be public?

products/llbuildSwift/CoreBindings.swift
```diff
@@ -237,6 +237,18 @@ public protocol BuildEngineDelegate {
...
+        return
```
aciidb0mb3r: nit: not required

products/llbuildSwift/CoreBindings.swift
dmbryson: Probably.  Smoke tests seem to indicate so.  Wonder why my local tests passed? 🤔

products/llbuildSwift/CoreBindings.swift
dmbryson: Is it not?  I don't want to break SwiftPM downstream by merging this with a newly required method.

aciidb0mb3r: Sorry, I meant the `return` keyword is not required 😅

dmbryson: Ahh, that makes sense.

dmbryson:
@swift-ci please smoke test

PR #400, opened 2018-12-01
Add support for setting the working directory of tasks on Linux

jakepetroules:
[Glibc 2.29](https://sourceware.org/git/?p=glibc.git;a=commit;h=4a938cb273e164a475dc123cc80ea6354d7248d4) will add posix_spawn_file_actions_addchdir_np. A method with
the same signature also exists on [Solaris](https://docs.oracle.com/cd/E86824_01/html/E54766/posix-spawn-file-actions-addchdir-np-3c.html), and there is currently a
[proposal](http://austingroupbugs.net/view.php?id=1208) to add a "posix_spawn_file_actions_addchdir" to POSIX, so
proactively re-structure the code to prefer using a
posix_spawn_file_actions_addchdir[_np] style function to set the working
directory.

We fall back to an alternate mechanism where available, or continue to
assert as before.

rdar://46069443

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -2205,17 +2205,12 @@ public:
       }
       inheritEnv = value == "true";
     } else if (name == "working-directory") {
-#if __APPLE__
       // Ensure the working directory is absolute. This will make sure any
       // relative directories are interpreted as relative to the CWD at the time
       // the rule is defined.
       SmallString<PATH_MAX> wd = value;
       llvm::sys::fs::make_absolute(wd);
       workingDirectory = StringRef(wd);
-#else
-      ctx.error("working-directory unsupported on this platform");
-      return false;
-#endif
     } else if (name == "control-enabled") {
       if (value != "true" && value != "false") {
         ctx.error("invalid value: '" + value + "' for attribute '" +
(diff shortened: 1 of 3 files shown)


jakepetroules:
@swift-ci please smoke test

jakepetroules:
Don't think `unshare` will work as a fallback on Linux because there's no equivalent to `pthread_fchdir_np(-1)` to revert back to the process-global working directory. So Glibc 2.29 is a hard requirement on Linux and we'll have to continue skipping the test for Linux in Swift CI until CI has Glibc 2.29.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2205,17 +2205,12 @@ class ShellCommand : public ExternalCommand {
       }
       inheritEnv = value == "true";
     } else if (name == "working-directory") {
-#if __APPLE__
```
dmbryson: Hmm, I wonder if it might be worth defining something like LLBUILD_SUBPROCESS_WORKING_DIRECTORY_SUPPORTED in Subprocess.h and using that here.  I think it is nicer for a UX perspective to let the user know it is unsupported as early as possible. It is conceivable that a client could have a manifest/build that does a substantial amount of work before the error in the actual spawn method is triggered.

jakepetroules: This is non-trivial and not something we expect that users would actually see outside of a tool development environment.

dmbryson: Fair enough.

PR #402, opened 2018-12-05
[Basic] Add cap for background tasks in LaneBasedExecutionQueue

dmbryson:
rdar://problem/46491559


dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

lib/Basic/LaneBasedExecutionQueue.cpp
```diff
@@ -100,6 +100,12 @@ class LaneBasedExecutionQueue : public ExecutionQueue {
   std::mutex queueCompleteMutex;
   bool queueComplete { false };
 
+  /// Background (lane released) task management
+  std::mutex backgroundTaskMutex;
+  unsigned backgroundTaskMax = 0;
+  unsigned backgroundTaskCount = 0;
```
ddunbar: Can we just use atomics for these?

dmbryson: Hmm, maybe.  The phrasing around the check site would get a little more awkward if we were to try to keep the hard cap semantics while using a compare_exchange.  Could be a soft cap, though.

dmbryson:
@swift-ci please smoke test

lib/Basic/LaneBasedExecutionQueue.cpp
dmbryson: Updated to the soft cap approach (with comment explaining).

dmbryson:
@swift-ci please smoke test

dmbryson:

lib/Basic/LaneBasedExecutionQueue.cpp
```diff
@@ -190,6 +196,7 @@ class LaneBasedExecutionQueue : public ExecutionQueue {
   : ExecutionQueue(delegate), buildID(std::random_device()()), numLanes(numLanes),
         readyJobs(Scheduler::make(alg)), environment(environment)
   {
+    backgroundTaskMax = numLanes * 16;
```
hpux735: I have a question about this.  I'm working on [SR-10325](https://bugs.swift.org/browse/SR-10325?focusedCommentId=45543&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-45543) and I'm trying to understand where the total number of background tasks come from.  In particular, it seems like this is allowing 16 tasks _per lane_.  Is that correct?  Elsewhere in SPM, the number of lanes are set by the number of CPUs.  On an my 8-core system (Big+little, 4 each) SPM spawns enough processes to essentially DoS the system.

dmbryson: This is indeed allowing 16 background tasks per lane. Note, however, that as far as I am aware, there are no native SwiftPM tasks that participate in the background task support of llbuild at this time. Build tsks that wish to use this feature must explicitly opt in to it by sending the control message to the build system and are expected to behave nicely (not use excessive CPU) once they have done so.

hpux735: Thanks!  I believe you're correct.  I set this to `numLanes * 1` and it behaved the same way.  It turns out that I need to control the number of lanes rather than the backgroundTaskMax.

futurejones:
@dmbryson would it be possible to get a copy of rdar://problem/46491559 so we can better understand the reason for these changes?
Thanks.

dmbryson:
> Add a cap for released lanes to prevent system overloading

As noted above, this cap is only relevant for build tasks that make use of the explicit lane release feature.

dmbryson:
For a little more context that the radar does not explicitly spell out (given my brevity in writing it 😉), the cap exists primarily because we burn a whole thread for each background task at this time. We'd prefer to have a lighter weight IO handler so that we can keep our thread overhead low.  But we don't have such a framework in llbuild at this time.

futurejones:
Ok, thanks.

PR #404, opened 2018-12-06
Use Clang attributes to import llbuild enumerations more nicely in Swift

jakepetroules:
This obviates the need to duplicate every llbuild enum in the Swift
bindings, where we may forget to update the mapping. It also provides
the Swift compiler with a real enumeration instead of a series of
integers, which allows it to emit diagnostics in cases where a switch
does not cover all enum cases, for example.

rdar://problem/46476230

### DIFF
--- a/products/libllbuild/include/llbuild/buildsystem.h
+++ b/products/libllbuild/include/llbuild/buildsystem.h
@@ -72,11 +72,11 @@ typedef struct llb_fs_file_info_t_ {
 /// @name Diagnostics
 /// @{
 
-typedef enum {
-  llb_buildsystem_diagnostic_kind_note,
-  llb_buildsystem_diagnostic_kind_warning,
-  llb_buildsystem_diagnostic_kind_error,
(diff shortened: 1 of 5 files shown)


jakepetroules:
@swift-ci please smoke test

aciidb0mb3r:
nice!

jakepetroules:
Missed a couple, updated again.

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please smoke test

neonichu:
@swift-ci please smoke test

jakepetroules:
So this is failing on Linux because we don't have a Clang there which supports the `swift_name` attribute. Apparently we're relying on the system Clang instead of shipping our own fork.

aciidb0mb3r:
https://forums.swift.org/t/shipping-clang-with-the-swift-toolchains/18446/1

jakepetroules:
@swift-ci please smoke test

PR #405, opened 2018-12-17
Make a couple of our enumerations string-convertible for diagnostics

jakepetroules:


jakepetroules:
@swift-ci please smoke test

PR #406, opened 2018-12-19
Raise open file limit before building

dmbryson:
- Adds raiseOpenFileLimit to PlatformUtility
- Raises open file limit before constructing execution queue in
BuildSystem
- Raises open file limit at the start of Ninja builds

rdar://problem/46823730


dmbryson:
@swift-ci please smoke test

PR #407, opened 2018-12-19
[PerfTests] Fix another compiler warning

dmbryson:


dmbryson:
@swift-ci please smoke test

PR #408, opened 2018-12-19
[Xcode] Set 'find implicit dependencies'

dmbryson:
rdar://problem/46706652

### DIFF
--- a/llbuild.xcodeproj/project.pbxproj
+++ b/llbuild.xcodeproj/project.pbxproj
@@ -243,6 +243,13 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		4035804021C3060F0091FE09 /* PBXContainerItemProxy */ = {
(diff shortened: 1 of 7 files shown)


dmbryson:
@swift-ci please smoke test

ddunbar:
Ugh, I hate implicit dependencies, but ok if you want it.

dmbryson:
Well, I don't *want* it per se, but because of the way we link via search paths, linking against a static library is insufficient to establish the build ordering dependency.  I have also added the explicit target dependency for the build failure observed (NinjaTests requiring llbuildNinja), and could reduce this change to just that.  It was through discussion with Michael that I gathered the impression that "find implicit dependencies" was the strongly encouraged way to do it.

I'd prefer that the project model and build system actually modeled the action performed in the UI where dragging a static library in to a link phase made that an explicit input dependency (
rdar://problem/46850444).

PR #410, opened 2019-01-10
[BuildSystem] Handle relative paths in 'shell' makefile-style dependencies

dmbryson:
Under some situations clang will emit relative paths for included
dependencies. This changes 'shell' commands that use makefile-style
deps files to attempt to make dependecy paths absolute before generating
the node key.

rdar://problem/44463222

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -52,6 +52,7 @@
 #ifdef _WIN32
 #include <Shlwapi.h>
 #else
+#include <limits.h>
 #include <fnmatch.h>
 #include <unistd.h>
 #endif
@@ -2074,7 +2075,21 @@ class ShellCommand : public ExternalCommand {
       virtual void actOnRuleDependency(const char* dependency,
                                        uint64_t length,
                                        const StringRef unescapedWord) override {
-        bsci.taskDiscoveredDependency(task, BuildKey::makeNode(unescapedWord));
+        if (llvm::sys::path::is_absolute(unescapedWord)) {
+          bsci.taskDiscoveredDependency(task, BuildKey::makeNode(unescapedWord));
+          return;
+        }
+
+        // Generate absolute path
+        //
+        // NOTE: This is making the assumption that relative paths coming in a
+        // dependency file are in relation to the explictly set working
+        // directory, or the current working directory when it has not been set.
+        SmallString<PATH_MAX> absPath = StringRef(command->workingDirectory);
+        llvm::sys::path::append(absPath, unescapedWord);
+        llvm::sys::fs::make_absolute(absPath);
+
+        bsci.taskDiscoveredDependency(task, BuildKey::makeNode(absPath));
       }
 
       virtual void actOnRuleStart(const char* name, uint64_t length,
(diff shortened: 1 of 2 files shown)


dmbryson:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2074,7 +2075,24 @@ class ShellCommand : public ExternalCommand {
...
+          bsci.taskDiscoveredDependency(task, BuildKey::makeNode(unescapedWord));
+          return;
+        }
+
+        // Generate absolute path
+        //
+        // NOTE: This is making the assumption that relative paths coming in a
+        // dependency file are in relation to the explictly set working
+        // directory, or the current working directory when it has not been set.
+        SmallString<PATH_MAX> absPath = StringRef(command->workingDirectory);
```
dmbryson: Hmm... this isn't quite absolute just yet.  It is assuming that 'workingDirectory' is absolute, which may not be the case.

ddunbar: Are we sure this isn't a clang bug? If clang generates relative paths, and something uses -fworking-directory, could we ever handle that case correctly?

ddunbar: In any case, I agree we should handle this case, just wondering if there is a corner case we can't fix

dmbryson:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
ddunbar: Should we just make sure that every command's working directory is absolute?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2074,7 +2075,24 @@ class ShellCommand : public ExternalCommand {
...
+        if (!absPath.empty()) {
+          absPath.append("/");
+        }
+        absPath.append(unescapedWord);
+        llvm::sys::fs::make_absolute(absPath);
```
ddunbar: Doesn't llvm::sys::fs have support for appending a path, versus the manual code above?

tests/BuildSystem/Build/discovered-makefile-deps-relative.llbuild
```diff
@@ -0,0 +1,58 @@
...
+# Check a build that modifies the header.
+#
+# RUN: echo "mod" >> %t.build/header-1
+# RUN: %{llbuild} buildsystem build --serial --chdir %t.build &> %t2.out
+# RUN: %{FileCheck} --check-prefix=CHECK-AFTER-MOD --input-file=%t2.out %s
+#
+# CHECK-AFTER-MOD: CC output-1
+# CHECK-AFTER-MOD: cat output-1 > ../output
+#
+# REQUIRES: platform=Darwin
```
ddunbar: Why does this require Darwin?

dmbryson: Its using the working-directory support, which is not yet implemented on linux (requires #400 to be finished).

lib/BuildSystem/BuildSystem.cpp
dmbryson: Probably not a bad idea.

lib/BuildSystem/BuildSystem.cpp
dmbryson: It does indeed.  I looked for it before, but I was looking for "join" rather than "append" and overlooked it.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
jakepetroules: Agreed, I don't think it makes any sense to allow relative working directories given that the manifest is generally intended to be entirely non-relocatable.

tests/BuildSystem/Build/discovered-makefile-deps-relative.llbuild
jakepetroules: Ah yes, I should deal with that...

PR #412, opened 2019-01-11
Update manifest version to 5.0

aciidb0mb3r:
This allows us to use the latest build settings API for linking the
required libraries.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,13 +1,7 @@
-// swift-tools-version:4.0
+// swift-tools-version:5.0
 
 // This file defines Swift package manager support for llbuild. See:
 //  https://github.com/apple/swift-package-manager/tree/master/Documentation
-//
-// You can build using:
-//
-// ```shell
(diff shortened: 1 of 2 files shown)


ddunbar:
My only concern here is that this will break things for older users. Should we add a compatibility manifest for them, and leave the entries in the module map for now?

aciidb0mb3r:
I was thinking about that but I don't know if it's worth it.

jakepetroules:
Why are you not using `systemLibrary` targets for this?

aciidb0mb3r:
for what?

jakepetroules:
For linking sqlite and curses.

aciidb0mb3r:
system library targets are required when you want to expose a system library to Swift but in this case we just need to link them.

jakepetroules:
@aciidb0mb3r You can merge this now.

dmbryson:
Concerns of breaking things were founded.  Some downstream projects have experienced build failures.  Going to revert for now.

PR #413, opened 2019-01-12
Reimplement Hashable with hash(into:)

hartbit:
Resolves a FIX to use a real hash function.


hartbit:
@swift-ci please smoke test

PR #414, opened 2019-01-12
Fix Swift warning about future unknown enum values

hartbit:


hartbit:
@swift-ci please smoke test

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -158,6 +160,8 @@ extension BuildKeyKind: CustomStringConvertible {
             return "filteredDirectoryContents"
         case .stat:
             return "stat"
+        @unknown default:
+            return "unknown"
```
jakepetroules: This is ambiguous with the one above, can you please use something else, perhaps `"BuildKeyKind(\(rawValue))"`?

dmbryson:
LGTM after the changes suggested by Jake

products/llbuildSwift/BuildSystemBindings.swift
hartbit: Not sure what you mean by `"BuildKeyKind(\(rawValue))"`. I can return something like `"<unkown>"` if you want.

hartbit: Done 😄 Thanks for the review!

dmbryson:
@swift-ci please smoke test

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@dmbryson Will this break anything if merged, RE the revert done in #419? As in, do we need to wait until we revert that revert, to merge this, or can we merge it now?

PR #416, opened 2019-01-18
Implemented process spawning for windows

jmittert:
Use CreateProcessW to spawn processes on Windows. This enables simple swift-build files to work on Windows.

```
S:\build\Ninja-DebugAssert\llbuild-windows-amd64\tests\SwiftBuildTool\Output\basic.swift-build.tmp.build> cat .\build.swift-build                                         
# Check that the tool works.                                                                                                                                              
#                                                                                                                                                                         
# RUN: rm -rf %t.build                                                                                                                                                    
# RUN: mkdir -p %t.build                                                                                                                                                  
# RUN: cp %s %t.build/build.swift-build                                                                                                                                   
# RUN: %{swift-build-tool} --chdir %t.build > %t.out                                                                                                                      
# RUN: %{FileCheck} < %t.out %s                                                                                                                                           
#                                                                                                                                                                         
# CHECK: ECHO                                                                                                                                                             
# CHECK: Hello Swift Build Tool                                                                                                                                           
                                                                                                                                                                          
client:                                                                                                                                                                   
  name: swift-build                                                                                                                                                       
                                                                                                                                                                          
targets:                                                                                                                                                                  
  "": ["<echo>"]                                                                                                                                                          
                                                                                                                                                                          
commands:                                                                                                                                                                 
  C1:                                                                                                                                                                     
    tool: shell                                                                                                                                                           
    outputs: ["<echo>"]                                                                                                                                                   
    description: ECHO                                                                                                                                                     
    args: echo "Hello Swift Build Tool"                                                                                                                                   
S:\build\Ninja-DebugAssert\llbuild-windows-amd64\tests\SwiftBuildTool\Output\basic.swift-build.tmp.build> &"S:\\build\\Ninja-DebugAssert\\llbuild-windows-amd64\\bin\\swift-build-tool" --no-db                                                                                                                                                     
ECHO                                                                                                                                                                      
"Hello Swift Build Tool"                                                                                                                                                  
S:\build\Ninja-DebugAssert\llbuild-windows-amd64\tests\SwiftBuildTool\Output\basic.swift-build.tmp.build>
```

I don't particularly like putting the `#if defined(_WIN32)`s everywhere, but I'm not really sure how to set up for `CreateProcessW/posix_spawn` in a cleaner way. Any advice on that would be appreciated, most of it doesn't seem like things that could be lifted out and moved to PlatformUtility.

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -12,6 +12,7 @@
 
 #include "llbuild/Basic/Subprocess.h"
 
+#include "llbuild/Basic/CrossPlatformCompatibility.h"
 #include "llbuild/Basic/PlatformUtility.h"
 
 #include "llvm/ADT/ArrayRef.h"
@@ -78,6 +79,7 @@ int pthread_fchdir_np(int fd)
 #endif
 #endif
 
+#if !defined(_WIN32)
 static int posix_spawn_file_actions_addchdir(posix_spawn_file_actions_t * __restrict file_actions,
                                              const char * __restrict path) {
 #if HAVE_POSIX_SPAWN_CHDIR
@@ -88,6 +90,7 @@ static int posix_spawn_file_actions_addchdir(posix_spawn_file_actions_t * __rest
   return ENOSYS;
 #endif
 }
+#endif
 
 using namespace llbuild;
 using namespace llbuild::basic;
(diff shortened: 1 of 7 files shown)


lib/Basic/Subprocess.cpp
```diff
@@ -421,23 +409,35 @@ void llbuild::basic::spawnProcess(
 
   // Form the complete C string command line.
   std::vector<std::string> argsStorage(commandLine.begin(), commandLine.end());
+#if defined(_WIN32)
+  std::string args;
+  std::for_each(std::begin(argsStorage) + 1, std::end(argsStorage),
+                [&args](auto arg) { args += arg + " "; });
```
jakepetroules: This is a really hard problem because the command line formatting rules on Windows are quite complex. You essentially need to do the inverse of `CommandLineToArgvW` here, which isn't something provided by the Windows API. Does LLVM have an implementation of converting a string array to a string suitable for passing to `CreateProcess`?

There's also a great MSDN article which details the process for how to do this: https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

jmittert: LLVM exposes a handy `std::string flattenWindowsCommandLine(ArrayRef<StringRef> Args);`, but only in a more [recent version (mid 2018)](https://github.com/llvm-mirror/llvm/commit/07d33bf7d7df1df00c40bdc859c17025d8457cf8) than what's in [llbuild](https://github.com/apple/swift-llbuild/commit/089f9922c6cdf68cee068593ba805fb4dceb55ea). What's the process like for updating the included llvm support libraries?

jakepetroules: @ddunbar, can you advise?

jakepetroules: @ddunbar We should copy the code from the apple/swift-llvm repo, right?

ddunbar: Yes. See `lib/llvm/import-llvm` for a little more info

ddunbar: I would suggest decoupling that from this particular PR though, we can just start with a not 100% functional version and improve it when the LLVM code lands

jakepetroules:
@swift-ci please smoke test

PR #418, opened 2019-01-23
[tests] Add ninja test that outputs to nonexistent subdir

ePirat:
This is an attempt to add a test for [SR-9735](https://bugs.swift.org/browse/SR-9735), until the bug is fixed it is supposed to fail with llbuild but succeed with the "real" Ninja.

I could not figure out how to actually run these tests so I just tried to deduce how to write these tests by looking at the existing tests.

### DIFF
--- a/tests/Ninja/Build/nonexistent-subdir.ninja
+++ b/tests/Ninja/Build/nonexistent-subdir.ninja
@@ -1,0 +1,20 @@
+# Check that outputs to not yet existing directories works fine
+
+# We run the build in a sandbox in the temp directory to ensure we don't
+# interact with the source dirs.
+
+# This test currently fails even though this works fine with
+# upstream Ninja, see SR-9735
+
+# XFAIL: *
+# RUN: rm -rf %t.build
+# RUN: mkdir -p %t.build
+# RUN: cp %s %t.build/build.ninja
+# RUN: touch %t.build/input
+# RUN: %{llbuild} ninja build --jobs 1 --no-db --chdir %t.build &> %t.out
+# RUN: %{FileCheck} < %t.out %s
+
+rule CAT
+     command = cat ${in} > ${out}
+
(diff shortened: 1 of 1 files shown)


dmbryson:
You should be able to run `make test`, though it does require the llvm 'FileCheck' utility.  On Ubuntu you can install it with the 'llvm-X.X-tools' package (where X.X is an appropriate version number).  On macOS you will likely have to build it from llvm.  @ddunbar is there an easier way to get FileCheck for macOS?

dmbryson:
@swift-ci please smoke test

ddunbar:
Not a good one, no

dmbryson:
Given this is supposed to fail, we probably need an XFAIL in there.

ePirat:
Well my idea was to merge this once SR-9735 is fixed so that it succeeds, not sure if it makes sense to test that something is broken, as the test would fail once the broken behavior is fixed.

Unless there are no plans to fix SR-9735, though that would be quite a big behavior difference compared to upstream ninja.

dmbryson:
Its worth having a test in place, even if we know it currently fails.  Will allow it to be clear when a fix is committed that it addresses the problem.  i.e. that PR would include the change that removes the XFAIL line.

dmbryson:
A comment adjacent to XFAIL explaining the problem and referencing the SR would also be good.

ePirat:
How do I do the XFAIL properly? The only usage I found is in `tests/BuildSystem/Build/mkdir.llbuild` and that refers to a specific configuration (`windows_subsystem_linux`).
Can I just add a `# XFAIL` without the `: something` part?

ePirat:
Updated with XFAIL and a comment.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #421, opened 2019-01-26
Make it a little easier to find FileCheck via Homebrew

jakepetroules:
...by falling back to adding the default Homebrew install path as a
search path. Xcode won't see /usr/local/bin when added to PATH via
.profile.


jakepetroules:
@swift-ci please smoke test

PR #422, opened 2019-01-30
Fixed the Unit Tests on Windows

jmittert:
This fixes all the unittests on Windows!

```
S:\llbuild> C:\Python27\python.exe C:/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/llvm-lit.py --param build_mode=Debug S:/build/Ninja-DebugAssert/llbuild-windows-amd64/tests --filter "unit" -sv                                                      
-- Testing: 64 of 187 tests, 12 threads --                                                                                                  
Testing: 0 .. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90..                                                                                  
Testing Time: 22.26s                                                                                                                        
  Expected Passes    : 64  
```


lib/Basic/PlatformUtility.cpp
```diff
@@ -42,11 +44,68 @@ int sys::close(int fileHandle) {
...
+  return (time_t)((ltime - 116444736000000000) / 10000000);
+}
+#endif
+
 int sys::lstat(const char *fileName, sys::StatStruct *buf) {
 #if defined(_WIN32)
-  // We deliberately ignore lstat on Windows, and delegate
-  // to stat.
-  return ::_stat(fileName, buf);
+  llvm::SmallVector<UTF16, 20> wfilename;
```
jmittert: This is definitely not a _good_ implementation of lstat, but Windows doesn't provide it and it's good enough for the purposes of checking if file is a directory/regular file/symlink so llbuild knows to recurse into it or not.

unittests/Basic/FileSystemTest.cpp
```diff
@@ -40,7 +40,7 @@ TEST(FileSystemTest, basic) {
     std::error_code ec;
     llvm::raw_fd_ostream os(tempPath.str(), ec, llvm::sys::fs::F_Text);
     EXPECT_FALSE(ec);
-    os << "Hello, world!\n";
+    os << "Hello, world!";
```
jmittert: This was causing problems with `\r\n`. Rather than `ifdef`ing, I figured it should be fine to just not check the newlines

unittests/Basic/FileSystemTest.cpp
```diff
@@ -54,15 +54,15 @@ TEST(FileSystemTest, basic) {
...
+  EXPECT_EQ(ourFileContents->getBuffer().str(), "Hello, world!");
 
   // Remote the temporary file.
   auto ec = llvm::sys::fs::remove(tempPath.str());
   EXPECT_FALSE(ec);
 }
 
 TEST(FileSystemTest, testRecursiveRemoval) {
-  TmpDir rootTempDir{ __FUNCTION__ };
+  TmpDir rootTempDir(__func__);
```
jmittert: `__FUNCTION__` in MSVC doesn't give the same thing as `__FUNCTION__` in Clang. `__func__` however is standardized in C++11 and is the same thing as `__FUNCTION__` with Clang

unittests/BuildSystem/BuildSystemFrontendTest.cpp
```diff
@@ -343,6 +345,20 @@ TEST_F(BuildSystemFrontendTest, commandSkippingFailure) {
...
+#if defined(_WIN32)
+    ASSERT_TRUE(delegate.checkTrace(R"END(
+commandPreparing: 2
+commandPreparing: 1
+shouldCommandStart: 1
+commandFinished: 1: 3
+shouldCommandStart: 2
+commandStarted: 2
+commandProcessStarted: 2
+commandProcessFinished: 2: 1
```
jmittert: Should be the same except that Windows gives a different error code

unittests/BuildSystem/BuildSystemFrontendTest.cpp
```diff
@@ -441,11 +457,16 @@ TEST_F(BuildSystemFrontendTest, commandSkipping_NonExternalCommands_symlink) {
         outputs: ["1"]
         contents: "x"
     2:
-        tool: symlink
+        tool: shell
```
jmittert: Windows seems to be unable to rm a symlink if the target doesn't exist though it doesn't have an issue creating one.

jmittert: Can you expand on what you mean by "handle that internally to the shell tool"?

jakepetroules: I know it's with the deletion, but isn't there a Win32 API we can call to delete a symlink without attempting to follow it?

jmittert: I don't follow... there are Win32 APIs to do exactly that, but gnu msys rm doesn't do it properly. It says it's using lstat, but whatever implementation of lstat it's using doesn't seem to be correct.  Unless you're alluding to providing a RemoveCommand that does this properly?

jakepetroules: Ohhh, sorry, I completely misread the diff due to the way git is presenting it here. Disregard all this :)

jmittert:
Alright, remove the extra formatting changes, changed all winapi calls to the W ones and addressed the remaining comments. I don't have access to a macOS machine right now, but the tests look good on the Vagrant box so... *crosses fingers*.

dmbryson:
@swift-ci please smoke test

jakepetroules:
> Alright, removed the extra formatting changes, changed all winapi calls to the W ones and addressed the remaining comments. I don't have access to a macOS machine right now, but the tests look good on the Vagrant box so... _crosses fingers_.

Our CI will validate the macOS build, no worries.

jakepetroules:
@swift-ci please smoke test

PR #423, opened 2019-01-31
[llbuildSwift] Replace hashValue with hash(into:)

aciidb0mb3r:
Fixes the deprecation warning and this change is source compatible with Swift 4.2


aciidb0mb3r:
@swift-ci smoke test

PR #425, opened 2019-02-02
Fix the Unit Tests on Windows [Fixed Mismatch]

jmittert:
- Fixing Symlink handling on Windows
- The windows process handling I did was broken in a few places
- Handling Windows paths
- Working around a few Clang/MSVC inconsistencies

Same thing as https://github.com/apple/swift-llbuild/pull/422 but avoiding the use of `mismatch(first1, last1, first2, last2)`


jakepetroules:
@swift-ci please smoke test

PR #426, opened 2019-02-05
Handling Memory Across the API Boundary

jmittert:
Freeing across modules [can cause issues in Windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=vs-2017) which llbuild currently does when [getting file contents](https://github.com/apple/swift-llbuild/blob/master/products/libllbuild/include/llbuild/buildsystem.h#L260) using a delegate. This adds a fs_free field which can be used to call the correct free in the library.

However, [there's a comment](https://github.com/apple/swift-llbuild/blob/master/products/libllbuild/include/llbuild/buildsystem.h#L272):   
```
// FIXME: Design clean data types for clients to return unowned/unowned
// memory.
```
which seems like a better idea than adding the fs_free field which would be highly prone to error. Are there existing ideas/proposals for this? I'd refer to "do it properly" and implement that instead. If I want to add something like that, how stable is the llbuild public api considered to be right now?

### DIFF
--- a/examples/c-api/buildsystem/main.c
+++ b/examples/c-api/buildsystem/main.c
@@ -22,6 +22,9 @@
 #include <string.h>
 
 #include <sys/stat.h>
+#if defined(_WIN32)
+#include <windows.h>
+#endif
 
 static const char* progname;
 static void usage() {
@@ -33,6 +36,27 @@ static const char* basename(const char* path) {
   const char* result = strrchr(path, '/');
   return result ? result : path;
 }
+
+#if defined(_WIN32)
+static wchar_t* convertToMultiByte(const char* s) {
+  int numChars = MultiByteToWideChar(
+      /*CodePage=*/CP_UTF8,
+      /*dwFlags=*/0,
+      /*lpMultibyteStr=*/s,
+      /*cbMultiByte=*/-1,
+      /*lpWideCharStr=*/NULL,
+      /*ccWideChar=*/0);
+  wchar_t* mbs = malloc(numChars * sizeof(wchar_t));
+  MultiByteToWideChar(
+      /*CodePage=*/CP_UTF8,
+      /*dwFlags=*/0,
+      /*lpMultibyteStr=*/s,
+      /*cbMultiByte=*/-1,
+      /*lpWideCharStr=*/mbs,
+      /*ccWideChar=*/numChars);
+  return mbs;
+}
+#endif
 
 // "Fancy" Command Implementation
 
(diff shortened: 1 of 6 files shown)


examples/c-api/buildsystem/main.c
```diff
@@ -64,11 +64,18 @@ static bool fs_get_file_contents(void* context, const char* path,
   printf(" -- read file contents: %s\n", path);
   fflush(stdout);
 
+#if defined(_WIN32)
+  FILE* fp;
+  if (fopen_s(&fp, path, "rb")) {
```
jakepetroules: `_wfopen_s`, please :)

Probably not a bad idea to move this into PlatformUtility...

unittests/CAPI/BuildSystem-C-API.cpp
```diff
@@ -136,12 +136,19 @@ static bool fs_get_file_contents(void* context, const char* path,
                                 llb_data_t* data_out) {
   printf(" -- read file contents: %s\n", path);
   fflush(stdout);
-  
+
+#if defined(_WIN32)
+  FILE* fp;
+  if (fopen_s(&fp, path, "rb")) {
```
jakepetroules: same here

jakepetroules:
> If I want to add something like that, how stable is the llbuild public api considered to be right now?

@ddunbar or @dmbryson are better equipped to answer this, but in general I don't think you need to be too worried about breaking clients. There aren't many.

ddunbar:
We try to keep the public API stable, but it is not a hard requirement, we just have to be careful about how we stage it in so that clients have a way to adapt. If there is a way we can do it by adding a new API, it might be worth it just to keep the API transition simple for clients.

ddunbar:
As far as this:

> I'd refer to "do it properly" and implement that instead. If I want to add something like that, how stable is the llbuild public api considered to be right now?

goes, I haven't thought in depth about it, but I would imagine we would augment the `llb_data_t` type to include information on how to release the data, or introduce a new, similar, type which knows how to release its data.

I would probably suggest we define API methods for actually doing the allocation and release (even if we don't need them yet).

However, I haven't honestly thought about this in depth... how does `libclang` solve this problem (I assume it works on Windows, but maybe not)? We often look to LLVM for precedent.

jmittert:
> However, I haven't honestly thought about this in depth... how does `libclang` solve this problem (I assume it works on Windows, but maybe not)? We often look to LLVM for precedent.

Looks like clang does something similar where it exposes a clang_free which just calls through to the right free. https://reviews.llvm.org/D10949

ddunbar:
@jmittert ok, well that's probably fine for us as well then, and avoids any API changes so maybe worth just copying that approach for data we allocate and expect the client to free

examples/c-api/buildsystem/main.c
jmittert: This one's in the client C sample code. I assume that it shouldn't be using the C++ functions in PlatformUtility? _wfopen_s makes sense though.

jakepetroules: Oh, right. I wasn't looking at the path. But yes, still should use _wfopen_s.

jmittert:
- Changed to to add `llb_alloc` and `llb_free` to allocate and free build system memory respectively
- Changed the example to use these new functions
- Use _w/s functions

examples/c-api/buildsystem/main.c
```diff
@@ -61,18 +85,29 @@ fancy_tool_create_command(void *context, const llb_data_t* name) {
...
 static bool fs_get_file_contents(void* context, const char* path,
                                  llb_data_t* data_out) {
+
+#if defined(_WIN32)
+  wchar_t* wPath = convertToMultiByte(path);
+  wprintf(L" -- read file contents: %ls\n", wPath);
+  fflush(stdout);
+  FILE* fp;
+  if (_wfopen_s(&fp, wPath, L"rb")) {
+    return false;
```
jakepetroules: wPath leaks if _wfopen_s fails

jmittert: Ayiii.. fixed

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please smoke test

PR #427, opened 2019-02-06
Rename llbuild.dll to libllbuild.dll on Windows

jmittert:
Cl by default produces .pdb and .ilk files by simply substituting the
extension of the desired target. This produces conflicts between the
debug files for llbuild.exe and llbuild.dll. This changes llbuild.dll to
be libllbuild.dll to avoid the problem.

### DIFF
--- a/products/libllbuild/CMakeLists.txt
+++ b/products/libllbuild/CMakeLists.txt
@@ -9,6 +9,12 @@ set(DEPENDENCIES
   llvmSupport
   sqlite3)
 
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+set(LIBLLBUILD_NAME libllbuild)
+else()
+set(LIBLLBUILD_NAME llbuild)
+endif()
+
 if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
   list(APPEND DEPENDENCIES curses)
 endif()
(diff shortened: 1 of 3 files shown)


jmittert:
> It's against Windows convention to prefix DLLs with "lib", so I don't really like this.
> 
> On the other hand, I'm not sure what else to do, so I guess we'll have to accept this for now.

Another options was rename it to `libllbuild/llbuild.dll` which would also solve the issue. 
Then in the build tree you'd have `bin/libllbuild/llbuild.dll` and `lib/libllbuild/llbuild.lib`

jakepetroules:
> > It's against Windows convention to prefix DLLs with "lib", so I don't really like this.
> > On the other hand, I'm not sure what else to do, so I guess we'll have to accept this for now.
> 
> Another options would be to rename it to `libllbuild/llbuild.dll` which would also solve the issue.
> Then in the build tree you'd have `bin/libllbuild/llbuild.dll` and `lib/libllbuild/llbuild.lib`

No, you want them in the same directory so that the normal DLL search path will pick them up automatically without having to edit `PATH`.

jakepetroules:
@swift-ci please smoke test

compnerd:
@jakepetroules - yeah, I don't think that this is good either.  This can be problematic when dealing with the raw tools (e.g. MSBuild, VS) and using this as the conventions are pretty commonly adhered to.

PR #428, opened 2019-02-07
Update LLVM to the latest stable

jmittert:
I've pulled in the latest version of llvm using llvm-swift's stable branch (~`f63b283c7143aef31863d5915c28ee79ed390be3`~ `1fb4269ad27c1cb3da13a5aa989601b9239663d3`). What I've essentially done is delete the llvm include and source directories, run the import script, and kept adding things to the script until all the dependencies were pulled, then fixed the config headers.

The files in the llvm directories that are not just straight copies are:
- [`include/llvm/Config/config.h`](https://github.com/apple/swift-llbuild/pull/428/files#diff-2675b3ae9211823220c66e10862f930c)
- [`include/llvm/Config/llvm-config.h`](https://github.com/apple/swift-llbuild/pull/428/files#diff-8d0743191f6a46765e222facae7573f2)
- [`lib/llvm/import-llvm`](https://github.com/apple/swift-llbuild/pull/428/files#diff-94e193fffd9eb7f5b7ef27ec1dd65676)

as well as a number of files in llbuild itself that needed updating for new llvm apis.

Given that llbuild would now "built against" a more recent llvm, what are the thoughts about making llbuild properly depends/build with llvm (a la LLVM_EXTERNAL_PROJECTS or similar) rather than copying over bits of it? Per the FAQ, I don't see llvm looking to factor out the Support libraries or llbuild re-implementing the support libraries in the near future.


ddunbar:
> Given that llbuild would now "built against" a more recent llvm, what are the thoughts about making llbuild properly depends/build with llvm (a la LLVM_EXTERNAL_PROJECTS or similar) rather than copying over bits of it? Per the FAQ, I don't see llvm looking to factor out the Support libraries or llbuild re-implementing the support libraries in the near future.

The problem is that we want llbuild to be very thin and easy to build, and LLVM is a beast. For example, llbuild can be built with the Swift package manager without much shenanigans. The import process is horrible, but if we wanted to do something here I would rather see if we can get LLVM to factor their libraries out, or move away from it entirely.

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@jmittert Looks like you might've forgotten to remove some deleted files from the Xcode project? Also, another C++ compatibility issue on the Linux side; didn't take a deeper look yet.

jakepetroules:
@swift-ci please smoke test

jmittert:
Fixed the xcodeproj, the swiftpm build, and the Linux build issue.

jmittert:
Ah, will that pick up the right commit? I pushed the wrong commit the first time.

jakepetroules:
IIRC it'll cancel a running test if you force push to the branch. I'll start it again.

jakepetroules:
@swift-ci please smoke test

jakepetroules:
> I've pulled in the latest version of llvm using llvm-swift's stable branch

Can you please update your PR to note the exact commit hash you used (helps for historical verification)?

jakepetroules:
@swift-ci please smoke test

aciidb0mb3r:
Note that we're currently auto-merging to swift-5.0-branch and we should be careful not to land such large changes right now.

jmittert:
> 
> 
> Note that we're currently auto-merging to swift-5.0-branch and we should be careful not to land such large changes right now.

Hrm, alright. Really, I only needed the command args parsing function, I just figured I'd pull the whole thing. If that's the case, then I think it would be reasonable to only update the cpp and header files. I believe it just adds a function (that's behind a _WIN32 if def), so it should be compatible and safer.

jakepetroules:
We do still want this change, we might just want to wait a bit before merging it. It'll take some time to properly review as it is (but having that commit hash will help too).

Once this is in, we should try to make an effort to update more regularly, so that the deltas are more manageable each time we do so and we don't have to contend with 50,000 line diffs :)

jmittert:
The commit hash I used from llvm-swift is `f63b283c7143aef31863d5915c28ee79ed390be3`

jakepetroules:
I think you're just missing the following from `Support_imports` in the import script:
- ConvertUTFWrapper
- Unix
- WindowsSupport

The changes to files in `include/llvm/Config` also look OK, although it would be nice to add those to the import script (possibly behind a flag) so that we can re-import those as well, and manually apply any diffs we need, rather than keeping them perpetually stale. Or, commit a diff file to the repo that applies necessary changes to the config files, and have the import script apply the diff after import.

jmittert:
Yeah, config files are generated to be platform specific by the llvm build and require manual changes to be used cross platform. This seems to be [what happened last import](https://github.com/apple/swift-llbuild/commit/07bc2cfe72911fcd62a1cfd13bc7da356c007ed0). I don't think abi-breaking.h needed changes though, that could be added to be just copied. I suppose a diff file would work for the over config files, but I'd suspect it would be fairly prone to breaking between imports as the config headers change.

I'll look into why those imports aren't being copied over.

jmittert:
- added the missing imports
- moved the script into utils
- the script now rms the llvm directories for you, recreates them, and additionally copies on the modified configs, creates the symlinks, and the CMakeLists.txts

I now get no diff after running the script.

I think that commit hash was from the wrong machine. `1fb4269ad27c1cb3da13a5aa989601b9239663d3` is latest stable and should be the hash.

jakepetroules:
> * the script now rms the llvm directories for you, recreates them, and additionally copies on the modified configs, creates the symlinks, and the CMakeLists.txts

Nice!

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please smoke test

PR #429, opened 2019-02-08
Escape strings for JSON when writing output-file-map.json

lanza:
Json interprets the backslashes in `C:\path\to\file` as escape
sequences. Properly escape the string so that the output-file-map
works properly on Windows.

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -45,6 +45,7 @@
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <iomanip>
 #include <memory>
 #include <mutex>
 #include <set>
@@ -2856,12 +2857,35 @@ public:
     return ExternalCommand::configureAttribute(ctx, name, values);
   }
 
+  template <typename String>
+  std::string escapeForJson(String const& in) const {
+    std::ostringstream ss;
+    for (auto i = in.begin(); i != in.end(); i++) {
+      switch (*i) {
+      case '\\': ss << "\\\\"; break;
+      case '"': ss << "\\\""; break;
+      case '\b': ss << "\\b"; break;
+      case '\n': ss << "\\n"; break;
+      case '\f': ss << "\\f"; break;
+      case '\r': ss << "\\r"; break;
+      case '\t': ss << "\\t"; break;
+      default:
+        if ('\x00' <= *i && *i <= '\x1f') {
+          ss << "\\u" << std::hex << std::setw(4) << std::setfill('0')
+             << (int)*i;
+        } else {
+          ss << *i;
+        }
+      }
+    }
+    return ss.str();
+  }
+
   bool writeOutputFileMap(BuildSystemCommandInterface& bsci,
                           StringRef outputFileMapPath,
                           std::vector<std::string>& depsFiles_out) const {
-    // FIXME: We need to properly escape everything we write here.
     assert(sourcesList.size() == objectsList.size());
-    
+
     SmallString<16> data;
     std::error_code ec;
     llvm::raw_fd_ostream os(outputFileMapPath, ec,
@@ -2873,7 +2897,7 @@ public:
     }
 
     os << "{\n";
-    
+
     // Write the master file dependencies entry.
     SmallString<16> masterDepsPath;
     llvm::sys::path::append(masterDepsPath, tempsPath, "master.swiftdeps");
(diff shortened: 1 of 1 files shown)


lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+  template <typename String>
+  std::string escapeForJson(String const& in) const {
+    std::ostringstream ss;
+    for (auto i = in.begin(); i != in.end(); i++) {
+      switch (*i) {
+      case '\\': ss << "\\\\"; break;
+      case '"': ss << "\\\""; break;
+      case '/': ss << "\\/"; break;
+      case '\b': ss << "\\b"; break;
+      case '\f': ss << "\\f"; break;
```
compnerd: json.org doesn't indicate that `\f` should be escaped.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+      case '\n': ss << "\\n"; break;
+      case '\r': ss << "\\r"; break;
+      case '\t': ss << "\\t"; break;
+      default: ss << *i; break;
```
compnerd: What about unicode sequences?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     return ExternalCommand::configureAttribute(ctx, name, values);
   }
 
+  template <typename String>
+  std::string escapeForJson(String const& in) const {
+    std::ostringstream ss;
+    for (auto i = in.begin(); i != in.end(); i++) {
+      switch (*i) {
+      //FIXME: We need to properly escape unicode sequences.
```
jakepetroules: Unicode sequences do not need to be escaped.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+      case '\\': ss << "\\\\"; break;
+      case '"': ss << "\\\""; break;
+      case '/': ss << "\\/"; break;
+      case '\b': ss << "\\b"; break;
+      case '\n': ss << "\\n"; break;
+      case '\r': ss << "\\r"; break;
+      case '\t': ss << "\\t"; break;
```
jakepetroules: Please escape \f

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
 
+  template <typename String>
+  std::string escapeForJson(String const& in) const {
+    std::ostringstream ss;
+    for (auto i = in.begin(); i != in.end(); i++) {
+      switch (*i) {
+      //FIXME: We need to properly escape unicode sequences.
+      case '\\': ss << "\\\\"; break;
+      case '"': ss << "\\\""; break;
+      case '/': ss << "\\/"; break;
```
jakepetroules: Don't escape / as it is not required and it's very annoying for readability with file paths

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
     return ExternalCommand::configureAttribute(ctx, name, values);
   }
 
+  template <typename String>
+  std::string escapeForJson(String const& in) const {
+    std::ostringstream ss;
+    for (auto i = in.begin(); i != in.end(); i++) {
+      switch (*i) {
```
jakepetroules: Only escape characters less than 0x20, backslash, and quote, otherwise add to the output literally.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -2853,12 +2853,30 @@ class SwiftCompilerShellCommand : public ExternalCommand {
...
+      //FIXME: We need to properly escape unicode sequences.
+      case '\\': ss << "\\\\"; break;
+      case '"': ss << "\\\""; break;
+      case '/': ss << "\\/"; break;
+      case '\b': ss << "\\b"; break;
+      case '\n': ss << "\\n"; break;
+      case '\r': ss << "\\r"; break;
+      case '\t': ss << "\\t"; break;
+      default: ss << *i; break;
```
jakepetroules: All nonprintables should be escaped as \uXXXX where XXXX is the character in hex

lib/BuildSystem/BuildSystem.cpp
jakepetroules: Yes it does.

lib/BuildSystem/BuildSystem.cpp
jakepetroules: Not relevant.

lanza:
@jmittert

jmittert:
I pulled the latest change and it still passes the tests I needed it for.

aciidb0mb3r:
@swift-ci smoke test

jakepetroules:
Thanks!

PR #430, opened 2019-02-11
Fix some process spawning issues on Windows

jmittert:
- CreateProcessW Command args _should_ include the executable name as
  argv[0]
- cmd doesn't know what single quotes are, use double quotes instead on
  Windows

### DIFF
--- a/lib/Basic/ShellUtility.cpp
+++ b/lib/Basic/ShellUtility.cpp
@@ -27,27 +27,32 @@ void appendShellEscapedString(llvm::raw_ostream& os, StringRef string) {
     return;
   }
 
+#if defined(_WIN32)
+  std::string escQuote = "\"";
+#else
+  std::string escQuote = "'";
+#endif
   // We only need to escape the single quote, if it isn't present we can
   // escape using single quotes.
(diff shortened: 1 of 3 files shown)


jmittert:
@lanza With https://github.com/apple/swift-llbuild/pull/429 applied, I'm able to get the swiftpm Windows build all the way to failing on `import Foundation` when calling swiftc.

lib/Basic/ShellUtility.cpp
```diff
@@ -27,27 +27,32 @@ void appendShellEscapedString(llvm::raw_ostream& os, StringRef string) {
     return;
   }
 
+#if defined(_WIN32)
+  std::string escQuote = "\"";
+#else
+  std::string escQuote = "'";
```
jakepetroules: Yes, good. Double quotes are dangerous on Unix since they allow shell expansion.

Of course, this isn't strictly correct on Windows either because the escaping rules for cmd.exe are immensely complicated (and that's ON TOP OF the CommandLineToArgvW related escaping)... but of course that's for another PR. :/

jakepetroules:
@swift-ci please smoke test

PR #431, opened 2019-02-14
Fix warnings introduced by llvm support update

dmbryson:
rdar://problem/48066858


dmbryson:
@swift-ci please smoke test

PR #433, opened 2019-02-19
Handle Carriage Return in MakeFileDepsParser

jmittert:
The deps lexer didn't consider '\r' a non word character, so it would
look for a file "filename.txt\r" on Windows and fail to find it.

### DIFF
--- a/lib/Core/MakefileDepsParser.cpp
+++ b/lib/Core/MakefileDepsParser.cpp
@@ -26,6 +26,7 @@ static bool isWordChar(int c) {
   switch (c) {
   case '\0':
   case '\t':
+  case '\r':
   case '\n':
   case ' ':
   case '$':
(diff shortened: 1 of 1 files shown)


ddunbar:
Can we get a matching test update?

jmittert:
> 
> 
> Can we get a matching test update?

 This is already tested by `tests/SwiftBuildTool/dependencies.swift-build` which fails without this. ([See also](https://github.com/apple/swift-llbuild/pull/434/files#diff-65cda9af668f535261f13f25cd9593f6))

ddunbar:
Fair enough!

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #434, opened 2019-02-19
Fix SwiftBuildToolTests on Windows

jmittert:
This (along with https://github.com/apple/swift-llbuild/pull/433) fixes the SwiftBuildToolTests on Windows. The main issue is the Windows doesn't know what a shebang is so it is unable to exec the python script. It's ugly and I hate that this is the only solution I'm able to come up with, but this adds a shim which the tests compile and execute which then calls the python script.

### DIFF
--- a/tests/SwiftBuildTool/Inputs/swiftc-shim.cpp
+++ b/tests/SwiftBuildTool/Inputs/swiftc-shim.cpp
@@ -1,0 +1,17 @@
+#include <stdlib.h>
+#include <string>
+int main(int argc, char** argv) {
+  std::string args;
+  for (int i = 1; i < argc; i++) {
+    args += " \"";
+    args += argv[i];
+    args += "\"";
+  }
+
+#if defined(_WIN32)
+  system(("python SOURCEDIR/Inputs/pseudo-swiftc" + args).c_str());
+#else
+  system(("SOURCEDIR/Inputs/pseudo-swiftc" + args).c_str());
+#endif
+  return 0;
+}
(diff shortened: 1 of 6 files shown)


tests/SwiftBuildTool/Inputs/swiftc-shim.cpp
```diff
@@ -0,0 +1,17 @@
+#include <stdlib.h>
+#include <string>
+int main(int argc, char** argv) {
+  std::string args;
+  for (int i = 1; i < argc; i++) {
+    args += " \"";
+    args += argv[i];
```
jmittert: This doesn't escape arguments properly, but it seems overkill to link against llvm or implement the logic for the handful of tests that use this.

ddunbar: Thats fine with me.

I would call this file `pseudo-swiftc-shim`, for consistency.

jmittert:
> 
> 
> Hmmm, this is indeed fairly gross.
> 
> We don't actually leverage the builtin Swift tool that much, and I expect it to eventually go away. I wonder if we should just DISABLE these tests on Windows instead?

Is the swift-build-tool in llbuild the same one called by `swiftpm`/`swift-build`? Because if so, I'd prefer the coverage, these found some issues with the `.d` file parsing and the backslashes in the output file maps. If not, I'd agree just disabling them is the better thing to do.

ddunbar:
@jmittert SwiftPM has moved to calling llbuild using its API directly, rather than using swift-build-tool. /cc @aciidb0mb3r

jmittert:
Makes sense. I think I'll disable the tests in a different PR and close this one. Is swiftpm the only thing that uses it? It looks like SwiftPM has an option to enable it, plus it looks for swift-build-tool to find llbuild.

aciidb0mb3r:
> @jmittert SwiftPM has moved to calling llbuild using its API directly, rather than using swift-build-tool. /cc @aciidb0mb3r

We still need `swift-build-tool` for bootstrapping SwiftPM but otherwise we use the API, yeah.

PR #439, opened 2019-03-04
Add list-keys and dump db subcommands

dmbryson:


dmbryson:
@swift-ci please smoke test

PR #441, opened 2019-03-07
Add some missing sqlite3_finalize() statements.

AlanQuatermain:
Some (error) paths in `SQLiteBuildDB.cpp` failed to finalize temporary statements, possibly causing leaks if the resultant errors were handled gracefully further up the stack (i.e. by not quitting immediately). Seemed like a potential issue that'd be easy to clear up—in case some future changes don't treat the resulting errors as fatal (since that doesn't seem to be delineated anywhere).


ddunbar:
@swift-ci please smoke test

ddunbar:
Thanks!!

PR #443, opened 2019-03-08
[BuildSystem] Track and validate rule signatures

dmbryson:
Promotes signatures to a first class feature of rules in the engine and
db. Signatures are supplied during rule construction and validated
against the last signature used to build the prior result during
scanning.

Command signatures have been moved out of the build value and into rule
signatures. Nodes now make use of signatures as track changes in the
build graph.

rdar://problem/48578317

### DIFF
--- a/include/llbuild/BuildSystem/BuildDescription.h
+++ b/include/llbuild/BuildSystem/BuildDescription.h
@@ -15,6 +15,7 @@
 
 #include "llbuild/Basic/Compiler.h"
 #include "llbuild/Basic/ExecutionQueue.h"
+#include "llbuild/Basic/Hashing.h"
 #include "llbuild/Basic/LLVM.h"
 
 #include "llvm/ADT/SmallString.h"
@@ -187,6 +188,8 @@ public:
   
   /// Get a verbose description of the command, for use in status reporting.
   virtual void getVerboseDescription(SmallVectorImpl<char> &result) const override = 0;
+
+  virtual basic::CommandSignature getSignature() const;
   
   /// @}
 
--- a/include/llbuild/BuildSystem/BuildNode.h
+++ b/include/llbuild/BuildSystem/BuildNode.h
@@ -33,30 +33,27 @@ namespace buildsystem {
   
 // FIXME: Figure out how this is going to be organized.
 class BuildNode : public Node {
-  /// Whether or not this node represents a full directory.
-  //
-  // FIXME: We need a type enumeration here.
-  bool directory;
 
-  /// Whether or not this node represents the full directory structure.
-  //
-  // FIXME: We need a type enumeration here.
-  bool directoryStructure;
-  
-  /// Whether or not this node is "virtual" (i.e., not a filesystem path).
-  bool virtualNode;
+  enum class NodeType : unsigned char {
+    Plain = 0,
+    Directory = 1,
+    DirectoryStructure = 2,
+    Virtual = 3,
+  };
+
+  NodeType type = NodeType::Plain;
 
   /// Whether this node represents a "command timestamp".
   ///
   /// Such nodes should always also be virtual.
-  bool commandTimestamp;
+  bool commandTimestamp = false;
 
   /// Whether this node is mutated by the build.
   ///
   /// This flag cannot currently be honored to provide a strongly consistent
   /// build, but it is used to detect when the file system information on a node
   /// cannot be safely used to track *output* file state.
-  bool mutated;
+  bool mutated = false;
 
   /// Exclusion filters for directory listings
   ///
@@ -64,24 +61,20 @@ class BuildNode : public Node {
   /// signature for directory and directory structure nodes.
   basic::StringList exclusionPatterns;
 
-public:
-  explicit BuildNode(StringRef name, bool isDirectory,
-                     bool isDirectoryStructure, bool isVirtual,
-                     bool isCommandTimestamp, bool isMutated)
-      : Node(name), directory(isDirectory),
-        directoryStructure(isDirectoryStructure), virtualNode(isVirtual),
-        commandTimestamp(isCommandTimestamp), mutated(isMutated) {}
+  explicit BuildNode(StringRef name, NodeType type)
+      : Node(name), type(type) {}
 
+public:
   /// Check whether this is a "virtual" (non-filesystem related) node.
-  bool isVirtual() const { return virtualNode; }
+  bool isVirtual() const { return (type == NodeType::Virtual); }
 
   /// Check whether this node is intended to represent a directory's contents
   /// recursively.
-  bool isDirectory() const { return directory; }
+  bool isDirectory() const { return (type == NodeType::Directory); }
 
   /// Check whether this node is intended to represent a directory's structure
   /// recursively.
-  bool isDirectoryStructure() const { return directoryStructure; }
+  bool isDirectoryStructure() const { return (type == NodeType::DirectoryStructure); }
 
   bool isCommandTimestamp() const { return commandTimestamp; }
 
@@ -101,6 +94,20 @@ public:
 
   basic::FileInfo getFileInfo(basic::FileSystem&) const;
   basic::FileInfo getLinkInfo(basic::FileSystem&) const;
+
+  basic::CommandSignature getSignature() const;
+
+  static std::unique_ptr<BuildNode> makePlain(StringRef name) {
+    return std::unique_ptr<BuildNode>(new BuildNode(name, NodeType::Plain));
+  }
+
+  static std::unique_ptr<BuildNode> makeDirectory(StringRef name) {
+    return std::unique_ptr<BuildNode>(new BuildNode(name, NodeType::Directory));
+  }
+
+  static std::unique_ptr<BuildNode> makeVirtual(StringRef name) {
+    return std::unique_ptr<BuildNode>(new BuildNode(name, NodeType::Virtual));
+  }
 };
 
 
--- a/include/llbuild/BuildSystem/BuildValue.h
+++ b/include/llbuild/BuildSystem/BuildValue.h
@@ -94,6 +94,8 @@ class BuildValue {
     /// The filtered contents of a directory.
     FilteredDirectoryContents,
 
+    /// A value produced by a successful command with an output signature.
+    SuccessfulCommandWithOutputSignature,
   };
   static StringRef stringForKind(Kind);
 
@@ -105,8 +107,8 @@ class BuildValue {
   /// The number of attached output infos.
   uint32_t numOutputInfos = 0;
 
-  /// The command hash, for successful commands.
-  basic::CommandSignature commandSignature;
+  /// A hash value (used by some build value types).
+  basic::CommandSignature signature;
 
   union {
     /// The file info for the rule output, for existing inputs, successful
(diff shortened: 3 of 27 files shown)


dmbryson:
See individual commit messages for more detailed explanation of the two changes.

dmbryson:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -772,6 +773,28 @@ class ProducedNodeTask : public Task {
     if (value.isMissingInput())
       return false;
 
+    // Check that the producer is currently recorded as a dependency
+    // This is an N x M scan, but we expect N (number of producers) to be 1
+    for (auto producer : node.getProducers()) {
```
ddunbar: Would it make sense to instead use a "signature" like approach, where we take a signature of the producer and record that? OTOH, that would make the result size bigger, so its not obviously a win. I'm not thrilled about exposing the full result to each callback, but I also don't have any good argument against it.

I also think that this logic can be wrong if a node starts having multiple producers, maybe?

tests/BuildSystem/Build/missing-command-rule.llbuild
```diff
@@ -0,0 +1,57 @@
+# Check the handling of stale file removal.
```
ddunbar: Out of date comment

tests/BuildSystem/Build/missing-command-rule.llbuild
```diff
@@ -0,0 +1,57 @@
...
+client:
+  name: basic
+
+targets:
+  "": ["<all>"]
+
+commands:
+  C.1:
+    tool: stale-file-removal
+    description: STALE-FILE-REMOVAL
```
ddunbar: Out of date description, etc

tests/BuildSystem/Build/missing-command-rule.llbuild
```diff
@@ -0,0 +1,57 @@
...
+#
+# CHECK-VERSION-1: STALE-FILE-REMOVAL
+#
+# RUN: grep -A1000 "VERSION-BEGIN-[2]" %s | grep -B10000 "VERSION-END-2" | grep -ve '^--$' > %t.build/build-2.llbuild
+# Create stale file
+# RUN: touch %t.build/a.out
+# RUN: %{llbuild} buildsystem build --serial --chdir %t.build -f build-2.llbuild > %t2.out
+# RUN: %{FileCheck} --check-prefix CHECK-VERSION-2 --input-file %t2.out %s
+# Check for absence of stale file
+# RUN: test ! -f %t.build/a.out
```
ddunbar: Is this really the right test?

include/llbuild/Core/BuildEngine.h
```diff
@@ -66,6 +66,9 @@ struct Result {
   // FIXME: At some point, figure out the optimal representation for this field,
   // which is likely to be a lot of the resident memory size.
   std::vector<KeyID> dependencies;
+
+  /// Is this result transient.
+  bool transient = false;
```
ddunbar: This makes me nervous. For one thing, it feels like it exposes a potential behavioral difference when running with a cached vs uncached engine. More generally, I am nervous about introducing more semantic options in the core, without a clear understanding of their meaning. It *feels* to me like we should be able to solve this problem without this, which makes me thing this is introducing a semantic behavior that isn't actually necessary (and may be hard to prove correct).

tests/BuildSystem/Build/missing-command-rule.llbuild
dmbryson: Heh, `git` oddity.  The test was added in the follow up commit.  This was Xcode adding it when I made the copy.   I'll fix up the git history.

lib/BuildSystem/BuildSystem.cpp
dmbryson: If a node does start having multiple producers this should still be conservative and classify it as invalid if any of its producers is not currently listed as a dependency.

dmbryson:
I split the second change out from this PR and added a test just for this specific change.

include/llbuild/Core/BuildEngine.h
dmbryson: Moving to a separate PR.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

include/llbuild/BuildSystem/BuildNode.h
```diff
@@ -33,55 +33,48 @@ namespace buildsystem {
...
-  
-  /// Whether or not this node is "virtual" (i.e., not a filesystem path).
-  bool virtualNode;
+
+  enum NodeType {
+    Plain = 0,
+    Directory = 1,
+    DirectoryStructure = 2,
+    Virtual = 3,
+  };
```
ddunbar: Thank you!

ddunbar: Should this use `enum class`, I think we prefer it most other places

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -375,11 +383,6 @@ class BuildValue {
     }
   }
 
-  basic::CommandSignature getCommandSignature() const {
-    assert(isSuccessfulCommand() && "invalid call for value kind");
-    return commandSignature;
-  }
```
ddunbar: If there is no accessor for the signature field, what is it used for?

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -293,6 +294,9 @@ class BuildValue {
   static BuildValue makeFilteredDirectoryContents(ArrayRef<std::string> values) {
     return BuildValue(Kind::FilteredDirectoryContents, values);
   }
+  static BuildValue makeSuccessfulCommandWithOutputSignature(ArrayRef<FileInfo> outputInfos, basic::CommandSignature signature) {
+    return BuildValue(Kind::SuccessfulCommandWithOutputSignature, outputInfos, signature);
```
ddunbar: Should the `commandSignature` field of this type get renamed?

lib/Core/BuildEngine.cpp
```diff
@@ -422,6 +422,13 @@ class BuildEngineImpl : public BuildDBDelegate {
       return true;
     }
 
+    if (ruleInfo.rule.signature != ruleInfo.result.signature) {
```
dmbryson: Here is the key change to behavior.

include/llbuild/BuildSystem/BuildDescription.h
```diff
@@ -187,6 +188,8 @@ class Command : public basic::JobDescriptor {
   
   /// Get a verbose description of the command, for use in status reporting.
   virtual void getVerboseDescription(SmallVectorImpl<char> &result) const override = 0;
+
+  virtual basic::CommandSignature getSignature() const = 0;
```
ddunbar: Should we rename `CommandSignature` (ideally in a follow on commit)?

lib/BuildSystem/BuildNode.cpp
```diff
@@ -27,10 +27,10 @@ bool BuildNode::configureAttribute(const ConfigureContext& ctx, StringRef name,
                                    StringRef value) {
   if (name == "is-directory") {
```
ddunbar: Maybe at the same time we are doing this we should introduce a new setter ("type"), and soft deprecate these old ones?

lib/BuildSystem/BuildNode.cpp
```diff
@@ -126,6 +125,15 @@ FileInfo BuildNode::getLinkInfo(basic::FileSystem& fileSystem) const {
   return fileSystem.getLinkInfo(getName());
 }
 
+basic::CommandSignature BuildNode::getSignature() const {
+  basic::CommandSignature sig;
+  sig.combine(type);
+  for (auto producer : getProducers()) {
```
ddunbar: Probably worth a comment here explaining why we do this. I also would like `auto* producer` here so the reader can be easily convinced there is no copy.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -3309,6 +3338,10 @@ class StaleFileRemovalCommand : public Command {
     llvm::raw_svector_ostream(result) << "]";
   }
 
+  virtual basic::CommandSignature getSignature() const override {
+    return {};
```
ddunbar: Is this safe, shouldn't it have some unique type code assigned to it (so rule type changes can be detected)? Actually, the same comment applies to all the other places a null signature is created, maybe they should each have a unique code.

lib/BuildSystem/ExternalCommand.cpp
```diff
@@ -223,10 +219,11 @@ void ExternalCommand::start(BuildSystemCommandInterface& bsci,
 
 void ExternalCommand::providePriorValue(BuildSystemCommandInterface&,
                                         core::Task*,
-                                        const BuildValue& value) {
+                                        const BuildValue& value,
+                                        const basic::CommandSignature& signature) {
   if (value.isSuccessfulCommand()) {
     hasPriorResult = true;
-    priorResultCommandSignature = value.getCommandSignature();
```
dmbryson: This is kind of ugly. I think that it would be preferable for tasks to not use the signature, since the engine has already decided it should run.   However, the 'can-update-if-newer' bits below rely on this to internally decide to skip the computation anyway.

dmbryson:
@swift-ci please smoke test

lib/BuildSystem/ExternalCommand.cpp
```diff
@@ -223,10 +219,11 @@ void ExternalCommand::start(BuildSystemCommandInterface& bsci,
...
+    priorResultCommandSignature = signature;
   }
```
ddunbar: Question: If a commands signature is changed, should it really receive the priorValue? I think the answer may be no, and if it is no, then the `signature` argument to `providePriorValue` can go away (because this is the only place we use it, and we basically use it because this wasn't modeled where we now think it should be modeled)

include/llbuild/BuildSystem/BuildValue.h
ddunbar: I get it now

lib/BuildSystem/BuildSystem.cpp
dmbryson: For many of the discrete key types it is unnecessary as the key itself would change. I think the same applies to these commands as well, but I'll take a second look.

lib/BuildSystem/ExternalCommand.cpp
dmbryson: Resolved this by only providing the prior value to the rule if the signature has not changed, as discussed in person.

lib/BuildSystem/BuildSystem.cpp
dmbryson: Yeah, after closer inspection I agree these should have signatures. I've added a default implementation for the Command object and removed these null signature overrides.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

PR #445, opened 2019-03-08
[Commands] Fix printf formatting warning

dmbryson:
Linux and macOS differ in what they consider uint64_t to be (%lu vs.
%llu, respectively).


dmbryson:
@swift-ci please smoke test

dmbryson:
Yeah, I'd seen `PRIu64` before but hadn't grokked its meaning.  I filed it under "you learn something new about C++ every day".

PR #447, opened 2019-03-13
[plugins] Allocate shell handlers once per build.

ddunbar:
- This is more efficient, while also allowing extensions to easily tie their
   lifetime to the lifetime of the build.

 - This also allows the buildsystem to avoid a lock on the common lookup path.

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -208,6 +208,9 @@ private:
 
   /// Flag indicating if the build has been cancelled.
   std::atomic<bool> isCancelled_{ false };
+
+  /// Cache of instantiated shell command handlers.
+  llvm::StringMap<std::unique_ptr<ShellCommandHandler>> shellHandlers;
   
   /// @name BuildSystemCommandInterface Implementation
   /// @{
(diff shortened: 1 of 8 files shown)


ddunbar:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -208,6 +208,9 @@ class BuildSystemImpl : public BuildSystemCommandInterface {
 
   /// Flag indicating if the build has been cancelled.
   std::atomic<bool> isCancelled_{ false };
+
+  /// Cache of instantiated shell command handlers.
+  llvm::StringMap<std::unique_ptr<ShellCommandHandler>> shellHandlers;
```
dmbryson: Shouldn't we be clearing this at the start of build()?   As it currently stands we will reuse these across multiple builds if the engine is cached.

ddunbar: Oops, yes, indeed.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test osx

dmbryson:
@swift-ci please smoke test osx

dmbryson:
@swift-ci please smoke test

PR #449, opened 2019-03-13
Replace write-swift-version with a python script

jmittert:
Bash isn't that Windows friendly. Let's do this in python instead.


jakepetroules:
Can't you just do this in CMake language instead? Then there is no external tool dependency...

jmittert:
Is it fine to create the file at configure time instead of build time? I don't quite understand its purpose from just looking at how it's used.

dmbryson:
The generated file is a dependency of the swiftmodule build command.  My guess is that this file is used to ensure that the swiftmodule is rebuilt on incremental builds of Swift itself as the version changes.

dmbryson:
Or in other words, it does need to be checked each time the build is run to achieve the same effect.   It may be possible to do in CMake, but I haven't looked that deeply.

dmbryson:
I believe the need for this has been obviated by #453

PR #452, opened 2019-03-14
Use Windows Lean and Mean

jmittert:
Importing non lean and mean Windows.h conflicts with the lean and mean
Windows imported by (Core)Foundation


ddunbar:
@swift-ci please smoke test

PR #454, opened 2019-03-14
BuildSystem: use `llvm-ar` rather than `ar`

compnerd:
Unlike UNIX systems, Windows does not have an archiver.  Traditionally,
this role has been fulfilled by `link`.  Since we already need `clang`
and `LLVM`, let us just switch to `llvm-ar` which can support BSD style
archives (Darwin), GNU style archives (Linux), and SysV archives
(Windows).

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -3252,7 +3252,10 @@ class ArchiveShellCommand : public ExternalCommand {
 
   std::vector<std::string> getArgs() const {
     std::vector<std::string> args;
-    args.push_back("ar");
+    if (const char *ar = std::getenv("AR"))
+      args.push_back(std::string(ar));
+    else
+      args.push_back("ar");
     args.push_back("cr");
     args.push_back(archiveName);
     args.insert(args.end(), archiveInputs.begin(), archiveInputs.end());


compnerd:
@swift-ci please test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -3252,7 +3252,7 @@ class ArchiveShellCommand : public ExternalCommand {
 
   std::vector<std::string> getArgs() const {
     std::vector<std::string> args;
-    args.push_back("ar");
+    args.push_back("llvm-ar");
```
ddunbar: I would prefer to conditionalize this on Win32, I'm not sure we should avoid using whatever is named "ar" on other systems.

compnerd: Sure, I can conditionalize it to Win32, but, I do think that we really should prefer to use the LLVM tools.  They give us broader compatibility as the single tool can then be used to cross-compile (e.g. I can build a BSD archive on Linux) which I cannot do otherwise.

ddunbar: Cross-compilation is definitely something I'd like to support, but llbuild is general purpose and can easily be used on a system without LLVM/Clang. It'd be a different story if we were directly integrating it to provide the `ar` functionality, but if we are shelling out I'd like to use the standard name.

compnerd: Well, that is the problem - there is no standard name that works across platforms.  Perhaps we should do this the GNU way - use `AR` from the environment and fall back to `ar`?

ddunbar: That sounds reasonable to me.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

keith:
I had a conversation with Daniel on slack about actually changing the invocation to call libtool in some cases. Specifically we found an issue where if you had a .o file with no symbols and create an archive from it with ar, it will fail to link into the final binary since it's missing the table of contents. If you link the same thing with libtool this isn't a problem. I'm not sure how that fits in with this change.

compnerd:
@keith this wouldn't really be related to that.  If you are building a single library, I suppose that you could use it to use `lipo` rather than `ar`.

compnerd:
@swift-ci please test

keith:
FEIW this currently goes through this same codepath. But we could possible figure that out separately, I'm not sure how that would affect the public API

compnerd:
@ddunbar @dmbryson - ping?

dmbryson:
@swift-ci please smoke test

dmbryson:

PR #455, opened 2019-03-15
Windows: Pass the Environment to Subprocesses

jmittert:
- Properly set up the environment on Windows
- Fix tests which invoke `/usr/bin/env`
- Properly set up the thread array when using WFMO

### DIFF
--- a/tests/BuildSystem/Build/basic.llbuild
+++ b/tests/BuildSystem/Build/basic.llbuild
@@ -3,18 +3,18 @@
 # RUN: rm -rf %t.build
 # RUN: mkdir -p %t.build
 # RUN: touch "%t.build/input A"
-# RUN: cp %s %t.build/build.llbuild
+# RUN: cat %s | sed -e 's#ENV_PATH#%{env}#g' -e 's#SORT_PATH#%{sort}#g' >  %t.build/build.llbuild
 # RUN: %{llbuild} buildsystem build --serial --chdir %t.build --trace %t.trace > %t.out
 # RUN: %{FileCheck} --input-file=%t.out %s
 # RUN: diff "%t.build/input A" %t.build/output
 #
-# CHECK: /usr/bin/env
+# CHECK: env
 # CHECK: ENV_KEY=ENV_VALUE_1
-# CHECK-NOT: PATH=
-# CHECK: /usr/bin/env
+# CHECK-NOT: {{^PATH=}}
+# CHECK: env
 # CHECK: ENV_KEY=ENV_VALUE_2
 # CHECK: LLBUILD_TASK_ID={{[a-f0-9]+}}
-# CHECK: PATH=random-overridden-path
+# CHECK: {{^PATH=random-overridden-path}}
 # CHECK: cp 'input A' output
 
 # Check the engine trace.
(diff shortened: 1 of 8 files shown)


tests/BuildSystem/Build/basic.llbuild
```diff
@@ -3,19 +3,19 @@
...
 # RUN: %{llbuild} buildsystem build --serial --chdir %t.build --trace %t.trace > %t.out
 # RUN: %{FileCheck} --input-file=%t.out %s
 # RUN: diff "%t.build/input A" %t.build/output
 #
-# CHECK: /usr/bin/env
+# CHECK: env
 # CHECK: ENV_KEY=ENV_VALUE_1
-# CHECK-NOT: PATH=
-# CHECK: /usr/bin/env
+# CHECK-NOT: {{^PATH=}}
```
jmittert: Windows also sets `HOMEPATH=` so check for begining of line.

tests/BuildSystem/Build/basic.llbuild
```diff
@@ -3,19 +3,19 @@
...
+# CHECK: env
 # CHECK: ENV_KEY=ENV_VALUE_2
 # CHECK: LLBUILD_TASK_ID={{[a-f0-9]+}}
-# CHECK: PATH=random-overridden-path
-# CHECK: cp 'input A' output
+# CHECK: {{^PATH=random-overridden-path}}
+# CHECK: cp {{'|\^"}}input{{\^?}} A{{'|\^"}} output
```
jmittert: On Windows, this gets shell escaped as `cp ^"input^ A^" output`

dmbryson: I wonder if we should change most of these tests to use descriptions in the manifest so that we don't have to litter these escape sequence matches all over the place.   We could then have a single test that explicitly matches these. Maybe even better would be a platform specific tests, one for windows and one for POSIX (i.e. not windows).  That way it is clearer to the reader what the tests are actually expecting.

jmittert: I'm not sure what you mean by descriptions in the manifest. Do you have an example of what you're referring to?

We could do platform specific tests, but I think there would be a lot of duplication if we wanted to get it to the point where there are no regexes matching patch separators or quotes.

For now, this change needs to only check for ' vs ", I accidentally included some changes from a different patch which attempts to properly shell escape on Windows that will add the ^ everywhere though.

dmbryson: As a random example: https://github.com/apple/swift-llbuild/blob/master/tests/BuildSystem/Build/missing-command-rule.llbuild

There we set 'description:' for each of the rules and match against that.

jmittert: I see, that makes sense to do.

jmittert:
ping

unittests/Basic/POSIXEnvironmentTest.cpp
```diff
@@ -18,15 +18,17 @@ using namespace llbuild;
...
+  POSIXEnvironment env;
+  env.setIfMissing("a", "aValue");
+  env.setIfMissing("b", "bValue");
+  env.setIfMissing("a", "NOT HERE");
 
-    auto result = env.getEnvp();
-    EXPECT_EQ(StringRef(result[0]), "a=aValue");
-    EXPECT_EQ(StringRef(result[1]), "b=bValue");
-    EXPECT_EQ(result[2], nullptr);
+  auto result = env.getEnvp();
```
dmbryson: Would be nice to have the #if move down here and have some coverage for the Windows variant.

jmittert: Sure! Done.

dmbryson:
@swift-ci please smoke test

PR #457, opened 2019-03-16
[tests] Make tests less dependent on fs time resolution

dmbryson:
'touch' can be insufficient to change stat info on low resolution file
systems.

rdar://problem/48946422


dmbryson:
@swift-ci please smoke test

PR #458, opened 2019-03-17
[tests] Fix now-produced-rule.llbuild on HFS+

dmbryson:
"This time for sure!"

rdar://problem/48946422


dmbryson:
@swift-ci please smoke test

PR #459, opened 2019-03-18
Do Proper Shell Escaping on Windows

jmittert:
If the comments are too much/over the top, I can remove/reduce them. They were mostly there for my own benefit, but I left them in because exactly what it's doing is a bit cryptic otherwise.

### DIFF
--- a/include/llbuild/Basic/ShellUtility.h
+++ b/include/llbuild/Basic/ShellUtility.h
@@ -15,11 +15,19 @@
 
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/raw_ostream.h"
-
-using namespace llvm;
+#include <vector>
 
 namespace llbuild {
 namespace basic {
+
+#if defined(_WIN32)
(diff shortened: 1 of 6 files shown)


jakepetroules:
@swift-ci please smoke test

jmittert:
Oh hold up, llvm has a function to quote args which does the same thing, I'll use that instead. However, it doesn't have one to escape the meta chars which we need if we're calling cmd.exe.

jakepetroules:
Quoting for cmd.exe is different from quoting for CreateProcess, too. The former is much harder...

jmittert:
Finally got around to fixing this. Also enhanced the tests somewhat.

This no longer applies metachar escaping, which is correct for CreateProcessW, but means when it prints out the commands, there are cases (if the command contains a `"`) where you wouldn't just be able to copy and past the command directly into cmd.exe and press enter. However, this means less visual clutter when printing it out and would still be fine for other shells with more sane quoting (e.g. powershell).

jakepetroules:
@swift-ci please smoke test

PR #460, opened 2019-03-24
build: adjust for new build layout

compnerd:
Moving CoreFoundation into the root is needed to actually get the proper
dependency tracking for Foundation builds. This adjusts the use of it
in llbuild.

### DIFF
--- a/products/llbuildSwift/CMakeLists.txt
+++ b/products/llbuildSwift/CMakeLists.txt
@@ -13,8 +13,8 @@ else()
   if(FOUNDATION_BUILD_DIR)
     list(APPEND additional_args
       -L${FOUNDATION_BUILD_DIR}
-      -Fsystem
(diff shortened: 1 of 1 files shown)


dmbryson:
@swift-ci please smoke test

compnerd:
@dmbryson - this won't pass on its own, it needs at least, apple/swift-corelibs-foundation#2004 apple/swift-corelibs-xctest#261, apple/swift-package-manager#2054.  I think that we should commit this without testing once the change in swift-corelibs-foundation passes as that will pull this change in for testing as well.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
Alright, I've done this in a staging manner where we can be compatible with both and then I can come around and clean up.

compnerd:
@swift-ci please smoke test

PR #291, opened 2018-04-27
[BuildSystem] Enable fast cancellation.

ddunbar:
- This uses the new engine support for immediately cancelling, rather than
   waiting for all tasks to be processed and marked as cancelled.

PR #461, opened 2019-03-26
[BuildSystem] Enable fast cancellation

dmbryson:
Re-enable use of engine support for fast cancellation. This time,
updating BuildSystemFrontend to mark the build as unsuccessful when
cancelled.

rdar://problem/31792710

### DIFF
--- a/lib/BuildSystem/BuildSystemFrontend.cpp
+++ b/lib/BuildSystem/BuildSystemFrontend.cpp
@@ -792,9 +792,18 @@ bool BuildSystemFrontend::build(StringRef targetToBuild) {
   if (!buildSystem->build(targetToBuild))
     return false;
 
+  bool wasCancelled = false;
+  auto impl = static_cast<BuildSystemFrontendDelegateImpl*>(delegate.impl);
+  if (impl->getStatus() == BuildSystemFrontendDelegateImpl::Status::Cancelled) {
+    wasCancelled = true;
+  }
+
   // The build was successful if there were no failed commands or unspecified
-  // errors.
+  // errors, including cancellation. We explicitly include cancellation because
+  // it is possible for a build to complete with no failed commands or errors,
+  // yet not have done any actual work (i.e. if we are cancelled immediately).
   //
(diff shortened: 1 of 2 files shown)


lib/BuildSystem/BuildSystemFrontend.cpp
```diff
@@ -792,9 +792,16 @@ bool BuildSystemFrontend::build(StringRef targetToBuild) {
...
+  if (impl->getStatus() == BuildSystemFrontendDelegateImpl::Status::Cancelled) {
+    wasCancelled = true;
+  }
+
   // The build was successful if there were no failed commands or unspecified
   // errors.
   //
   // It is the job of the client to report a summary, if desired.
-  return delegate.getNumFailedCommands() == 0 && delegate.getNumErrors() == 0;
+  return !wasCancelled && delegate.getNumFailedCommands() == 0
```
dmbryson: This was the cause of downstream test failures when #291 was first landed.   In that scenario, the client triggers cancellation immediately and thus no tasks actually start (thus none fail).  Because we previously returned success (i.e. true), the client then proceeded to attempt to validate that all of the commands it expected to run and failed because none of them actually were run.

dmbryson: (and in writing this, it reminds me to put this into a comment in the code)

dmbryson:
@swift-ci please smoke test

dmbryson:

lib/BuildSystem/BuildSystemFrontend.cpp
ddunbar: 👍

dmbryson:

dmbryson:
:-/ macOS-36 continues to spuriously fail llbuild smoke tests.  cc @shahmishal

shahmishal:
@dmbryson I just requested macOS-36 to be cleaned, however it's now running on macOS-44.

dmbryson:
Thanks @shahmishal.

PR #462, opened 2019-03-26
[llvm] Fix or silence analyzer warnings

dmbryson:
This does not attempt to address all of the warnings, as we really want
to just import llvm's support code.  However, we also want to quell the
bots' ire.

rdar://problem/48374266


dmbryson:
@swift-ci please smoke test

dmbryson:

PR #463, opened 2019-03-27
[perftests] Clean up after testSupprocessSpawnWorkingDirectory

dmbryson:
Spawning processes with a working directory may leave the thread working
directory set.  Clean up by resetting it to the process level working
dir.

rdar://problem/48978131

### DIFF
--- a/perftests/Xcode/PerfTests/BuildSystemPerfTests.mm
+++ b/perftests/Xcode/PerfTests/BuildSystemPerfTests.mm
@@ -18,6 +18,11 @@
 
 #import <XCTest/XCTest.h>
 
+extern "C" {
+    // Provided by System.framework's libsystem_kernel interface
+    extern int __pthread_fchdir(int fd);
+}
+
 using namespace llbuild;
 using namespace llbuild::basic;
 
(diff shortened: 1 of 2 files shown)


dmbryson:
@swift-ci please smoke test

perftests/Xcode/PerfTests/BuildSystemPerfTests.mm
```diff
@@ -109,6 +114,10 @@ - (void)testSupprocessSpawnWorkingDirectory {
             spawnProcess(delegate, nullptr, pgrp, handle, cmd, environment, attr, std::move(releaseFn), std::move(completionFn));
         }
     }];
+
+    // Reset (remove) per-thread working directory in case it was set by the
+    // above process spawning.
+    __pthread_fchdir(-1);
```
ddunbar: Should this go in the process spawning code rather than here?

dmbryson:

perftests/Xcode/PerfTests/BuildSystemPerfTests.mm
dmbryson: The process code currently is optimized for common case performance when used under the execution queue.  If we expand our use of the subprocess code, we should revisit that behavior. For our current use of it, I think it is fine to leave it as is.

dmbryson:

PR #464, opened 2019-03-31
Introduce a CommandProtocol for SwiftPM tests

hartbit:
This introduces a `CommandProtocol` which `Command` conformances to and which is used in `BuildDelegate` so we can more easily unit-test SwiftPM's `BuildDelegate` with mock commands.

@ddunbar @aciidb0mb3r Does this sound okay to you?

### DIFF
--- a/products/llbuildSwift/BuildSystemBindings.swift
+++ b/products/llbuildSwift/BuildSystemBindings.swift
@@ -197,17 +197,28 @@ extension SchedulerAlgorithm {
     }
 }
 
+/// Represents a BuildSystem command.
+public protocol CommandProtocol: CustomStringConvertible, CustomDebugStringConvertible {
+    /// The command name.
+    //
+    // FIXME: We shouldn't need to expose this to use for mapping purposes, we should be able to use something more efficient.
+    var name: String { get }
+
+    /// Whether the default status reporting shows status for the command.
+    var shouldShowStatus: Bool { get }
+
+    /// The verbose description provided by the command.
+    var verboseDescription: String { get }
+}
+
 /// Handle for a command as invoked by the low-level BuildSystem.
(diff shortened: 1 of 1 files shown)


ddunbar:
For some reason I’m ambivalent — it doesn’t feel like llbuild should need to do this for the testabiloty of something. What if instead we made Command have a constructable version that clients could call, would that also work? For some reason that seems more appropriate to me

hartbit:
That's another option, but that requires `Command` to be modified to contain extra properties `name, `shouldShowStatus`, `description`, `verboseDescription`: are you okay with that?

```swift
struct Command {
    private let pointer: OpaquePointer?
    private let name: String?
    private let shouldShowStatus: Bool?
    private let description: String?
    private let verboseDescription: String?

    fileprivate init(_ handle: OpaquePointer?) {
        self.handle = handle
    }

    public init(name: String, shouldShowStatus: Bool, description: String, verboseDescription: String) {
        self.handle = nil
        self.name = name
        self.shouldShowStatus = shouldShowStatus
        self.description = description
        self.verboseDescription = verboseDescription
    }

    var name: String {
        if let name = name { return name }
        var data = llb_data_t()
	withUnsafeMutablePointer(to: &data) { (ptr: UnsafeMutablePointer<llb_data_t>) in
            llb_buildsystem_command_get_name(handle, ptr)
        }
        return stringFromData(data)
    }

    // ...
}
```

ddunbar:
I wouldn't implement it that way, I would make handle an enum which could either be opaque (using llb_, or it could hold the other values). Otherwise, yeah, that is what I expected

hartbit:
Yeah, I was thinking about the num solution too. Sounds good.

hartbit:
Closing as re-implemented in https://github.com/apple/swift-llbuild/pull/466

PR #465, opened 2019-04-01
Update manifest to tools version 5

aciidb0mb3r:
This updates the manifest to tools version 5 so we can use the new API
for setting linker flags. There is also a version-specific manifest for
older Swift 4.2 clients.


aciidb0mb3r:
@swift-ci smoke test

SDGGiesbrecht:
🎉 

Thanks!

jakepetroules:
:+1:

PR #467, opened 2019-04-09
Fix ArchiveShellCommand's verbose description output.

AlanQuatermain:
`ArchiveShellCommand::getVerboseDescription()` had wasn't correctly using the `bool first` variable it created. It was set to `true`,  but the loop only ever checked if it was `false` and, if so, it would insert a space and set it to false. Since it never was false, the verbose description would never contain spaces between arguments.

This change makes this code do what I imagine was the intent: insert a space _before_ all arguments except the first.


dmbryson:
@swift-ci please smoke test

dmbryson:
@shahmishal ubuntu16.04-10 appears to have a full disk

dmbryson:

dmbryson:
@swift-ci please smoke test

PR #473, opened 2019-04-17
Add Shared Library Tool

gmittert:
Added a tool to link and produce .dll/.so outputs.

I'm planning on using this in SwiftPM for linking Windows Swift code since Windows Swift currently doesn't support static linking. It currently works for that purpose and produces .dlls and import libs on Windows which I've used to (partially) compile SwiftPM.

### DIFF
--- a/docs/buildsystem.rst
+++ b/docs/buildsystem.rst
@@ -785,6 +785,28 @@ A typical use for this tool is creating static libraries.
    preferable in future to correctly update/delete/create the archive file
    as required.
 
+Shared Library Tool
+------------
+
+**Identifier**: *shared-library*
+
+A tool used to create a shared library (``.so``, ``.dll``)
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 80
+
+   * - Name
+     - Description
+
+   * - executable
+     - A string indicating the path to a compiler that will be used to link
+       the objects
+
+   * - compiler-style
+     - The type of compiler pass, one of: ``swiftc``, ``clang``, ``cl``
+
+
 Stale File Removal Tool
 -----------------------
 
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -3332,6 +3332,183 @@ public:
   }
 };
 
+#pragma mark - SharedLibraryTool implementation
+
+class SharedLibraryShellCommand : public ExternalCommand {
+
+  // Defaults compilers. Can be overwritten on the build config is required.
+  std::string executable = "";
+  std::string sharedLibName;
+  std::string compilerStyle;
+  std::vector<std::string> sharedLibInputs;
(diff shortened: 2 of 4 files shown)


lib/BuildSystem/BuildSystem.cpp
```diff
@@ -3320,6 +3320,195 @@ class ArchiveTool : public Tool {
...
+        args.push_back(clangExecutable);
+      } else if (language == "C++") {
+        args.push_back(clangPlusPlusExecutable);
+      }
+      args.insert(args.end(), sharedLibInputs.begin(), sharedLibInputs.end());
+      args.push_back("-o");
+      args.push_back(sharedLibName);
+      args.insert(args.end(), otherArgs.begin(), otherArgs.end());
+
+#if defined(_WIN32)
```
gmittert: Do we care about cross compilation? If so, I'll add a target parameter and base the options added on that.

jakepetroules: Yes, please do so. LLVM philosophy :)

docs/buildsystem.rst
```diff
@@ -785,6 +785,36 @@ A typical use for this tool is creating static libraries.
...
+
+   * - swiftExecutable
+     - A string indicating the path to a ``swiftc`` compiler that will be used
+       to link Swift code.
+
+   * - clangExecutable
+     - A string indicating the path to a ``clang`` compiler that will be used
+       to link C code.
+
+   * - clang++Executable
```
jakepetroules: It feels weird having "+" in an identifier, can you please use something like "clangPlusPlusExecutable" instead?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -3320,6 +3320,195 @@ class ArchiveTool : public Tool {
   }
 };
 
+#pragma mark - SharedLibraryTool implementation
+
+class SharedLibraryShellCommand : public ExternalCommand {
+
+  std::string swiftExecutable = "swiftc";
+#if defined(_WIN32)
```
jakepetroules: Please add a comment that these are only defaults (to make clear that the compile time check for host OS isn't preventing cross compilation).

gmittert:
Huh, so turns out detecting what compiler you've been given is... somewhat difficult... to say the least. Rather than trying to guess the compiler by its executable name or determining what it is by running it. I've just added a "compiler-style" field when can be specified to be "cl, clang, or swiftc".

gmittert:
<OneWeekCourtesyPing>

gmittert:
@swift-ci please test

gmittert:
@swift-ci please smoke test

gmittert:
rebased and addressed feedback

@swift-ci please smoke test

gmittert:
@jakepetroules Re I believe I've addressed your requested changes.

PR #474, opened 2019-04-18
Close open files, even when exiting early due to an error.

AlanQuatermain:
Always Be Closing.


dmbryson:
@swift-ci please smoke test

PR #453, opened 2019-03-14
build: adjust for Windows

compnerd:

PR #475, opened 2019-04-26
Fix Build Warnings on Windows

gmittert:
- Properly link against msvcrt by not overriding the /MD in
  CMAKE_CXX_FLAGS_RELWITHDEBINFO
- Rename libllbuild.lib back to llbuild.lib
- Rename libllbuildSwift.lib back to llbuildSwift.lib
- Disable incremental linking so llbuild.ilk's don't conflict

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,44 +146,40 @@ if (XCODE)
   # Disable headermaps.
   set(CMAKE_XCODE_ATTRIBUTE_USE_HEADERMAP "NO") 
 elseif(MSVC)
-  # Disable unknown pragma warnings (e.g. for #pragma mark).
-  add_compile_options(-wd4068)
-
-  # TODO: these warnings come from llvmSupport. Since we don't want to diverge from llvmSupport by
-  # addressing these warnings , disable these for now to clean the build log.
-  # If/when we move to use LLVM's own llvmSupport, we should reenable these warnings.
-  # Unfortunately, the headers are intermingled into the public headers
-  # directory so we cannot mark the include search path as a SYSTEM directory
-  # and avoid the warnings.
   add_compile_options(
-    -wd4141 # 'inline' used more than once
-    -wd4146 # Unary minus applied to unsigned type
-    -wd4244 # Signed conversion
-    -wd4267 # Possible loss of data conversions
-    -wd4291 # operator new with no matching delete found
-    -wd4800 # Forcing value to bool 'true' or 'false'
-    -wd4996 # POSIX name for this item is deprecated
-  )
+    # Disable unknown pragma warnings (e.g. for #pragma mark).
+    "-wd4068"
+
+    # TODO: these warnings come from llvmSupport. Since we don't want to diverge from llvmSupport by
+    # addressing these warnings , disable these for now to clean the build log.
+    # If/when we move to use LLVM's own llvmSupport, we should reenable these warnings.
+    "-wd4141" # 'inline' used more than once.
+    "-wd4146" # Unary minus applied to unsigned type.
+    "-wd4244" # Signed conversion.
+    "-wd4267" # Possible loss of data conversions.
+    "-wd4291" # Operator new with no matching delete found.
+    "-wd4800" # Forcing value to bool 'true' or 'false'.
+    "-wd4996" # POSIX name for this item is deprecated.
+    "/EHsc")
 else ()
-  # Compile with C++14, without RTTI or exceptions.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fno-exceptions")
-  # Enable additional Clang warnings.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmost")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wdocumentation  -Woverloaded-virtual")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wparentheses -Wswitch")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wunused-function -Wunused-variable")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wunused-value -Wempty-body")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wuninitialized -Wconstant-conversion")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wint-conversion -Wbool-conversion")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wenum-conversion -Wsign-compare")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnewline-eof -Wdeprecated-declarations")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Winvalid-offsetof")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnon-virtual-dtor")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wimplicit-fallthrough")
+  add_compile_options(
+    # Enable additional Clang warnings.
+   "-Wmost"
+   "-Wdocumentation" "-Woverloaded-virtual"
+   "-Wparentheses" "-Wswitch"
+   "-Wunused-function" "-Wunused-variable"
+   "-Wunused-value" "-Wempty-body"
+   "-Wuninitialized" "-Wconstant-conversion"
+   "-Wint-conversion" "-Wbool-conversion"
+   "-Wenum-conversion" "-Wsign-compare"
+   "-Wnewline-eof" "-Wdeprecated-declarations"
+   "-Winvalid-offsetof"
+   "-Wnon-virtual-dtor"
+   "-Wimplicit-fallthrough")
 endif ()
 
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DLLVM_ON_WIN32")
+  add_compile_definitions("LLVM_ON_WIN32")
 endif()
 
 # On Linux, we may need to include a workaround for legacy libstdc++.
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -164,47 +164,47 @@ if (XCODE)
...
-  # addressing these warnings , disable these for now to clean the build log.
-  # If/when we move to use LLVM's own llvmSupport, we should reenable these warnings.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4996") # POSIX name for this item is deprecated.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4244") # Signed conversion.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4267") # Possible loss of data conversions.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4291") # Operator new with no matching delete found.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4800") # Forcing value to bool 'true' or 'false'.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4146") # Unary minus applied to unsigned type.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4141") # 'inline' used more than once.
+  add_compile_options(
```
gmittert: If I understand correctly, setting CMAKE_CXX_FLAGS overrides the value of CMAKE_CXX_FLAGS_RELWITHDEBINFO.  On Windows this means that /MD doesn't get passed in. Using add_compile_options properly appends to the correct CMAKE_CXX_FLAGS variable.

products/llbuildSwift/CMakeLists.txt
```diff
@@ -24,11 +24,32 @@ else()
...
+    list(APPEND additional_args -Xcc;-DDEPLOYMENT_TARGET_WINDOWS;-Xcc;-D_DLL;-Xlinker;/NODEFAULTLIB:libcmt)
+  else()
+    message(SEND_ERROR "unknown deployment target: ${CMAKE_SYSTEM_NAME}")
+  endif()
+  if(LLBUILD_PATH_TO_SQLITE_SOURCE)
+    list(APPEND additional_args -I${LLBUILD_PATH_TO_SQLITE_SOURCE})
+  endif()
+  if(LLBUILD_PATH_TO_SQLITE_BUILD)
+    list(APPEND additional_args -L${LLBUILD_PATH_TO_SQLITE_BUILD})
+  endif()
```
gmittert: This is (partially) from @compnerd's patch that got reverted (https://github.com/apple/swift-llbuild/pull/453/files). DEPLOYMENT_TARGET has to be set to build with foundation on Windows.

cmake/modules/Utility.cmake
```diff
@@ -88,6 +88,17 @@ macro(add_llbuild_library name)
...
+    if(ARG_OUTPUT_NAME)
+      set_target_properties(${name}
+        PROPERTIES
+        PDB_NAME lib${ARG_OUTPUT_NAME})
+	else()
+      set_target_properties(${name}
+        PROPERTIES
+        PDB_NAME lib${name})
+    endif()
+  endif()
```
gmittert: Now on Windows, this produces:
bin/llbuild.exe
bn/llbuild.pdb
bin/libllbuild.pdb
lib/llbuild.lib
lib/llbuildSwift.lib

And -lllbuild or -lllbuildswift will work across platforms. Darwin/Linux should be unchanged.

CMakeLists.txt
```diff
@@ -164,47 +164,47 @@ if (XCODE)
...
+   "-Wenum-conversion" "-Wsign-compare"
+   "-Wnewline-eof" "-Wdeprecated-declarations"
+   "-Winvalid-offsetof"
+   "-Wnon-virtual-dtor"
+   "-Wimplicit-fallthrough")
 endif ()
 
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DLLVM_ON_WIN32")
+  add_compile_options("-DLLVM_ON_WIN32")
```
compnerd: Please use `add_compile_definitions` for the defines.

CMakeLists.txt
```diff
@@ -164,47 +164,47 @@ if (XCODE)
...
 endif()
 
 # On BSD and Linux, always build with PIC.
 if(${CMAKE_SYSTEM_NAME} MATCHES ".*BSD" OR ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+  add_compile_options("-fPIC")
 endif ()
```
compnerd: Please replace the 3 lines above with `set(CMAKE_POSITION_INDEPENDENT_CODE YES)`

CMakeLists.txt
```diff
@@ -164,47 +164,47 @@ if (XCODE)
...
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wunused-value -Wempty-body")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wuninitialized -Wconstant-conversion")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wint-conversion -Wbool-conversion")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wenum-conversion -Wsign-compare")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnewline-eof -Wdeprecated-declarations")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Winvalid-offsetof")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnon-virtual-dtor")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wimplicit-fallthrough")
+  add_compile_options(
+   "-std=c++14" "-fno-rtti" "-fno-exceptions"
```
compnerd: Please don't add `-std=c++14`  Use the following instead:
```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED YES)
set(CMAKE_CXX_EXTENSIONS OFF)
```

cmake/modules/Utility.cmake
compnerd: Don't think that you need the `MSVC` check.  Plus, clang is able to generate PDBs as well.  Can you just default the parameter instead to avoid the conditional?

```
if(NOT ARG_OUTPUT_NAME)
  set(ARG_OUTPUT_NAME ${name})
endif()
set_target_properties(${name} PROPERTIES PDB_NAME lib${ARG_OUTPUT_NAME})
```

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -586,13 +590,17 @@ public final class BuildSystem {
...
         let dbPathPtr = strdup(databaseFile)
-        defer { free(dbPathPtr) }
-
         let tracePathPtr = strdup(traceFile ?? "")
+        #endif
+        defer { free(pathPtr) }
+        defer { free(dbPathPtr) }
         defer { free(tracePathPtr) }
 
         // Allocate a C style environment, if necessary.
```
compnerd: I'm really not a fan of the changes here. I don't see a strong reason to fork the paths here.  You can use `_CRT_NONSTDC_NO_DEPRECATE` as a macro to avoid the warnings instead.

products/llbuildSwift/CMakeLists.txt
compnerd: There is a pending PR #477 which should repair this.

products/llbuildSwift/CMakeLists.txt
```diff
@@ -24,11 +24,32 @@ else()
...
 endif()
 
 # Add swift bindings target if swift compiler is present.
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+set(LIBLLBUILD_NAME llbuild)
+else()
+set(LIBLLBUILD_NAME libllbuild)
+endif()
```
compnerd: Can we avoid setting the `OUTPUT_NAME` instead?

products/llbuildSwift/CMakeLists.txt
```diff
@@ -38,19 +59,19 @@ if (SWIFTC_FOUND)
   endif()
 
   # Install both libllbuild and libllbuildSwift.
-  list(APPEND LLBUILD_LIBRARIES "${LLBUILD_LIBRARY_OUTPUT_INTDIR}/libllbuild.${DYLIB_EXT}")
-  list(APPEND LLBUILD_LIBRARIES "${LLBUILD_LIBRARY_OUTPUT_INTDIR}/libllbuildSwift.${DYLIB_EXT}")
+  list(APPEND LLBUILD_LIBRARIES "${LLBUILD_LIBRARY_OUTPUT_INTDIR}/${LIBLLBUILD_NAME}${CMAKE_SHARED_LIBRARY_SUFFIX}")
```
compnerd: We do explicitly pass `SHARED` to those libraries right?

gmittert: Yeah for both.

gmittert:
Rebased and addressed comments

@swift-ci please smoke test

gmittert:
@ddunbar @dmbryson One week courtesy ping

CMakeLists.txt
```diff
@@ -146,44 +146,40 @@ if (XCODE)
...
+    "-wd4146" # Unary minus applied to unsigned type.
+    "-wd4244" # Signed conversion.
+    "-wd4267" # Possible loss of data conversions.
+    "-wd4291" # Operator new with no matching delete found.
+    "-wd4800" # Forcing value to bool 'true' or 'false'.
+    "-wd4996" # POSIX name for this item is deprecated.
+    "/EHsc")
 else ()
-  # Compile with C++14, without RTTI or exceptions.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fno-exceptions")
```
buttaface: These two flags were left out of the list below, oversight? I ask because this change adds EH functions that breaks linking the Swift package manager on Android. I can always add these flags back specifically for Android, but I wonder if you really wanted to keep them here too.

gmittert:
Hrm no, I meant for those not to have any changes, I must have dropped them
by accident while rebasing :( You're welcome to add them back, else, I'll
put up a patch if I don't see one in a few days. Thanks for catching that!

buttaface:
Done in #544, thanks for the quick answer.

PR #476, opened 2019-04-26
Respect schedulerLanes when using the C API.

TaborKelly:
Very simple change to respect `schedulerLanes` when using the C API.


dmbryson:
@swift-ci please smoke test

PR #477, opened 2019-04-26
build: adjust for Windows

compnerd:


gmittert:
@swift-ci please test

compnerd:
@swift-ci please smoke test

cmake/modules/Utility.cmake
```diff
@@ -206,19 +201,20 @@ function(add_swift_module target name deps sources additional_args)
...
-  # FIXME: Ideally, this should be passed from the swift-ci invocation.
-  if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    # Runpath for finding Swift core libraries in the toolchain.
+    # FIXME: Ideally, this should be passed from the swift-ci invocation.
     list(APPEND DYLYB_ARGS -Xlinker -rpath -Xlinker @loader_path/../../macosx)
-  else()
+  elseif(CMAKE_SYSTEM_NAME STREQUAL Linux)
+    list(APPEND DYLYB_ARGS -Xlinker "-rpath=\\$$ORIGIN")
     list(APPEND DYLYB_ARGS -Xlinker "-rpath=\\$$ORIGIN/../../linux")
+  elseif(CMAKE_SYSTEM_NAME STREQUAL Windows)
```
jakepetroules: It's fine to not do this for this patch, but in future I'd put the Windows condition above the Linux one, turn Linux into the "else" case, and then construct the last path component of the rpath argument using the lowercased form of `uname -s` - that way Linux, FreeBSD, and any other Unix-like OSes fall into the default case automatically and get something sensible.

cmake/modules/Utility.cmake
```diff
@@ -206,19 +201,20 @@ function(add_swift_module target name deps sources additional_args)
...
+    # NOTE(compnerd) Windows does not support RPATH
```
jakepetroules: I'd remove your name from this comment; it makes it look like there's a TODO here, but there isn't - Windows simply doesn't have the concept of rpaths, so there's nothing to do.

cmake/modules/Utility.cmake
compnerd: That definitely wouldn't work here...the rpath encodes the OS name itself.

cmake/modules/Utility.cmake
compnerd: Done

compnerd:
@swift-ci please smoke test and merge

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test and merge

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test Linux platform

cmake/modules/Utility.cmake
jakepetroules: I think you might have missed part of my comment:

> [...] construct the last path component of the rpath argument using the lowercased form of uname -s - that way Linux, FreeBSD, and any other Unix-like OSes fall into the default case automatically and get something sensible.

`uname -s` will return Linux on Linux, FreeBSD on FreeBSD, etc., thus the lowercased forms of that would be "linux", "freebsd", which should be correct for all non-Apple, non-Windows platforms.

PR #478, opened 2019-04-26
Explicitly declare static linking dependencies

gmittert:
- Explicitly list out the libraries to link the target against
- Explicitly declare that we're linking static

This allows us to set BUILD_SHARED_LIBS to build shared swift libraries for Windows.


gmittert:
@swift-ci please smoke test

PR #479, opened 2019-04-26
[Windows] Fix libllbuild(Swift) install location

gmittert:
Previously, it was looking for llbuild(Swift).dll in lib, however, the
dll gets put in bin

Requires https://github.com/apple/swift-llbuild/pull/478 to set BUILD_SHARED_LIBS globally.


products/libllbuild/CMakeLists.txt
```diff
@@ -44,7 +44,9 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/
   PATTERN "*.h")
 
 install(TARGETS libllbuild
-  DESTINATION lib${LLBUILD_LIBDIR_SUFFIX}
+  ARCHIVE DESTINATION lib${LLBUILD_LIBDIR_SUFFIX}
+  LIBRARY DESTINATION lib${LLBUILD_LIBDIR_SUFFIX}
+  RUNTIME DESTINATION bin
```
jakepetroules: Cool! Didn't know about the `RUNTIME` specifier.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,34 @@
...
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
+# with the sanitized host architecture name derived from CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
+  if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
+    set("${result_var_name}" "x86_64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
+    set("${result_var_name}" "aarch64" PARENT_SCOPE)
```
jakepetroules: Shouldn't this one be `arm64`?

compnerd: I think that the result needs to be based on the target; Windows and Darwin use `arm64`, Linux and android use `aarch64`.  Uniformity be damned.

gmittert:
It now checks for both cases:
```
  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
    set("${result_var_name}" "aarch64" PARENT_SCOPE)
  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm64")
    set("${result_var_name}" "arm64" PARENT_SCOPE)
```

@swift-ci please test

ddunbar:
@compnerd what is your opinion on this, should we still have it open?

compnerd:
If llbuild's swift bindings are supposed to be available to others outside of swift-package-manager, this fixes one thing - the installation of the Swift module and doc, but needs to be rebased.  If those are not meant to be used outside of swift-package-manager, this can probably be closed.

gmittert:
I believe this is no longer needed. The behavior of checking `"${LLBUILD_LIBRARY_OUTPUT_INTDIR}` instead of bin no longer exists.

PR #482, opened 2019-05-21
Fixed assertion checks in SQLiteBuildDB

BenchR267:
Used the correct number of expected columns for those SQL statements.

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -706,7 +706,7 @@ public:
 
     os << "keys:\n";
     while (sqlite3_step(stmt) == SQLITE_ROW) {
-      assert(sqlite3_column_count(stmt) == 1);
+      assert(sqlite3_column_count(stmt) == 2);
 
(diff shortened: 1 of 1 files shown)


ddunbar:
@swift-ci please smoke test

BenchR267:
I don't know but they clearly are wrong (comparing the statements vs the expectations).

jakepetroules:
It's possible the `assert` is getting compiled out in our tests.

BenchR267:
If the `assert`s get compiled out during development and in the tests it might not be useful to have them there 🙂 Should we use a different mechanism instead of assert?

ddunbar:
They will be compiled out in release builds, but I assumed we had a debug builder somewhere.

@dmbryson can you check on this (and the PR)?

dmbryson:
We do have a debug builder.  This particular method `void dump(raw_ostream& os)`, however, is not exercised in our tests.  (my fault, d'oh!)

The changes look good.  We probably _should_ have a test that exercises this code.  That said, it is only used for the command line db debug tool, so I wouldn't consider adding such coverage a blocker for this PR.

PR #483, opened 2019-05-21
Made KeyID usages in BuildDBDelegate const

BenchR267:
KeyID used in the methods for BuildDBDelegates should be const.


ddunbar:
@swift-ci please smoke test

PR #484, opened 2019-05-21
Added basic Swift interface for build database

BenchR267:
The build database is not accessible from the Swift interface. While that's okay for building we might want to make it accessible to do calculations on the information we have in the database.
Added a non-destructive version of the sqlite database which doesn't re-create the database automatically if versions mismatch. In this case, an error is thrown instead.

**NOTE**: This change requires the changes in https://github.com/apple/swift-llbuild/pull/483.

This enables the following usage:
```Swift
import llbuild

let path = "path/to/build.db"

// if clientSchemaVersion is different than in the specified database, an error will be thrown
// we still want to explicitly set this version to avoid wrong expectations
let db = try BuildDB(path: path, clientSchemaVersion: 1)
let keys = try db.getKeys()
dump(keys)
```

### DIFF
--- a/include/llbuild/Core/BuildDB.h
+++ b/include/llbuild/Core/BuildDB.h
@@ -138,9 +138,10 @@ public:
 /// \param clientSchemaVersion An uninterpreted version number for use by the
 /// client to allow batch changes to the stored build results; if the stored
 /// schema does not match the provided version the database will be cleared upon
-/// opening.
+/// opening; to avoid this behavior, use `createNondestructiveSQLiteBuildDB`.
 std::unique_ptr<BuildDB> createSQLiteBuildDB(StringRef path,
                                              uint32_t clientSchemaVersion,
+                                             bool recreateUnmatchedVersion,
                                              std::string* error_out);
 
 }
--- a/products/libllbuild/BuildDB-C-API.cpp
+++ b/products/libllbuild/BuildDB-C-API.cpp
@@ -1,0 +1,156 @@
+//===-- BuildDB-C-API.cpp ---------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#include <llbuild/llbuild.h>
+
+#include "llbuild/Core/BuildDB.h"
+
+#include <mutex>
+
+using namespace llbuild;
+using namespace llbuild::core;
+
+namespace {
+  
+  class CAPIBuildDBKeysResult {
+  private:
+    std::vector<KeyType> keys;
+    
+  public:
+    CAPIBuildDBKeysResult(std::vector<KeyType> keys): keys(keys) { }
+    
+    CAPIBuildDBKeysResult(const CAPIBuildDBKeysResult&) LLBUILD_DELETED_FUNCTION;
+    CAPIBuildDBKeysResult& operator=(const CAPIBuildDBKeysResult&) LLBUILD_DELETED_FUNCTION;
+    
+    const uint32_t size() {
+      return keys.size();
+    }
+    
+    const KeyType keyAtIndex(KeyID index) {
+      return keys[index];
+    }
+  };
+  
+  class CAPIBuildDB: public BuildDBDelegate {
+    /// The key table, for memory caching of key to key id mapping.
+    llvm::StringMap<KeyID> keyTable;
+    
+    /// The mutex that protects the key table.
+    std::mutex keyTableMutex;
+    
+    std::unique_ptr<BuildDB> _db;
+    
+  public:
+    CAPIBuildDB(StringRef path, uint32_t clientSchemaVersion, std::string *error_out) {
+      _db = createSQLiteBuildDB(path, clientSchemaVersion, /* recreateUnmatchedVersion = */ false, error_out);
+      if (error_out != NULL) {
+        return;
+      }
+      // The delegate  caches keys for key ids
+      _db.get()->attachDelegate(this);
+      
+      if (!buildStarted(error_out)) {
+        buildComplete();
+        return;
+      }
+    }
+    
+    virtual const KeyID getKeyID(const KeyType& key) override {
+      std::lock_guard<std::mutex> guard(keyTableMutex);
+      
+      // The RHS of the mapping is actually ignored, we use the StringMap's ptr
+      // identity because it allows us to efficiently map back to the key string
+      // in `getRuleInfoForKey`.
+      auto it = keyTable.insert(std::make_pair(key, 0)).first;
+      return (KeyID)(uintptr_t)it->getKey().data();
+    }
+    
+    virtual KeyType getKeyForID(const KeyID key) override {
+      // Note that we don't need to lock `keyTable` here because the key entries
+      // themselves don't change once created.
+      return llvm::StringMapEntry<KeyID>::GetStringMapEntryFromKeyData(
+                                                                       (const char*)(uintptr_t)key).getKey();
+    }
+    
+    const bool buildStarted(std::string *error_out) {
+      return _db.get()->buildStarted(error_out);
+    }
+    
+    void buildComplete() {
+      _db.get()->buildComplete();
+    }
+    
+    const bool getKeys(std::vector<KeyType>& keys_out, std::string *error_out) {
+      return _db.get()->getKeys(keys_out, error_out);
+    }
+  };
+  
+}
+
+const llb_database_t* llb_database_open(
+                                          char *path,
+                                          uint32_t clientSchemaVersion,
+                                          llb_data_t *error_out) {
+  std::string error;
+  
+  auto database = new CAPIBuildDB(StringRef(path), clientSchemaVersion, &error);
+  
+  if (!error.empty() && error_out) {
+    error_out->length = error.size();
+    error_out->data = (const uint8_t*)strdup(error.c_str());
+  }
+  
+  return (llb_database_t *)database;
+}
+
+void llb_database_destroy(llb_database_t *database) {
+  auto db = (CAPIBuildDB *)database;
+  db->buildComplete();
+  delete db;
+}
+
+const llb_database_key_id llb_database_result_keys_get_count(llb_database_result_keys_t *result) {
+  auto resultKeys = (CAPIBuildDBKeysResult *)result;
+  return resultKeys->size();
+}
+
+void llb_database_result_keys_get_key_at_index(llb_database_result_keys_t *result, llb_database_key_id keyID, llb_data_t *key_out) {
+  auto resultKeys = (CAPIBuildDBKeysResult *)result;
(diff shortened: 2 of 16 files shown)


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

products/llbuildSwift/BuildDBBindings.swift
```diff
@@ -0,0 +1,172 @@
...
+    /// The opaque pointer to the database object
+    private var _database: OpaquePointer? = nil
+    
+    /// Initializes the build database at a given path
+    /// If the database at this path doesn't exist, it will created
+    /// If the clientSchemaVersion is different to the one in the database at this path, its content will be automatically erased!
+    public init(path: String, clientSchemaVersion: UInt32) throws {
+        // Safety check that we have linked against a compatibile llbuild framework version
+        if llb_get_api_version() != LLBUILD_C_API_VERSION {
+            fatalError("llbuild C API version mismatch, found \(llb_get_api_version()), expect \(LLBUILD_C_API_VERSION)")
```
jakepetroules: Can you throw an `Error` instead?

products/llbuildSwift/BuildDBBindings.swift
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+/// Database object that defines a connection to a llbuild database
+public final class BuildDB {
+    
+    /// Errors that can happen when opening the database or performing operations on it
+    public enum Error: Swift.Error {
+        /// If the system can't open the database, this error is thrown at init
+        case couldNotOpenDB(String)
```
jakepetroules: Can you please add labels to the parameters here to make it clear what the purpose of the argument is?

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

include/llbuild/Core/BuildDB.h
```diff
@@ -138,11 +138,20 @@ class BuildDB {
...
                                              std::string* error_out);
 
+/// Create a BuildDB instance backed by a SQLite3 database.
+/// If the given client schema version doesn't match the one of the database,
+/// an error will be thrown at use instead of automatically re-creating the
+/// database. This behavior might be useful if the database's content is important
+/// instead of the caching functionality.
+std::unique_ptr<BuildDB> createNondestructiveSQLiteBuildDB(StringRef path,
+                                                           uint32_t clientSchemaVersion,
+                                                           std::string* error_out);
```
ddunbar: This feels like it should be a parameter to `createSQLiteBuildDB`, not an extra method?

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
+//===-- BuildDB-C-API.cpp ---------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2015 - 2019 Apple Inc. and the Swift project authors
```
ddunbar: Wrong dates, FWIW

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
...
+      // The delegate  caches keys for key ids
+      _db.get()->attachDelegate(this);
+      
+      if (!buildStarted(error_out)) {
+        buildComplete();
+        return;
+      }
+    }
+    
+    virtual const KeyID getKeyID(const KeyType& key) override {
```
ddunbar: Let's have a reference here to the BuildEngine copy of the same code, for consistency if one of them changes.

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+    
+    const bool getKeys(std::vector<KeyType>& keys_out, std::string *error_out) {
+      return _db.get()->getKeys(keys_out, error_out);
+    }
+  };
+  
+}
+
+const llb_data_t mapData(std::vector<uint8_t> input) {
```
ddunbar: + `static`, so this function isn't exported out of this file. The name seems odd to me, I'd probably use `_llb_data_new_from_vector` or something

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
...
+  const auto newData = (uint8_t *)malloc(size);
+  memcpy(newData, data, size);
+  
+  return llb_data_t {
+    input.size(),
+    newData
+  };
+}
+
+const llb_database_t* llb_database_create(
```
ddunbar: Should this be `create` or `open`? I think `open` makes more sense, maybe?

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                          uint32_t clientSchemaVersion,
+                                          llb_data_t *error_out) {
+  std::string error;
+  
+  auto database = new CAPIBuildDB(StringRef(path), clientSchemaVersion, &error);
+  
+  if (!error.empty() && error_out) {
+    error_out->length = error.size();
+    error_out->data = (const uint8_t*)strdup(error.c_str());
+  }
```
ddunbar: If there is an error, I think this function should return nullptr (and deallocate)

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
...
+  key_out->length = key.size();
+  key_out->data = (const uint8_t*) strdup(key.data());
+}
+
+void llb_database_destroy_result_keys(llb_database_result_keys_t *result) {
+  delete (CAPIBuildDBKeysResult *)result;
+}
+
+const bool llb_database_get_keys(llb_database_t *database, llb_database_result_keys_t **keysResult_out, llb_data_t *error_out) {
+  auto db = (CAPIBuildDB *)database;
```
ddunbar: For safety, please init `*keysResult_out` to nullptr at start.

products/libllbuild/include/llbuild/db.h
```diff
@@ -0,0 +1,49 @@
+//===- db.h --------------------------------------------------*- C -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
```
ddunbar: Wrong dates

products/libllbuild/BuildDB-C-API.cpp
```diff
@@ -0,0 +1,168 @@
...
+  auto key = resultKeys->keyAtIndex(keyID);
+  key_out->length = key.size();
+  key_out->data = (const uint8_t*) strdup(key.data());
+}
+
+void llb_database_destroy_result_keys(llb_database_result_keys_t *result) {
+  delete (CAPIBuildDBKeysResult *)result;
+}
+
+const bool llb_database_get_keys(llb_database_t *database, llb_database_result_keys_t **keysResult_out, llb_data_t *error_out) {
```
ddunbar: I think this API should return a `llb_database_result_keys_t` on success, or nullptr on error, rather than a bool.

products/libllbuild/BuildDB-C-API.cpp
ddunbar: (Except, as above, I think we should change the signature so this problem will go away)

products/libllbuild/BuildDB-C-API.cpp
ddunbar: Actually, I think we should probably change this API to be:
```
llb_database_enumerate_keys(llb_database_t* db, void *userdata, void (*fn)(void *userdata, llb_data_t key))
```

This is more efficient, and simplifies the memory management.

/cc @dmbryson WDYT?

products/llbuildSwift/BuildDBBindings.swift
```diff
@@ -0,0 +1,172 @@
...
+import Foundation
+
+// We don't need this import if we're building
+// this file as part of the llbuild framework.
+#if !LLBUILD_FRAMEWORK
+import llbuild
+#endif
+
+public typealias KeyID = UInt64
+public typealias KeyType = String
```
ddunbar: I'd prefer we use [UInt8]

products/llbuildSwift/BuildDBBindings.swift
```diff
@@ -0,0 +1,172 @@
...
+        if let error = errorPtr.msg {
+            throw Error.couldNotOpenDB(error: error)
+        }
+    }
+    
+    deinit {
+        llb_database_destroy(_database)
+    }
+    
+    /// Fetches all keys from the database, returns it and saves if as \see allKeys on the database object
```
ddunbar: This comment is wrong I think?

products/llbuildSwift/BuildDBBindings.swift
```diff
@@ -0,0 +1,172 @@
...
+#if !LLBUILD_FRAMEWORK
+import llbuild
+#endif
+
+public typealias KeyID = UInt64
+public typealias KeyType = String
+
+/// Defines the result of a call to fetch all keys from the database
+/// Wraps calls to the llbuild database, but all results are fetched and available with this result
+public class BuildDBKeysResult {
```
ddunbar: Let's land the C API, then we can decide what the Swift API should look like.

include/llbuild/Core/BuildDB.h
BenchR267: Yep, that's whatI had before. But I wanted to give it a default value which is only possible if the error_out parameter has a default value as well. Would you rather have it as a parameter without a default value then?

products/libllbuild/BuildDB-C-API.cpp
BenchR267: It creates an instance of the database type, but opens a channel to an existing database. I think depending on the interpretation both could be right. I'm fine with changing it to open though.

products/libllbuild/BuildDB-C-API.cpp
BenchR267: Turns out this function isn't even used 🤔

products/libllbuild/BuildDB-C-API.cpp
BenchR267: It does simplify the memory management but is it really more efficient? Wouldn't I need to copy *all* values in the Swift layer to have them available? Currently they don't get copied and all calls call down to the C++ instance.

products/llbuildSwift/BuildDBBindings.swift
BenchR267: Opened a new PR [here](https://github.com/apple/swift-llbuild/pull/485) for the C interface. Since I renamed the remote branch, Github automatically closed the PR but I think that's fine. I'll open another for the Swift interface once the C one is merged.

PR #486, opened 2019-05-28
[BuildSystem] Properly hold lock when releasing executionQueue

dmbryson:
rdar://problem/50993380


dmbryson:
@swift-ci please smoke test

PR #487, opened 2019-05-30
Adding fatal command result

BenchR267:
This adds a new enum case to the result of a command, `fatal`. If a command fails fatally, its command process should also fail.
This is used (also included in this PR) for dependency file parsing. This is started as a separate job next to the external command. Before this change, the build (and process) doesn't fail if the dependency parsing fails. With this change, the fail of parsing will be propagated up and the build fails (in addition to the errors that get emitted in `MakefileDepsParser` or its counterpart).

rdar://51069563

There is a related bug that made us think about doing that change in the first place: https://bugs.swift.org/browse/SR-10735

### DIFF
--- a/products/libllbuild/include/llbuild/buildsystem.h
+++ b/products/libllbuild/include/llbuild/buildsystem.h
@@ -111,6 +111,7 @@ typedef enum LLBUILD_ENUM_ATTRIBUTES {
   llb_buildsystem_command_result_failed = 1,
   llb_buildsystem_command_result_cancelled = 2,
   llb_buildsystem_command_result_skipped = 3,
+  llb_buildsystem_command_result_fatal = 4,
 } llb_buildsystem_command_result_t LLBUILD_SWIFT_NAME(CommandResult);
 
 /// Extended result of a command execution
--- a/products/libllbuild/include/llbuild/buildsystem.h
+++ b/products/libllbuild/include/llbuild/buildsystem.h
@@ -111,6 +111,7 @@ typedef enum LLBUILD_ENUM_ATTRIBUTES {
   llb_buildsystem_command_result_failed = 1,
   llb_buildsystem_command_result_cancelled = 2,
   llb_buildsystem_command_result_skipped = 3,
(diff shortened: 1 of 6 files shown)


BenchR267:
@swift-ci please smoke test

products/libllbuild/include/llbuild/buildsystem.h
```diff
@@ -111,6 +111,7 @@ typedef enum LLBUILD_ENUM_ATTRIBUTES {
   llb_buildsystem_command_result_failed = 1,
   llb_buildsystem_command_result_cancelled = 2,
   llb_buildsystem_command_result_skipped = 3,
+  llb_buildsystem_command_result_fatal = 4,
```
jakepetroules: Can you please add a doc comment to this member explaining what it's used for? I know the others don't have them right now, but given that we have 5 values now, it would be good to start :)

BenchR267: Yep, makes sense. Especially since that's the one with a more special behavior

BenchR267:
Opened https://github.com/apple/swift-package-manager/pull/2135 to fix the build fail on CI.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

dmbryson:
I'm not sure I understand this change.  The commands were already being marked as failed when dependency processing failed.

ddunbar:
@BenchR267 what is the status of this?

BenchR267:
The problem still exists. I didn't have time to work on that and adding a test - I will close the PR for now. But as a short summary of the intention of this change: llbuild spawns internal commands for some external ones, for example the MakefileDepsParser for Swift compiler invocations.
If that parser fails (occured for me with file names that include a `\n`) the compiler invocation succeeds (and the whole build as well) but an error will be emitted. Maybe that's even the intended behavior but from my understanding the build shouldn't succeed if llbuild can't parse the dependencies of that invocation.

dmbryson:
Hmm, yeah, still want to see the test, because from my reading of the changes, the compiler invocation is already being marked as failed.

PR #488, opened 2019-06-06
Use new posix_spawn file actions API to set working dir in macOS 10.15

jakepetroules:


jakepetroules:
@swift-ci please smoke test

PR #489, opened 2019-06-09
Guard new posix_spawn_* APIs with the macOS 10.15 SDK macro

jakepetroules:


jakepetroules:
@swift-ci please smoke test

lib/Basic/Subprocess.cpp
```diff
@@ -87,7 +87,7 @@ static int posix_spawn_file_actions_addchdir(posix_spawn_file_actions_t * __rest
 #if HAVE_POSIX_SPAWN_CHDIR
   return ::posix_spawn_file_actions_addchdir_np(file_actions, path);
 #else
-#ifdef __APPLE__
+#if defined(__APPLE__) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 101500
   if (__builtin_available(macOS 10.15, *)) {
```
aciidb0mb3r: What is this call doing then?

jakepetroules: It attempts to build in a call to the API if we have the 10.15 SDK or later, and calls that API only when running on 10.15 or later.

* __MAC_OS_X_VERSION_MAX_ALLOWED == SDK version
* __MAC_OS_X_VERSION_MIN_REQUIRED == deployment target version
* __builtin_available == runtime OS version

PR #494, opened 2019-06-21
build: remove a hack which is no longer needed

compnerd:
This workaround was to force clients to link against the libraries.
This is no longer required.  Removing these enables building libllbuild
on Windows where ncurses is not available.


compnerd:
@swift-ci please test

aciidb0mb3r:
We can also get rid of `Package@swift-4.2.swift` with this.

jakepetroules:
@swift-ci please smoke test

PR #495, opened 2019-06-21
build: add missing dependencies

compnerd:
This fixes building CAPITests on Windows.  Unlike Unix, Windows does not
permit undefined references to symbols at link time.  Add the missing
link dependencies to ensure that we can resolve all symbols.  This
permits building CAPITests on Windows.


compnerd:
@swift-ci please smoke test

PR #496, opened 2019-06-22
build: correct install rules for Windows

compnerd:
Install the runtime components into the `bin` directory and the import
libraries into the `lib` directory.  Adjust the naming for the very
confusingly named libraries (with incorrect prefixes).


compnerd:
@swift-ci please smoke test

jakepetroules:
👍

PR #498, opened 2019-06-26
Handle SkippedCommand values in DirectoryContents

dmbryson:
Under some cancellation conditions, the directory value observed by
DirectoryContentsTask may be a SkippedCommand. Prior to this change, the
provideInput() handler was assuming that such values had output info and
would read a garbage value and crash (in release mode).

rdar://problem/50380532

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -880,6 +880,13 @@ class DirectoryContentsTask : public Task {
       return;
     }
 
+    // The input directory may be a 'mkdir' command, which can be cancelled or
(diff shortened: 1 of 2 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:
@jakepetroules after adding some more to the test and some additional consideration, I think that 'skipped' is the correct choice here.  Tasks that actually *read* the value of directory contents, namely the DirectoryTree-family, check it already (or do so now with this change).   Other tasks, such as ExternalCommand, may run and fail, but the correct behavior of checking the contents will still be preserved in following builds.

PR #500, opened 2019-06-27
build: add a dependency on the compiler

compnerd:
Rather than depending on the version output, depend on the actual
compiler.  This is traditionally how CMake ensures that compiler changes
are reflected in the build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,17 +202,6 @@ if(LLBUILD_ENABLE_ASSERTIONS)
   string(REPLACE "-DNDEBUG" "" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
 endif()
 
-# Add an always out-of-date target to get the Swift compiler version.
-if (SWIFTC_FOUND)
-  set(SWIFT_VERSION ${LLBUILD_OBJ_DIR}/swift-version)
-  add_custom_target(
-    swiftversion ALL
-    COMMAND ${CMAKE_SOURCE_DIR}/utils/write-swift-version ${SWIFTC_EXECUTABLE} ${SWIFT_VERSION}
-    BYPRODUCTS ${SWIFT_VERSION}
-    COMMENT "Checking Swift compiler version"
-  )
-endif()
-
 ###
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -202,17 +202,6 @@ if(LLBUILD_ENABLE_ASSERTIONS)
   string(REPLACE "-DNDEBUG" "" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
 endif()
 
-# Add an always out-of-date target to get the Swift compiler version.
-if (SWIFTC_FOUND)
-  set(SWIFT_VERSION ${LLBUILD_OBJ_DIR}/swift-version)
-  add_custom_target(
-    swiftversion ALL
-    COMMAND ${CMAKE_SOURCE_DIR}/utils/write-swift-version ${SWIFTC_EXECUTABLE} ${SWIFT_VERSION}
```
aciidb0mb3r: can kill the script as well

dmbryson:
This seems reasonable to me, but I wonder if there is a compelling reason to depend on just the version?  i.e. do we expect/care about the case where the build has touched the compiler, but the version has *not* actually changed?

aciidb0mb3r:
> i.e. do we expect/care about the case where the build has touched the compiler, but the version has not actually changed?

IMO that doesn't sound like a big problem. The bindings are just two files anyway.

dmbryson:
Sounds good to me.

dmbryson:
Separately, @compnerd  any way we can get the Windows Swift-CI to at least post comments on these PRs?

compnerd:
@dmbryson - that is being worked on by @shahmishal!  For now, you can request an explicit test with "please test Windows platform" I believe.

dmbryson:
Nice!

dmbryson:
@swift-ci please test Windows platform

shahmishal:
Swift CI trigger for Windows is only available on apple/swift repo.

aciidb0mb3r:
Taking over this here: https://github.com/apple/swift-llbuild/pull/502

PR #501, opened 2019-06-27
Cherry-pick 'Handle SkippedCommand values in DirectoryContents'

dmbryson:
Under some cancellation conditions, the directory value observed by
DirectoryContentsTask may be a SkippedCommand. Prior to this change, the
provideInput() handler was assuming that such values had output info and
would read a garbage value and crash (in release mode).

rdar://problem/50380532


dmbryson:
@swift-ci please smoke test

PR #502, opened 2019-06-28
build: add a dependency on the compiler

aciidb0mb3r:
Rather than depending on the version output, depend on the actual
compiler.


aciidb0mb3r:
@swift-ci smoke test

aciidb0mb3r:
Hm, this doesn't work properly either

aciidb0mb3r:
@compnerd we might need to just go with python

jakepetroules:
What part is not working? Do you have a build log?

aciidb0mb3r:
cmake fails to rebuild when I change swiftc to point to a different compiler.

gottesmm:
@aciidb0mb3r @compnerd What is the status of this? What are we waiting on?

aciidb0mb3r:
This didn't work but we had already added back dependency on the compiler.

PR #503, opened 2019-07-01
Added unit test target for Swift layer

BenchR267:
This adds a unit test target to the Xcode project and the Swift package.
It can be used for unit tests of the Swift layer and contains some basic
tests of the Swift interface already.

rdar://52447946

I added some rudimentary tests for the Swift interface and want to use that target to write tests for the Swift interface of the build database (instead of having an example target).

**Question**: Should we add `swift test` to the `utils/build-and-test` script?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -92,6 +92,10 @@ let package = Package(
             name: "llbuildNinjaTests",
             dependencies: ["llbuildNinja", "gtest"],
             path: "unittests/Ninja"),
+        .testTarget(
+            name: "llbuildSwiftTests",
+            dependencies: ["llbuildSwift"],
+            path: "unittests/Swift"),
         
         // MARK: GoogleTest
 
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -188,7 +188,7 @@ Swift conventions.
 
 * Examples of using `llbuild` are available under `examples/`.
 
-* There are two kinds of correctness tests include in the project:
+* There are three kinds of correctness tests include in the project:
 
   **LLVM-Style Functional Tests**
 
@@ -203,6 +203,11 @@ Swift conventions.
     These tests are located under `unittests/` and then layed out according to
     library. The tests are written using the
     [Googletest](https://code.google.com/p/googletest/) framework.
+
+  **SwiftPM Unit Tests**
+
+    These tests are located under `unittests/Swift`. The tests are written using
+    the XCTest framework. Run them by executing `swift test` in the root level.
 
   All of these tests are run by default (by `lit`) during the build.
 
--- a/unittests/Swift/BuildSystemBindingsTests.swift
+++ b/unittests/Swift/BuildSystemBindingsTests.swift
@@ -1,0 +1,142 @@
+//
+//  BuildSystemBindingsTests.swift
+//  llbuildSwiftTests
+//
+//  Copyright © 2019 Apple Inc. All rights reserved.
+//
+
+import XCTest
+
+// The Swift package has llbuildSwift as module
+#if SWIFT_PACKAGE
+import llbuild
+import llbuildSwift
+#else
+import llbuild
+#endif
+
+extension Command {
+  func with(name: String? = nil, shouldShowStatus: Bool? = nil, description: String? = nil, verboseDescription: String? = nil) -> Command {
+    return Command(name: name ?? self.name, shouldShowStatus: shouldShowStatus ?? self.shouldShowStatus, description: description ?? self.description, verboseDescription: verboseDescription ?? self.verboseDescription)
+  }
+}
+
+class BuildSystemBindingsTests: XCTestCase {
+  
+  func testCommand() {
+    let command = Command(name: "aName", shouldShowStatus: true, description: "a description", verboseDescription: "a verbose description")
+    XCTAssertEqual(command.name, "aName")
+    XCTAssertTrue(command.shouldShowStatus)
+    XCTAssertEqual(command.description, "a description")
+    XCTAssertEqual(command.verboseDescription, "a verbose description")
+    
+    XCTAssertEqual(command, command.with())
+    XCTAssertNotEqual(command, command.with(name: "foobar"))
+    XCTAssertNotEqual(command, command.with(shouldShowStatus: false))
+    XCTAssertNotEqual(command, command.with(description: "another description"))
+    XCTAssertNotEqual(command, command.with(verboseDescription: "another verbose description"))
+    
+    XCTAssertEqual(command.hashValue, command.with().hashValue)
+    XCTAssertNotEqual(command.hashValue, command.with(name: "foobar").hashValue)
+    XCTAssertNotEqual(command.hashValue, command.with(shouldShowStatus: false).hashValue)
+    XCTAssertNotEqual(command.hashValue, command.with(description: "another description").hashValue)
+    XCTAssertNotEqual(command.hashValue, command.with(verboseDescription: "another verbose description").hashValue)
+  }
+  
+  func testCommandMetrics() {
+    let metrics = CommandMetrics(utime: 1, stime: 2, maxRSS: 3)
+    XCTAssertEqual(metrics.utime, 1)
+    XCTAssertEqual(metrics.stime, 2)
+    XCTAssertEqual(metrics.maxRSS, 3)
+  }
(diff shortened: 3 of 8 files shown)


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

Package.swift
```diff
@@ -92,6 +92,10 @@ let package = Package(
             name: "llbuildNinjaTests",
             dependencies: ["llbuildNinja", "gtest"],
             path: "unittests/Ninja"),
+        .testTarget(
+            name: "llbuildSwiftTests",
+            dependencies: ["llbuildSwift"],
+            path: "unittests/Swift"),
```
BenchR267: Does it make more sense to put it outside of `unittests` to not confuse the gtest based tests?

BenchR267:
@swift-ci please smoke test

Package.swift
jakepetroules: Unless it's causing an issue, I think it's fine as you have it.

BenchR267:
@swift-ci please smoke test

Package.swift
BenchR267: No, that seems fine

BenchR267:
@swift-ci please smoke test

BenchR267:
Hm, it seem to be a problem that the Xcode project uses the name 'llbuild' for the Swift framework while the Swift package uses llbuildSwift… It worked locally using the `#if canImport(lbuildSwift)` check but on the CI it seems there's already a llbuildSwift which was build with an older compiler version.
I think we need to use the same names for the package and Xcode target.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

unittests/Swift/BuildSystemBindingsTests.swift
```diff
@@ -0,0 +1,158 @@
...
+@testable import llbuild
+#endif
+
+extension Command {
+  func with(name: String? = nil, shouldShowStatus: Bool? = nil, description: String? = nil, verboseDescription: String? = nil) -> Command {
+    return Command(name: name ?? self.name, shouldShowStatus: shouldShowStatus ?? self.shouldShowStatus, description: description ?? self.description, verboseDescription: verboseDescription ?? self.verboseDescription)
+  }
+}
+
+class BuildSystemBindingsTests: XCTestCase {
```
aciidb0mb3r: Should we run these tests on Linux as well?

BenchR267: We should! I'll generate a LinuxMain

BenchR267: Done

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

unittests/Swift/BuildSystemBindingsTests.swift
aciidb0mb3r: We don't use SwiftPM to build llbuild on SwiftCI so LinuxMain won't really help..

BenchR267: What are we then using? Or where do I need to add the tests?

Package.swift
ddunbar: I actually probably would like to move this, but let’s do it in a separate PR (I have other related things)

docs/development.rst
```diff
@@ -204,6 +204,11 @@ Swift conventions.
     library. The tests are written using the
     [Googletest](https://code.google.com/p/googletest/) framework.
 
+  **SwiftPM Unit Tests**
+
+    These tests are located under `unittests/Swift`. The tests are written using
+    the XCTest framework. Run them by executing `swift test` in the root level.
+
```
ddunbar: The line after this is false, and I will want to fix it. By doing so, we will then immediately get these tests running in CI. The set up is complex though so I propose we merge this and I can take a look at the lot integration.

We actually already have a bit of it with the perftests (which are XCTest), so what is like to do is rename perftests to XCTests, sink perftests down one level, and move these into an adjacent directory.

unittests/Swift/BuildSystemBindingsTests.swift
```diff
@@ -0,0 +1,158 @@
...
+//
+
+import XCTest
+
+// The Swift package has llbuildSwift as module
+#if SWIFT_PACKAGE
+import llbuild
+@testable import llbuildSwift
+#else
+@testable import llbuild
```
ddunbar: I would like to avoid use of @testable, it prevents testing in release mode.

unittests/Swift/BuildSystemBindingsTests.swift
```diff
@@ -0,0 +1,158 @@
...
+    XCTAssertEqual(result.metrics?.stime, 2)
+    XCTAssertEqual(result.metrics?.maxRSS, 3)
+    
+    var cExtendedResult = llb_buildsystem_command_extended_result_t(result: .succeeded, exit_status: 0, utime: 1, stime: 2, maxrss: 3, pid: 123)
+    let result2 = CommandExtendedResult(&cExtendedResult)
+    XCTAssertEqual(result2.result, .succeeded)
+    XCTAssertEqual(result2.exitStatus, 0)
+    XCTAssertEqual(result2.pid, 123)
+    XCTAssertEqual(result2.metrics?.utime, 1)
+    XCTAssertEqual(result2.metrics?.stime, 2)
```
ddunbar: FWIW, I think this level of testing detail is overkill (checking that an unit function preroperly knitted)

Package.swift
BenchR267: I have another PR ready which builds on top of this by adding tests for the new Swift layer for the build database. Should we merge that as well before doing any work that refactors the testing infrastructure?

docs/development.rst
BenchR267: But the perf tests are probably running through `xcodebuild` and not via `swift test`? Would be great to have it integrated in the lit tests! 👍

unittests/Swift/BuildSystemBindingsTests.swift
BenchR267: @testable is necessary here due to internal initializer of `BuildKey` and `CommandExtendedResult` with their C-layer counterparts. I guess you're right about the necessity of those wiring tests in general, so we could remove those and remove @testable with it as well.

unittests/Swift/BuildSystemBindingsTests.swift
dmbryson: Swift-CI directly runs the CMake build (with Ninja) and the Xcode project.

utils/build-and-test
```diff
@@ -15,3 +15,10 @@ mkdir -p "${BUILDROOT}"
 cd "${BUILDROOT}"
 cmake -G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ "${SRCROOT}"
 ninja test
+
+# run Swift package tests if `swift` is available
+if [ -x "$(command -v swift)" ]; then
+  swift test
+elif
+  echo 'Skipping Swift package tests because `swift` couldn\'t be found.'
+fi
```
dmbryson: Note that this script is not part of Swift-CI either. This exists to facilitate local, Docker-based testing of llbuild.

BenchR267: Ok, so it makes most sense to have it integrated in `lit` like @ddunbar wrote. Sounds like it still makes sense to keep that here?

dmbryson: Well, at least until integrated with the lit tests.  It will be redundant then, and would be good to avoid the double build.

unittests/Swift/BuildSystemBindingsTests.swift
BenchR267: I agree, there's not really something to break here in the future and removing it also makes it possible to avoid the @testable (see above).

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

PR #504, opened 2019-07-10
[LLVM] Disable use of LLVM's CommandLine library in shared code.

ddunbar:
- This library injects static constructors which are unsuitable for use in
   library code, and in particular are causing the CAPI unit tests to fail at
   runtime (because of multiple definitions).

 - <rdar://problem/52894543> llbuild CAPI tests aren't running (the google test binary fails with LLVM command line errors)

### DIFF
--- a/lib/llvm/Support/Debug.cpp
+++ b/lib/llvm/Support/Debug.cpp
@@ -78,7 +78,10 @@ void setCurrentDebugTypes(const char **Types, unsigned Count) {
 } // namespace llvm
 
 // All Debug.h functionality is a no-op in NDEBUG mode.
-#ifndef NDEBUG
+//
+// LLBUILD-ONLY: Debug.h macros are disabled for llbuild (they inject static
+// constructors into library code).
+#if !defined(NDEBUG) && defined(LLVM_CODE_DISABLED_FOR_LLBUILD)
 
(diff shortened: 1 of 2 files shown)


ddunbar:
@swift-ci please smoke test

lib/llvm/Support/Debug.cpp
```diff
@@ -78,7 +78,10 @@ void setCurrentDebugTypes(const char **Types, unsigned Count) {
 } // namespace llvm
 
 // All Debug.h functionality is a no-op in NDEBUG mode.
-#ifndef NDEBUG
+//
+// LLBUILD-ONLY: Debug.h macros are disabled for llbuild (they inject static
+// constructors into library code).
+#if !defined(NDEBUG) && defined(LLVM_CODE_DISABLED_FOR_LLBUILD)
```
dmbryson: `!defined(LLVM...` ?

ie. a debug build, but not for llbuild

dmbryson: Oh, I see, it’s expected to be undefined for our case. 

The double negatives are strong with this one.

ddunbar: Ya, its gross, but easy

benlangmuir:
There are more of these in `CommandLine.cpp`.

PR #505, opened 2019-07-10
Created Swift interface for build database

BenchR267:
This adds a Swift interface to build on top of the added C interface.
It is not used at the moment but can be used for data fetching about builds that ran before.

rdar://50298712

### DIFF
--- a/unittests/Swift/BuildDBBindingsTests.swift
+++ b/unittests/Swift/BuildDBBindingsTests.swift
@@ -1,0 +1,156 @@
+//
+//  BuildDBBindingsTests.swift
+//  llbuildSwiftTests
+//
+//  Copyright © 2019 Apple Inc. All rights reserved.
+//
+
+import XCTest
+
+// The Swift package has llbuildSwift as module
+#if SWIFT_PACKAGE
+import llbuild
+import llbuildSwift
+#else
+import llbuild
+#endif
+
+private func createExampleBuildDB(at path: String, file: StaticString = #file, line: UInt = #line) throws {
+  typealias Compute = ([String]) -> String
+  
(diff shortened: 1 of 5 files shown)


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

unittests/Swift/BuildDBBindingsTests.swift
```diff
@@ -0,0 +1,89 @@
...
+  var exampleBuildDBPath: String {
+    return URL(fileURLWithPath: #file)
+      .deletingLastPathComponent()
+      .appendingPathComponent("Resources", isDirectory: true)
+      .appendingPathComponent("build.db")
+      .path
+  }
+  
+  func exampleDB() throws -> BuildDB {
+    let newPath = "\(self.tmpDirectory!)/build.db"
```
dmbryson: I'm a little worried about the maintenance overhead of including an example db for this test.  As your own comments attest, our db versioning strategy currently is to merely delete incompatible versions and move on. We don't maintain compatibility with older schemas.

It would be preferable to actually generate the test build db programatically.

BenchR267: Ah, that's a good idea 👍

BenchR267: I added a function that generates the database by basically doing similar things as the Swift example in the project.

BenchR267:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #506, opened 2019-07-13
[Ninja] Rename BindingSet to Scope.

ddunbar:
- Post Ninja 1.6, a scope can contain more than just variable bindings.

 - NIFC.


ddunbar:
@swift-ci please smoke test

PR #508, opened 2019-07-17
[BuildEngine] Fix potential deadlock during cancel

dmbryson:
If a failure occurs updating the build DB during finished task
processing, we trigger synchronous cancellation of the build. This
currently triggers a deadlock, waiting for the task that failed to be
marked as finished. This change prevents that deadlock by decrementing
the outstanding task count prior to such cancellation.

rdar://problem/50203529

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -904,6 +904,13 @@ private:
               ruleInfo->keyID, ruleInfo->rule, ruleInfo->result, &error);
           if (!result) {
             delegate.error(error);
+
+            // Decrement our count of outstanding tasks. This must be done prior
+            // the subsequent synchronous cancellation, since it will otherwise
+            // deadlock waiting for the task we have already popped off the
+            // queue. rdar://problem/50203529
+            --numOutstandingUnfinishedTasks;
+
             cancelRemainingTasks();
             return false;
           }


dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
```diff
@@ -904,6 +904,13 @@ class BuildEngineImpl : public BuildDBDelegate {
               ruleInfo->keyID, ruleInfo->rule, ruleInfo->result, &error);
           if (!result) {
             delegate.error(error);
+
+            // Decrement our count of outstanding tasks. This must be done prior
+            // the subsequent synchronous cancellation, since it will otherwise
+            // deadlock waiting for the task we have already popped off the
+            // queue. rdar://problem/50203529
+            --numOutstandingUnfinishedTasks;
+
```
ddunbar: Doesn’t cancellation also do this? Won’t that lead to an accounting error?

dmbryson: It only does so for items in the `finishedTaskInfos` vector. Since we have already removed it at this point, that is what causes the deadlock.

ddunbar: Got it, thanks!

dmbryson:
Probably could interpose a wrapper class around the BuildDB that would cause that specific path to fail.

dmbryson:
To keep the great being the enemy of the good, going to merge this and follow up with the test later <rdar://problem/53994643>.

PR #512, opened 2019-07-24
Added nullability information to C database interface

BenchR267:


dmbryson:
@swift-ci please smoke test

BenchR267:
Error on Linux: `db.h:26:48: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'llb_database_open'` - is nullability not available there?
Error on macOS: `error: 'XCUIKeyboardKey' is not a member type of 'XCTest'` - this was discussed in the Slack channel already but there was no solution. Could switching from the legacy to the new build system on the bots help here?

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
Seems that `__attribute__((nonnull))` works on Linux 👍

products/libllbuild/include/llbuild/db.h
```diff
@@ -23,27 +23,27 @@ typedef const char* llb_database_key_type;
 typedef struct llb_database_t_ llb_database_t;
 
 /// Open the database that's saved at the given path by creating a llb_database_t instance. If the creation fails due to an error, nullptr will be returned.
-LLBUILD_EXPORT const llb_database_t* llb_database_open(char *path, uint32_t clientSchemaVersion, llb_data_t *error_out);
+LLBUILD_EXPORT const llb_database_t *llb_database_open(char *path, uint32_t clientSchemaVersion, llb_data_t *error_out)  __attribute__((nonnull));
```
jakepetroules: I'm guessing that placing this at the end of the function declaration causes the nullability to apply to the return type as well as all of the parameters? Didn't know about this particular form...

BenchR267: From the documentation it seems to only apply to the parameters and there's no policy about the return type 😕 It seems to use an implicitely unwrapped optional for the return value and there's no way to change that.

BenchR267:
@swift-ci please smoke test

products/libllbuild/include/llbuild/db.h
jakepetroules: Typically you'd use the nullability _type qualifiers_ (_Nullable, _Nonnull, _Null_unspecified), like so:

    LLBUILD_EXPORT const llb_database_t * _Nullable llb_database_open(chat * _Nonnull path, uint32_t clientSchemaVersion, llb_data_t * _Nonnull error_out);

The nullability _attributes_ are a bit different and are more limited (including apparently the inability to specify nullable explicitly, as you mentioned). The Clang documentation has more info on nullability here: https://clang.llvm.org/docs/AttributeReference.html#nullability-attributes

Please use the type qualifiers instead of the attributes as they are better integrated with the type system. Further, you can save a bunch of typing throughout this file by using scope-based nullability. Define some macros:

	#if __has_feature(assume_nonnull)
	#define LLBUILD_ASSUME_NONNULL_BEGIN _Pragma("clang assume_nonnull begin")
	#define LLBUILD_ASSUME_NONNULL_END   _Pragma("clang assume_nonnull end")
	#else
	#define LLBUILD_ASSUME_NONNULL_BEGIN
	#define LLBUILD_ASSUME_NONNULL_END
	#endif

And then use them like so:

    LLBUILD_ASSUME_NONNULL_BEGIN
    void foo(void *bar, void * _Nullable baz);
    LLBUILD_ASSUME_NONNULL_END

All pointers within that scope will be marked as nonnull unless otherwise explicitly annotated with one of the other nullability qualifiers.

jakepetroules: Ah, I see you already updated to use the type qualifiers. Looks good!

BenchR267:
@swift-ci please smoke test

BenchR267:
So `__attribute__((nonnull))` works *but* there's no option for nullable, just nonnull. That means that everything not tagged will result in an implicitly unwrapped optional. I'm now adding dummy definitions for `_Nullable` and `_Nonnull` if they don't exist since they always exist when the Swift compiler evaluates the header - just not when `cc` does. So this should work on both platforms equally.

products/libllbuild/include/llbuild/db.h
BenchR267: Yeah, I had an offline discussion with Jordan about how to use `_Nullable` here 👍 I'll add the assume nonnull macros if the CI is green (macOS will probably still fail though).

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

PR #515, opened 2019-07-30
[BuildEngine] Guard against concurrent engine use

dmbryson:
Suspected as potential cause of rdar://problem/53261527

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -12,6 +12,7 @@
 
 #include "llbuild/Core/BuildEngine.h"
 
+#include "llbuild/Basic/Defer.h"
 #include "llbuild/Basic/Tracing.h"
 #include "llbuild/Core/BuildDB.h"
 
@@ -80,6 +81,9 @@ class BuildEngineImpl : public BuildDBDelegate {
 
   /// Whether the build should be cancelled.
   std::atomic<bool> buildCancelled{ false };
+
+  /// Whether a build is currently running.
+  std::atomic<bool> buildRunning{ false };
   
   /// The queue of input requests to process.
   struct TaskInputRequest {
@@ -1398,6 +1402,16 @@ public:
   /// @{
 
   const ValueType& build(const KeyType& key) {
+    // Protect the engine against invalid concurrent use.
+    if (buildRunning.exchange(true)) {
+      delegate.error("build engine busy");
+      static ValueType emptyValue{};
+      return emptyValue;
+    }
+    llbuild_defer {
+      buildRunning = false;
+    };
+
     if (db) {
       std::string error;
       bool result = db->buildStarted(&error);
--- a/include/llbuild/Basic/Defer.h
+++ b/include/llbuild/Basic/Defer.h
@@ -1,0 +1,64 @@
+//===- Defer.h --------------------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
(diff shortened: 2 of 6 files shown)


dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
```diff
@@ -1398,6 +1402,14 @@ class BuildEngineImpl : public BuildDBDelegate {
   /// @{
 
   const ValueType& build(const KeyType& key) {
+    // Protect the engine against invalid concurrent use.
+    if (buildRunning.exchange(true)) {
+      delegate.error("build engine busy");
+      static ValueType emptyValue{};
+      return emptyValue;
+    }
+    scope_defer(buildRunning = false);
```
ddunbar: This feels like a lot of potentially surprising "infrastructure" for a single use. Are you sure we will leverage this complexity?

ddunbar: Also, would it not be possible to build an RAII based defer which takes a lambda? Does LLVM have something like that?

dmbryson: I was hoping llvm would have something like it, but came up empty.  Maybe I just missed it? 

The ScopeDeferImpl is an RAII that takes a lambda (or other callable).  The macro just wraps it for easier use. 

I think this meme/util will be useful elsewhere, but it is not strictly necessary.  The number of exits is small and unlikely to grow in the method, so we could just litter it around.

dmbryson: Maybe if I drop the `Impl` part of the name, it would be more clear that the class itself could be used directly and that the macro is just syntactic sugar.

dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
ddunbar: I’d probably do that, and drop the macro, to start with.

@DougGregor does Swift have something like this?

dmbryson: I did rename it already in the current version.  I like the macro, but can defer* on it when back at my keyboard. 

* pun totally intended

jakepetroules: > does Swift have something like this?

Yeah, `@autoclosure`

DougGregor: The Swift compiler's implementation is at https://github.com/apple/swift/blob/master/include/swift/Basic/Defer.h

dmbryson: Well that looks quite familiar.  I like the use of braces in the macro form.

ddunbar:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
ddunbar: I cede all my review feedback to follow Swift's precedent

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

dmbryson:
Given that the (<rdar://problem/53413474> llbuild macOS PR bot is broken) hasn't moved, I expect macOS to fail.  I have confirmed the those tests locally.

aciidb0mb3r:
Maybe we should disable the xcodebuild tests until we fix this?

dmbryson:

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
> This time for sure!

Linux C++ headers don't cross include as much as macOS, apparently.

PR #516, opened 2019-08-01
Don't optimize Swift code when building for Debug

BenchR267:
This makes debugging much easier (or even possible).


BenchR267:
Thought so as well but nope, it was on optimize for speed (`-O`)

PR #519, opened 2019-08-06
Added timing information to the build database

BenchR267:
This renames existing timestamp properties and functions to epoch
to better differentiate between the new timestamp types.
A new Clock class gives access to static methods for getting global
timestamps in s.
The timing information is stored in the SQLite database in the rule
results table. That defines a new schema version.

rdar://52892354

### DIFF
--- a/include/llbuild/Basic/Clock.h
+++ b/include/llbuild/Basic/Clock.h
@@ -1,0 +1,42 @@
+//===- Clock.h --------------------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLBUILD_BASIC_CLOCK_h
+#define LLBUILD_BASIC_CLOCK_h
+
+#include <algorithm>
+#include <chrono>
+
+namespace llbuild {
+namespace basic {
+
+class Clock {
+public:
+  /// A timestamp is the number of seconds since the clock's epoch time.
+  typedef double Timestamp;
+  
+  Clock() LLBUILD_DELETED_FUNCTION;
+  
+  /// Returns a global timestamp that represents the current time in seconds since a reference data.
+  /// *NOTE*: This function uses a monotonic clock, so don't compare between systems.
+  inline static Timestamp now() {
+    // steady_clock is monotonic
+    auto now = std::chrono::steady_clock::now();
+    std::chrono::duration<double> difference = std::chrono::duration_cast<std::chrono::seconds>(now.time_since_epoch());
+    return difference.count();
+  }
+};
+
+}
+}
+
+#endif
--- a/lib/BuildSystem/ExternalCommand.cpp
+++ b/lib/BuildSystem/ExternalCommand.cpp
@@ -322,7 +322,7 @@ ExternalCommand::computeCommandResult(BuildSystemCommandInterface& bsci) {
       // info, but need to refactor the command result to just store the node
       // subvalues instead.
       FileInfo info{};
-      info.size = bsci.getBuildEngine().getCurrentTimestamp();
+      info.size = bsci.getBuildEngine().getCurrentEpoch();
       outputInfos.push_back(info);
     } else if (node->isVirtual()) {
       outputInfos.push_back(FileInfo{});
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
@@ -77,7 +77,7 @@ class BuildEngineImpl : public BuildDBDelegate {
   std::unique_ptr<BuildEngineTrace> trace;
 
   /// The current build iteration, used to sequentially timestamp build results.
-  uint64_t currentTimestamp = 0;
+  Epoch currentEpoch = 0;
 
   /// Whether the build should be cancelled.
   std::atomic<bool> buildCancelled{ false };
@@ -199,7 +199,7 @@ class BuildEngineImpl : public BuildDBDelegate {
 
     bool isComplete(const BuildEngineImpl* engine) const {
       return state == StateKind::Complete &&
-        result.builtAt == engine->getCurrentTimestamp();
+        result.builtAt == engine->getCurrentEpoch();
     }
 
     void setComplete(const BuildEngineImpl* engine) {
@@ -211,7 +211,7 @@ class BuildEngineImpl : public BuildDBDelegate {
       // FIXME: This is a bit dubious, and wouldn't work anymore if we moved the
       // Result to being totally managed by the database. However, it is just a
       // matter of keeping an extra timestamp outside the Result to fix.
-      result.builtAt = engine->getCurrentTimestamp();
+      result.builtAt = engine->getCurrentEpoch();
     }
 
     void setCancelled() {
@@ -864,7 +864,7 @@ private:
 
         // The task was changed if was computed in the current iteration.
         if (trace) {
-          bool wasChanged = ruleInfo->result.computedAt == currentTimestamp;
+          bool wasChanged = ruleInfo->result.computedAt == currentEpoch;
           trace->finishedTask(taskInfo->task.get(), &ruleInfo->rule,
                               wasChanged);
         }
@@ -1313,8 +1313,8 @@ public:
     return &delegate;
   }
 
-  Timestamp getCurrentTimestamp() {
-    return currentTimestamp;
+  Epoch getCurrentEpoch() {
+    return currentEpoch;
   }
 
   // When changing the implementation of those, do also copy
@@ -1437,7 +1437,7 @@ public:
     // an additional mark. When a rule is demanded, if its builtAt index isn't
     // up-to-date then we lazily reset it to be Incomplete, \see demandRule()
     // and \see RuleInfo::isComplete().
-    ++currentTimestamp;
+    ++currentEpoch;
 
     if (trace)
       trace->buildStarted();
@@ -1451,7 +1451,7 @@ public:
     // FIXME: Is it correct to do this here, or earlier?
     if (db) {
       std::string error;
-      bool result = db->setCurrentIteration(currentTimestamp, &error);
+      bool result = db->setCurrentIteration(currentEpoch, &error);
       if (!result) {
         delegate.error(error);
         static ValueType emptyValue{};
@@ -1495,14 +1495,14 @@ public:
   
   bool attachDB(std::unique_ptr<BuildDB> database, std::string* error_out) {
     assert(!db && "invalid attachDB() call");
-    assert(currentTimestamp == 0 && "invalid attachDB() call");
+    assert(currentEpoch == 0 && "invalid attachDB() call");
     assert(ruleInfos.empty() && "invalid attachDB() call");
     db = std::move(database);
     db->attachDelegate(this);
 
     // Load our initial state from the database.
     bool success;
-    currentTimestamp = db->getCurrentIteration(&success, error_out);
+    currentEpoch = db->getCurrentEpoch(&success, error_out);
     return success;
   }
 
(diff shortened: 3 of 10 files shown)


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

ddunbar:
I'll leave review to @dmbryson, but have you compared the llbuild performance tests (in release mode) with and without this? Let's make sure we have a good understanding of potential perf impact.

include/llbuild/Core/Clock.h
```diff
@@ -0,0 +1,39 @@
...
+  
+  /// Returns a global timestamp that represents the current time.
+  /// Use only for calculations of relative durations using durationSince().
+  static Timestamp getTimeInMicroseconds();
+  
+  /// Returns the relative duration since the given timestamp.
+  static Duration durationSince(Timestamp other);
+};
+
+}}
```
ddunbar: nit: 
```
}
}
```
plz

include/llbuild/Core/Clock.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef LLBUILD_CORE_CLOCK_h
+#define LLBUILD_CORE_CLOCK_h
+
+#include <cstdint>
+
+namespace llbuild {
+namespace core {
+
+class Clock {
```
ddunbar: This should go in Basic, I think. I'm no expert on C++14+'s time functionality, but I wonder if we should just be using that directly?

lib/BuildSystem/ExternalCommand.cpp
```diff
@@ -322,7 +322,7 @@ ExternalCommand::computeCommandResult(BuildSystemCommandInterface& bsci) {
       // info, but need to refactor the command result to just store the node
       // subvalues instead.
       FileInfo info{};
-      info.size = bsci.getBuildEngine().getCurrentTimestamp();
+      info.size = bsci.getBuildEngine().getCurrentIteration();
```
ddunbar: I suggest we rename to "Epoch", which often has the connotation we want.

include/llbuild/Core/Clock.h
```diff
@@ -0,0 +1,39 @@
...
+public:
+  /// A timestamp is the number of microseconds since a reference date
+  typedef uint64_t Timestamp;
```
ddunbar: I would prefer we use Double, and make it seconds, unless there is a compelling reason not to. I always probably wouldn't typedef it, unless we are going to make a wrapper struct to enforce type safety. Defer to @dmbryson on this though.

lib/Core/BuildEngine.cpp
```diff
@@ -77,7 +77,10 @@ class BuildEngineImpl : public BuildDBDelegate {
   std::unique_ptr<BuildEngineTrace> trace;
 
   /// The current build iteration, used to sequentially timestamp build results.
-  uint64_t currentTimestamp = 0;
+  Iteration currentIteration = 0;
+  
+  /// The timestamp when the build started or 0 if not running currently
+  Clock::Timestamp buildStartedTimestamp = 0;
```
ddunbar: uint64 max probably a safer sentinel value, if we need a sentinel. If we switch to double, maybe we can use -1 and enforce negative values can never be output from Clock().

lib/Core/BuildEngine.cpp
```diff
@@ -199,7 +202,13 @@ class BuildEngineImpl : public BuildDBDelegate {
...
     bool isComplete(const BuildEngineImpl* engine) const {
       return state == StateKind::Complete &&
-        result.builtAt == engine->getCurrentTimestamp();
+        result.builtAt == engine->getCurrentIteration();
+    }
+    
+    void setComputing(const BuildEngineImpl* engine) {
+      assert(isInProgressWaiting());
+      state = StateKind::InProgressComputing;
+      result.relativeStart = Clock::durationSince(engine->buildStartedTimestamp);
```
ddunbar: Why record relative start, instead of just getting a timestamp here, and then subtracting? Basically we currently do:
```
relativeStart = now() - buildStart
...
duration = now() - (relativeStart + buildStart)
```
instead of
```
start = now()
duration = now() - start
```

ddunbar: Ah, I see, you want to store relative start in the database. I wasn't expecting that, but can see why it might be useful (definitely nice for visualization)

I think I would still prefer if we put the absolute timestamp in the ruleInfo, and then did the math when we set the result properties, I think.

include/llbuild/Core/Clock.h
```diff
@@ -0,0 +1,39 @@
...
+  
+  /// Duration is the number of microseconds since a defined event
+  typedef uint64_t Duration;
+  
+  /// Returns a global timestamp that represents the current time.
+  /// Use only for calculations of relative durations using durationSince().
+  static Timestamp getTimeInMicroseconds();
+  
+  /// Returns the relative duration since the given timestamp.
+  static Duration durationSince(Timestamp other);
```
ddunbar: This should be declared inline, so it is efficient w/o LTO.

lib/Core/BuildEngine.cpp
```diff
@@ -1437,10 +1446,12 @@ class BuildEngineImpl : public BuildDBDelegate {
     // an additional mark. When a rule is demanded, if its builtAt index isn't
     // up-to-date then we lazily reset it to be Incomplete, \see demandRule()
     // and \see RuleInfo::isComplete().
-    ++currentTimestamp;
+    ++currentIteration;
```
ddunbar: FWIW, I'd land this rename in a separate patch, to improve readability.

lib/Core/BuildEngine.cpp
```diff
@@ -1418,7 +1427,7 @@ class BuildEngineImpl : public BuildDBDelegate {
     llbuild_defer {
       buildRunning = false;
     };
-
+
```
ddunbar: Nit, unnecessary diff

include/llbuild/Core/Clock.h
dmbryson: I'm inclined to suggest we use the standard library time functionality.

include/llbuild/Core/Clock.h
dmbryson: Agreed. I think storing it as a double makes sense.

lib/Core/BuildEngine.cpp
dmbryson: I agree with keeping the absolute timestamp in the ruleinfo.   I do wonder if it would make sense to store the timestamps in the database too.  With them we can do analysis on the scheduling of the tasks, and it is easy to derive the duration from the two values.

include/llbuild/Core/Clock.h
ddunbar: Have you used it though? I seem to recall chatter that some people think it has serious problems

lib/Core/BuildEngine.cpp
ddunbar: I like that idea, it would also let us clearly see things like when certain items were rebuilt

lib/Core/BuildEngine.cpp
BenchR267: I changed `Timestamp` to be of type `std::chrono::time_point` which has a static `max()` method. I think that's best as a sentinel value. In the `durationSince()` method I'm now checking for positive durations and will add that to the documentation.

lib/Core/BuildEngine.cpp
BenchR267: I think that depends what the timing information is being used for. I imagine that we'll use it to find out more about the events that happened during a build. With that in mind, the relative start is the only important information, it doesn't matter when in time the whole build was executed.
But if you think it makes sense to have the actual timestamp in the database, we can do that too. But then we also need to keep in mind what happens if the date/time gets changed/synchronized and what that means for the values.

ddunbar: Since the duration is easily derived, if we are storing both start and stop I think there is no good reason to avoid storing the absolute values, since it is strictly more info. Even if it only helps with debugging, that still has non-negative value

BenchR267:
@swift-ci please smoke test

include/llbuild/Core/Clock.h
BenchR267: I changed the implementation to use it. It looks much better as it handles conversion between units much better (or it does - in contrast to the previous implementation 🙂).
However: I don't see how it handles clock changes like traveling between timezones or getting the clock synchronized during a build (which is probably more likely).
I'll try to find out if it's possible to get monotonic values.

lib/BuildSystem/ExternalCommand.cpp
BenchR267: Done and also separated into an individual commit 👍

include/llbuild/Core/Clock.h
BenchR267: I changed it to double and seconds. My intention was that double is potentially not precise enough and storage. But uint64 uses 64bits as well so that doesn't matter and precision should be at least the same (if not better) as having microsecond integer.

BenchR267: Actually - I didn't change that specific part to be double but the representation in the database. This timestamp (in Clock.h) is now a specific time instance which should only be used to calculate durations.

lib/Core/BuildEngine.cpp
BenchR267: Yep, that's true. I'll change it

lib/Core/BuildEngine.cpp
BenchR267: If using actual timestamps and not relative start and duration this property is not needed anymore.

include/llbuild/Core/Clock.h
BenchR267: Timestamp will be a double and means the number of seconds sind the clock's epoch time.

include/llbuild/Core/Clock.h
BenchR267: `steady_clock` is monotonic and therefore does not get changed. I'll use that

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
I think we need to clean the CI as I changed the database scheme while using the same number between builds./

include/llbuild/Basic/Clock.h
```diff
@@ -0,0 +1,47 @@
...
+//
+// Copyright (c) 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLBUILD_CORE_CLOCK_h
```
dmbryson: include guards still say CORE

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -728,20 +741,21 @@ class SQLiteBuildDB : public BuildDB {
     sqlite3_finalize(stmt);
 
     // Dump Keys
-    result = sqlite3_prepare_v2(db, "SELECT key_id, built_at, computed_at FROM rule_results;",
+    result = sqlite3_prepare_v2(db, "SELECT key_id, built_at, computed_at, duration FROM rule_results;",
```
dmbryson: duration doesn't exist anymore (start, end)

include/llbuild/Basic/Clock.h
```diff
@@ -0,0 +1,47 @@
...
+    // steady_clock is monotonic
+    auto now = std::chrono::steady_clock::now();
+    std::chrono::duration<double> difference = std::chrono::duration_cast<std::chrono::seconds>(now.time_since_epoch());
+    return difference.count();
+  }
+  
+  /// Returns the relative duration in seconds since the given timestamp.
+  inline static double durationSince(Timestamp other) {
+    return now() - other;
+  }
```
ddunbar: I think we can probably skip this function, and just inline the math (which is actually shorted, and self explanatory).

BenchR267: It‘s actually not used so far but I wanted to use this for the rule results API in my other PR. But I guess you‘re right that it‘s self explanatory.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
I think it's now good to go. What do you think?

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -728,20 +741,22 @@ class SQLiteBuildDB : public BuildDB {
...
-    result = sqlite3_prepare_v2(db, "SELECT key_id, built_at, computed_at FROM rule_results;",
+    result = sqlite3_prepare_v2(db, "SELECT key_id, built_at, computed_at, start, end FROM rule_results;",
                                 -1, &stmt, nullptr);
     if (result != SQLITE_OK)
       return;
 
     os << "\nresults:\n";
     while (sqlite3_step(stmt) == SQLITE_ROW) {
-      assert(sqlite3_column_count(stmt) == 3);
+      assert(sqlite3_column_count(stmt) == 4);
```
dmbryson: This should be 5 columns.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

PR #522, opened 2019-08-14
Make CAPIRulesVector move the repsonsibility of destroying keys

BenchR267:
CAPIRulesVector's only purpose is to provide a C friendly interface to its vector. The llb_build_key_t ** interface does move the responsibility of destroying the keys after usage but the class does destroy them itself in its destructor.
This change removes the destructor and adds a comment about the usage of the data() getter.

rdar://54298130

### DIFF
--- a/products/llbuildSwift/BuildKey.swift
+++ b/products/llbuildSwift/BuildKey.swift
@@ -53,8 +53,12 @@ public class BuildKey: CustomStringConvertible, Equatable, Hashable {
     /// The opaque pointer to the BuildKey object
     internal let internalBuildKey: OpaquePointer
     
-    fileprivate init(_ buildKey: OpaquePointer) {
+    /// Will only destroy the internalBuildKey on deinit if this is `false`.
+    private let weakPointer: Bool
+    
+    fileprivate init(_ buildKey: OpaquePointer, weakPointer: Bool = false) {
         self.internalBuildKey = buildKey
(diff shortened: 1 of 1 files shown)


BenchR267:
@swift-ci please smoke test

BenchR267:
All clients pass the build keys to the Swift interface which handles the destroy in their deinit. Another possibility is to copy all keys in `data()` which would be more expensive but this isn't a performance critical part so I think it would be okay as well.
From the usage of CAPIRulesVector it seems like moving the responsibility to the receiver of `data()` seems intended.

dmbryson:
Yeah, I really don't want to move the cleanup responsibility to the client in these callbacks.  That we are allocating objects to bridge the interface should be kept as an implementation detail.  I think clients should treat the items as owned by the engine and thus be responsible for copying them if they want to keep them past the lifetime of the callback.   I would expect that in most cases the client shouldn't need to retain them.  This also leaves the engine free to do something more efficient in the future, if we so choose.

My thought here is that clients of the C API should expect that they must destroy things that come back from a `make` or `create` call.  But if they are passed a value in a callback, they did not create it, thus don't own it.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

PR #524, opened 2019-08-15
Don't destroy a weak pointer

BenchR267:
The state of having a weak pointer in BuildKey was previously added but the check before destroying wasn't included.

rdar://54298130

I don't know how that slipped through, that's the essential part of the change in https://github.com/apple/swift-llbuild/pull/522.


BenchR267:
@swift-ci please smoke test

PR #525, opened 2019-08-15
Critical Build Path

BenchR267:
This pull request adds the implementation and a tool infrastructure for calculating the critical build path from a given database. It's using the newly introduced Swift interfaces of BuildDB, BuildKey, BuildValue and RuleResult + the timing information for calculating the most expensive (measured in seconds) path to build a key.
The solver itself uses dynamic programming and therefore caches results. To get the critical path of a whole build it needs to calculate the critical path for every build key that was build and then search for the most expensive one.
The algorithm is provided by a new target in llbuild and is provided as a library product by its Swift package. The tooling produces a binary which is using this package and offers a basic interface to visualize the critical path on the command line by printing all the keys in order in addition to the cost of that path.

rdar://50298961

**NOTE**: Marking as **[WIP]** to add the following to the implementation before merging:
- [x] add a format option to export the critical path as json (next to text)
- [x] collect the actual start and end timestamps of the critical path while building it up to get more information of the 'realistic' critical path - including the wait gaps which exist due to system limitations


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

ddunbar:
Given the size of the PR I would propose splitting it up. Step 1 would be define the new analysis tool. I would kind of like it if it was llbuild-analysis and we could eventually make ‘llbuild analysis foo’ work (by shelling out from the CXX llbuild tool.

Package.swift
```diff
@@ -45,6 +48,13 @@ let package = Package(
             path: "products/libllbuild"
         ),
         
+        /// Framework providing analysis tools for the build database
+        .target(
+            name: "llbuildAnalysis",
+            dependencies: ["llbuildSwift"],
+            path: "products/llbuildAnalysis"
```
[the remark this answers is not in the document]
BenchR267: I've seen llbuildAnalysis as building on top of llbuildSwift which itself is using the stuff in lib. So moving llbuildAnalysis to lib seems a bit strange from a layering perspective I think.
About the executable: I see this more as a proof of concept currently which is using llbuildAnalysis which provides access to the actual algorithms (or the one algorithm at the moment). `build-analysis` is more the binary which gives CLI level access.
But I'm fine to move that into the llbuild package if you prefer that.

products/llbuildAnalysis/CriticalPath.swift
```diff
@@ -0,0 +1,101 @@
...
+    
+    /// The key that was built
+    public let key: BuildKey
+    /// The result containing the value, timing and dependency information
+    public let result: RuleResult
+    /// The cost of the whole path (sum up durations of all keys downstream)
+    public let cost: Double
+    /// The realistic cost of a critical path includes the waiting times between
+    /// the individual tasks.
+    public let realisticCost: Double
```
ddunbar: I don’t know what this means, and haven’t grokked the code yet. I wonder if this isn’t something better handled elsewhere though

ddunbar: (For example in a simulation tool). Critical path has a clear mathematical definition; I don’t know if this does as well

unittests/BuildAnalysis/CriticalPathSolverTests.swift
```diff
@@ -0,0 +1,145 @@
...
+        func path(with names: [String]) throws -> CriticalPath? {
+            let expectedKeys = names.compactMap { keyByName[$0] }
+            let rules = expectedKeys.compactMap { lookup[$0] }
+            guard expectedKeys.count == names.count, rules.count == names.count else { return nil }
+            return try CriticalPath.construct(Array(zip(expectedKeys, rules)))
+        }
+    }
+    
+    typealias InputData = [(key: String, deps: [String], start: Double, end: Double)]
+    static let exampleData: InputData = [
```
ddunbar: Can this be inlined into the tests that use it?

unittests/BuildAnalysis/CriticalPathSolverTests.swift
```diff
@@ -0,0 +1,145 @@
...
+            return result
+        }
+        return Input(keys: keys, lookup: [BuildKey: RuleResult](zip(keys, ruleResults), uniquingKeysWith: {a, b in a}), keyByName: keyByName)
+    }
+    
+    func createSolver(input: Input) -> CriticalPathSolver {
+        return CriticalPathSolver(keys: input.keys, ruleResultLookup: { input.lookup[$0] })
+    }
+    
+    func testCriticalPath_emptyKeys() throws {
```
ddunbar: The class name indicates these are critical path tests, no need to duplicate that

unittests/SwiftBinding/BuildDBBindingsTests.swift
```diff
@@ -194,4 +194,17 @@ class BuildDBBindingsTests: XCTestCase {
...
+    let result = RuleResult(value: BuildValue.FailedCommand(), signature: 0xff, computedAt: 1, builtAt: 2, start: 3, end: 4, dependencies: deps)
+    XCTAssertEqual(result.value, BuildValue.FailedCommand())
+    XCTAssertEqual(result.signature, 0xff)
+    XCTAssertEqual(result.computedAt, 1)
+    XCTAssertEqual(result.builtAt, 2)
+    XCTAssertEqual(result.start, 3)
+    XCTAssertEqual(result.end, 4)
+    XCTAssertEqual(result.dependencies.count, 2)
+    XCTAssertEqual(result.dependencies, deps)
+  }
```
ddunbar: Should be in a separate PR

Package.swift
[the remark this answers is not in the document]
dmbryson: Just because it depends on other things in `lib/`, I don't think that should preclude it from being in `lib/` itself.  I agree with @ddunbar here that I think it makes sense to move this into the top level package and put the analysis bits into `lib/` and have the prototype tool built as `products/`.

products/build-analysis/Sources/Core/ArgumentsParser.swift
```diff
@@ -0,0 +1,144 @@
+// This source file is part of the Swift.org open source project
```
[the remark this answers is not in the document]
BenchR267: Yeah, me neither and I'm also not very happy with the implementation. Would have loved to use the one from SwiftPM - is there any process of moving that out to the tools support repo?
My plan was to create a working, but effortless implemented, solution and replace that as soon as we have something. That's why I create an extra target for it.

unittests/BuildAnalysis/CriticalPathSolverTests.swift
BenchR267: Ha, didn't realize there's actually just one test using that :)

products/llbuildAnalysis/CriticalPath.swift
BenchR267: Yeah, this is not really part of the mathematical calculation but a measurement that helps us to reason about the build. If the difference between this and the `cost` gets higher it means a more powerful machine would have built faster since tasks needed to wait although they were ready to go.

BenchR267: On the other hand - it's easily calculatable afterwards with the start and end properties. I just thought it's easier to do that while building the graph and saving on the O(n) _(per graph)_ operations afterwards.

Package.swift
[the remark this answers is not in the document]
ddunbar: @BenchR267 Is that because `llbuildSwift` is in products? Maybe it should move to `lib` as well.

products/build-analysis/Sources/Core/ArgumentsParser.swift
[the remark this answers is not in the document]
ddunbar: No one is currently doing the work necessary to get the `tools-support-core` repo up and working. It needs someone to drive pulling that code out of SwiftPM, and figuring out how to maintain SwiftPM's build / bootstrap process once it is factored out. It is very sorely needed, but unfortunately that alone doesn't make a driver. :)

products/llbuildAnalysis/CriticalPath.swift
ddunbar: I doubt that cost matters much in the scheme of things, I think clarity of the code is worth a lot more (see also comment about generic critical path algorithm)

Package.swift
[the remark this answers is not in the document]
BenchR267: Should it? My understanding was that llbuildSwift is basically the Swift facing library which is a product. I think my confusion comes from the differentiation between the Xcode project and the Swift package.

products/llbuildAnalysis/CriticalPath.swift
```diff
@@ -0,0 +1,101 @@
...
+import llbuild
+import llbuildSwift
+#else
+import llbuild
+#endif
+
+/// Critical Path is a single linked list containing the first task built and pointers to dependents
+/// where the current element was the most expensive (long running) dependency of `next`.
+/// It conforms to Sequence providing access to all `BuildKey`s and their results in order.
+public final class CriticalPath: CustomStringConvertible {
```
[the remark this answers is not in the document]
BenchR267: I think that makes a lot of sense and I'm currently working on that. Found a few issues in the implementation and especially big performance problems while doing that. I'll close this PR for now and will open a few smaller ones with iterative additions in the next days.

PR #527, opened 2019-08-16
libllbuild: make build succeed on Windows x86_64

compnerd:
Windows does not support VLAs.  The current pattern would fail to build
with a failure to construct the local array as the size was not
constant.  Switch to a reserved vector which avoids the growth but
ensures that we can allocate a variable sized array.  This repairs the
Windows build.


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test Linux platform

PR #529, opened 2019-08-16
build: sink processor checks into Utility

compnerd:
The processor count is used by the Swift build function.  Locate it near
it.  This will eventually go away as CMake 3.15 supports Swift
natively.


compnerd:
@swift-ci please smoke test

PR #530, opened 2019-08-16
build: re-install libllbuild.so on Linux

compnerd:
This is needed to be in the same directory.  Ensure that we install it.
Ideally, we would install this in the proper directory.


compnerd:
@swift-ci please smoke test

PR #531, opened 2019-08-17
build: re-install libllbuild.so on Linux

compnerd:
Install libllbuild as part of the llbuild build at a second location on
non-Windows targets: `lib/swift/pm/llbuild`.  This is needed for the
proper execution on Linux platforms.


compnerd:
@swift-ci please smoke test

compnerd:
@dmbryson - yeah.  This ensures that the target tracking occurs properly within CMake rather than trying to track the library naming and location.

PR #532, opened 2019-08-17
Simplify llbuild CMake configuation

compnerd:
A series of simplifications and cleanups to the llbuild build.  This takes care of low hanging fruit.  CMake 3.15+ will enable much more cleanup as well as the ability to export the targets to make it easier to build s-p-m.


compnerd:
@swift-ci please smoke test

PR #533, opened 2019-08-17
libllbuild: fix definition of `LLBUILD_EXPORT`

compnerd:
The Windows build requires more careful handling of the exports macro.
WHen building the library as a static library, no dll storage should be
specified.  When building the library dynamic, DLL storage must be
applied.  This corrects the exported interfaces from the executable
(`llbuild`) exporting the interfaces for `libllbuild`.  Unix would
prevent the exporting of the symbols as `-rdynamic` is not the default.


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #534, opened 2019-08-17
Basic: correct dependencies for linking on Windows

compnerd:
This corrects the link dependencies for Windows.  llbuildBasic is always
built as a static library.  This requires that the dependencies be
linked PUBLIC so that users of the library are able to have the
dependencies fulfilled.  llbuild now properly links in ShLwApi (the
dependency was previously silently fulfilled by the implicitly linked
libraries from CMake).


compnerd:
@swift-ci please smoke test

PR #537, opened 2019-08-19
Create compatibility to Xcode 10

BenchR267:
Xcode 10 ships with Swift 5.0 and doesn't know about implicit return or using Self in class context. The changes before required people to use Xcode 11 which isn't the officially released version.

I think we can revert this commit eventually to switch back using the cool 5.1 features once there's an official version available.


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

PR #538, opened 2019-08-19
Allow building iOS + CocoaPods

igor-makarov:
This PR makes small adjustments so that swift-llbuild can be embedded in iOS applications and other projects targeting Apple systems.

Changes:
* Add a dylib-only target, because currently, that's the only way to make Xcode link the library dynamically.
* Add missing import.
* Invert several platform conditionals so that they're targeting all Apple systems and not only `#if os(macOS)`.
* Add minimum versions to `Package.swift`.

### DIFF
--- a/include/llbuild/BuildSystem/BuildSystemFrontend.h
+++ b/include/llbuild/BuildSystem/BuildSystemFrontend.h
@@ -116,7 +116,7 @@ public:
   /// Handle used to communicate information about a launched process.
   struct ProcessHandle {
     /// Opaque ID.
-    uintptr_t id;
+    uint64_t id;
   };
   
 private:
--- a/include/llvm/Config/config.h
+++ b/include/llvm/Config/config.h
@@ -256,7 +256,10 @@
 #if defined(__APPLE__)
 #include "TargetConditionals.h"
 #endif
-#if !defined(__APPLE__) || !TARGET_OS_IPHONE
+
+#if defined(__APPLE__) && TARGET_OS_IPHONE
+#undef HAVE_TERMINFO
+#else
 #define HAVE_TERMINFO 1
 #endif
 
--- a/include/llvm/Config/config.h
+++ b/include/llvm/Config/config.h
@@ -253,7 +253,12 @@
 #define HAVE_SYS_TYPES_H 1
 
 /* Define if the setupterm() function is supported this platform. */
+#if defined(__APPLE__)
+#include "TargetConditionals.h"
+#endif
+#if !defined(__APPLE__) || !TARGET_OS_IPHONE
 #define HAVE_TERMINFO 1
+#endif
 
 /* Define if the xar_open() function is supported this platform. */
 #define HAVE_LIBXAR 1
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -44,7 +44,8 @@
 
 #ifdef __APPLE__
 #include <pthread/spawn.h>
-
+#include "TargetConditionals.h"
+#if !TARGET_OS_IPHONE
 extern "C" {
   // Provided by System.framework's libsystem_kernel interface
   extern int __pthread_chdir(const char *path);
@@ -64,7 +65,7 @@ int pthread_fchdir_np(int fd)
 {
   return __pthread_fchdir(fd);
 }
-
+#endif
 #endif
 
 #ifndef __GLIBC_PREREQ
@@ -624,7 +625,7 @@ void llbuild::basic::spawnProcess(
 
 #if !defined(_WIN32)
       if (usePosixSpawnChdirFallback) {
-#ifdef __APPLE__
+#if defined(__APPLE__) && !TARGET_OS_IPHONE
         thread_local std::string threadWorkingDir;
 
         if (workingDir.empty()) {
@@ -646,7 +647,7 @@ void llbuild::basic::spawnProcess(
           workingDirectoryUnsupported = true;
           result = -1;
         }
-#endif // ifdef __APPLE__
+#endif // if defined(__APPLE__) && !TARGET_OS_IPHONE
       }
 #endif // else !defined(_WIN32)
 
--- a/llbuild.podspec
+++ b/llbuild.podspec
@@ -23,7 +23,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target     = "10.0"
   s.osx.deployment_target     = "10.10"
 
-  s.source = { git: "https://github.com/apple/swift-llbuild",
+  s.source = { git: "https://github.com/apple/swift-llbuild.git",
                tag: s.version }
 
   s.default_subspecs = ['Swift']
--- a/llbuild.podspec
+++ b/llbuild.podspec
@@ -74,7 +74,7 @@ Pod::Spec.new do |s|
     sp.source_files = 'lib/BuildSystem/**/*.cpp'
     # internal header files, used this way to prevent header clash between subspecs
     sp.preserve_paths = 'include/llbuild/BuildSystem', 'lib/BuildSystem/**/*.h'
-    sp.compiler_flags = '-I${PODS_TARGET_SRCROOT}/include'
+    sp.compiler_flags = '-I${PODS_TARGET_SRCROOT}/include -Wno-error=c++11-narrowing'
 
     sp.dependency 'llbuild/Core'
   end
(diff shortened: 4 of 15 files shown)


products/llbuildSwift/BuildDBBindings.swift
```diff
@@ -8,13 +8,13 @@
 
 // This file contains Swift bindings for the llbuild C API.
 
-#if os(macOS)
```
jakepetroules: Can you please instead write this as `#if canImport(Darwin)`? The `else` case should remain `Glibc` because that's what applies to other platforms as well, like FreeBSD or Android, and this patch would break builds there.

Same for the other places.

ddunbar:
Cool! Are you able to share what you are using this for?

igor-makarov:
I'm not using it for anything specific at the moment, but I have a few ideas for it in the mobile space.

Tangentially, how do you feel about adding a podspec? I already have one ready.

igor-makarov:
@jakepetroules regarding the dylib target - I tried to get Xcode 11 to link the existing library as a dylib, but apparently that isn't an option unless the library is explicitly declared as `.dynamic`.

I thought that potential  users might want the option of dynamic linking without needing a wrapper.

ddunbar:
I’m totally fine adding a Podspec, ideally if you are willing to help maintain it

igor-makarov:
Cool, I'll add it to this PR!

products/llbuildSwift/BuildDBBindings.swift
igor-makarov: Done, even added a check for `#canImport(Glibc)` and a custom error when neither of those is available.

igor-makarov:
Added the podspec, and a script named `utils/push-podspec` to push a specific version.  
Because `llbuild` doesn't have its current version anywhere in the repo itself, this means that the script would have to have a version passed to it manually for pushing the podspec to CocoaPods registry. Inspired by [swift-nio](https://github.com/apple/swift-nio/blob/master/scripts/build_podspecs.sh)

igor-makarov:
Question - before I give up on iOS 9 - there's a `thread_local` in [Subprocess.cpp](https://github.com/apple/swift-llbuild/blob/master/lib/Basic/Subprocess.cpp#L628).

Can it be replaced by something that would compile on iOS 9? 

Alternatively, it seems like this entire part of the library should even be on iOS - can it be factored out?

igor-makarov:
Turns out there are 'private APIs' in llbuild: 

```
ITMS-90338: Non-public API usage - The app references non-public symbols in Frameworks/llbuild.framework/llbuild: ___pthread_chdir, ___pthread_fchdir, _del_curterm, _set_curterm, _setupterm, _tigetnum.
```

I'm thinking that all of this isn't actually needed for iOS, any ideas on how to exclude the offending code?

ddunbar:
@igor-makarov you can start by just following the Linux code path on non macOS (which effectively disables support for the feature that uses those APIs, but otherwise things still work)

ddunbar:
Can we just support > iOS 9?

dmbryson:
Seems iTMS is also complaining about our linking to libcurses.

ddunbar:
We should probably jettison that, its only necessary for the command line tools. On iOS, I would presume the main use case for llbuild is linking the pure library code. We should make sure we only link libcurses when necessary.

dmbryson:
Agreed

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -707,15 +709,15 @@ public final class BuildSystem {
         info.pointee.inode = UInt64(s.st_ino)
         info.pointee.mode = UInt64(s.st_mode)
         info.pointee.size = UInt64(s.st_size)
-        #if os(macOS)
```
jakepetroules: Please change this to `#if canImport(Darwin)` as well

llbuild.podspec
```diff
@@ -0,0 +1,88 @@
...
+                    Ninja manifests.
+                   DESC
+
+  s.homepage     = "https://github.com/apple/swift-llbuild"
+  s.license      = { type: 'Apache 2.0', file: "LICENSE.txt" }
+
+  s.documentation_url  = "https://llbuild.readthedocs.io/"
+  s.author             = "Apple"
+
+  s.ios.deployment_target     = "10.0"
```
jakepetroules: Please make sure the podspec and SPM deployment targets are the same

igor-makarov: After removing the code that ITMS was complaining about, minimum version can be 9.

include/llvm/Config/config.h
```diff
@@ -253,7 +253,12 @@
 #define HAVE_SYS_TYPES_H 1
 
 /* Define if the setupterm() function is supported this platform. */
+#if defined(__APPLE__)
```
igor-makarov: This compiles, but is this okay? @ddunbar @dmbryson

include/llvm/Config/config.h
```diff
@@ -253,7 +253,12 @@
...
+#include "TargetConditionals.h"
+#endif
+#if !defined(__APPLE__) || !TARGET_OS_IPHONE
 #define HAVE_TERMINFO 1
+#endif
```
igor-makarov: This compiles, but is this okay? @ddunbar @dmbryson

lib/Basic/Subprocess.cpp
```diff
@@ -64,7 +65,7 @@ int pthread_fchdir_np(int fd)
 {
   return __pthread_fchdir(fd);
 }
-
+#endif
```
igor-makarov: This compiles, but is this okay? @ddunbar @dmbryson

products/llbuildSwift/BuildSystemBindings.swift
igor-makarov: Done.

igor-makarov:
Ok, so this has turned into a big yak shave, but I think I'm done.

Final changelog:
* Added podspec
* Exclude the BuildSystem API from the public header and Swift bridge when on iOS
* Lowest iOS version - 9.0 - the previously incompatible code was in BuildSystem
* `ProcessHandle` opaque id changed from `uintprt_t` to `uin64_t` - the header is not exported in `libllbuild` and no warnings occured
* `#undef HAVE_TERMINFO` for iOS - disables linking to `libcurses`
* Disable 'chdir' functionality on iOS - was private API linkage
* Turn all the `#if os(macOS)` to `#if canImport(Darwin)`and invert the conditionals

The podspec has additional optimizations that couldn't be done for SwiftPM due to `Package.swift` syntax limitations:
* Exclude all the BuildSystem code completely - only Core and DB are supported - makes sense since iOS apps cannot spawn processes or run CLI tools - less compiled code that ultimately isn't used
* Exclude several large source files that were affecting final binary size and were only ultimately used in BuildSystem

include/llbuild/BuildSystem/BuildSystemFrontend.h
```diff
@@ -116,7 +116,7 @@ class BuildSystemFrontendDelegate : public BuildSystemDelegate {
   /// Handle used to communicate information about a launched process.
   struct ProcessHandle {
     /// Opaque ID.
-    uintptr_t id;
+    uint64_t id;
```
dmbryson: I am personally fine with this, as I’m pretty sure that is what I originally used.  However, I think this was changed alongside some of the efforts to build out Windows compatibility.  We should be sure that this doesn’t break that.  cc @compnerd

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

igor-makarov:
I think I've found a problem in the podspec, checking.

igor-makarov:
Regarding the `ProcessHandle` concerns - I've located the commit where this was introduced:  
https://github.com/apple/swift-llbuild/commit/912b0df423ca42ea1ac4b2384df8940617c6c354

It wasn't changed from `uint64_t`. Do we have Windows tests?

dmbryson:
Ah, well then. I don’t have any objections.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

igor-makarov:
@dmbryson I meant I don't know whether making it a `uint64_t` would work okay on Windows because I don't know how much is a Windows `HANDLE`.

igor-makarov:
I can confirm the podspec is fixed, though. I'm good to merge.

PR #545, opened 2019-08-25
Infer podspec version from git command

igor-makarov:
This PR aims to provide a better solution to 'current version' inference for `llbuild` when used in CocoaPods directly from git. 

When requiring a Pod directly from git, like so:
```
pod 'llbuild', git: 'https://github.com/apple/swift-llbuild.git'
```
CocoaPods relies on the `.podspec` to provide the version that is then used, among other things, for dependency resolution.  
SwiftPM, on the other hand, relies solely on SemVer and git tags and implies no versioning in the code itself.  

The two ways are ultimately incompatible, but this new approach attempts to bridge them. The main use cases:

* When requiring a pod directly from git, the version inferred will be the latest SemVer-compatible tag.
* When requiring a pod from a directory, it's implied that the directory is a git repo. The same logic applies.
* When pushing a podspec to Trunk with `pod trunk push`, the validation will be performed for the most recent tag. If it was already pushed, an error will be raised.
* For a podspec that is uploaded to Trunk, the logic will not be performed - Trunk podspecs are static JSON files with immutable versions.


igor-makarov:
It finds the most recent tag on the current branch. I've tested this on several different scenarios and it seemed to work okay.

PR #548, opened 2019-09-04
[examples] Update GameOfLife to be console only.

ddunbar:
- The HTTP library is no longer maintained, so for now just change this example
   to print to the console.


ddunbar:
Skipping testing, examples aren't covered. I manually tested.

PR #550, opened 2019-09-04
Avoid xcrun Call on Non Darwin

gmittert:
xcrun doesn't exist on non Darwin, so we should skip the execute_process
that calls it on non Darwin.


gmittert:
@swift-ci please test

gmittert:
@swift-ci please smoke test

PR #551, opened 2019-09-05
Add warning for sign-compare when building in Xcode

BenchR267:
The warning for sign compare is already turned on in CMake but not in the Xcode configuration.
That way a comparison with a signed and unsigned number slipped through which is fixed as well.


BenchR267:
@swift-ci please smoke test

PR #552, opened 2019-09-06
[Ninja] Properly clean up BuildContext sig thread

dmbryson:
rdar://problem/55036265


dmbryson:
@swift-ci please smoke test

PR #553, opened 2019-09-07
Abstract out default shell path and add Android LLVM config

buttaface:
Using this pull and a couple other small tweaks ([like this one](https://github.com/termux/termux-packages/blob/master/packages/llbuild/lib-Basic-PlatformUtility.cpp.patch)), I got all but one of the tests (BuildSystemCAPI.CustomToolWithDiscoveredDependencies) to pass when building llbuild and running it natively on Android in [the Termux app](https://github.com/termux/termux-app).

Abstracting out the shell path is particularly useful on Android as the system shell is in /system/bin/sh and the local shell at /data/data/com.termux/files/usr/bin/sh, so setting the path in only two places like this makes it much easier to patch.

### DIFF
--- a/include/llbuild/Basic/ExecutionQueue.h
+++ b/include/llbuild/Basic/ExecutionQueue.h
@@ -26,6 +26,9 @@ namespace llbuild {
     
     class ExecutionQueueDelegate;
 
+    // Default shell system path
+    const std::string DefaultShellPath = "/bin/sh";
+
     /// Description of the queue job, used for scheduling and diagnostics.
     class JobDescriptor {
     public:
--- a/include/llvm/Config/config.h
+++ b/include/llvm/Config/config.h
@@ -14,7 +14,9 @@
 #define ENABLE_CRASH_OVERRIDES 1
 
 /* Define to 1 if you have the `backtrace' function. */
+#if !defined(__ANDROID__)
 #define HAVE_BACKTRACE TRUE
+#endif
 
 #define BACKTRACE_HEADER <execinfo.h>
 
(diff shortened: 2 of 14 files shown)


include/llbuild/Basic/ExecutionQueue.h
```diff
@@ -26,6 +26,9 @@ namespace llbuild {
     
     class ExecutionQueueDelegate;
 
+    // Default shell system path
+    const std::string ShellPath = "/bin/sh";
```
buttaface: No particular reason to define this here, other than this header is already directly included by three of the four files where this variable's needed. I can move it elsewhere if preferred.

include/llvm/Config/config.h
```diff
@@ -104,7 +111,10 @@
 #define HAVE_LIBPTHREAD 1
 
 /* Define to 1 if you have the `pthread_getname_np' function. */
+#if defined(__ANDROID__)
+#else
 #define HAVE_PTHREAD_GETNAME_NP 1
+#endif
```
buttaface: Added in Android 8 as long as _GNU_SOURCE is defined, but for backwards compatibility commented out.

ddunbar:
That is awesome!

include/llbuild/Basic/ExecutionQueue.h
ddunbar: Seems ok to start. Maybe DefaultShellPath?

ddunbar:
@swift-ci please smoke test

include/llbuild/Basic/ExecutionQueue.h
buttaface: Done.

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #554, opened 2019-09-09
Create a basic tooling infrastructure for build analysis tools.

BenchR267:
This infrastructure does only contain a dummy command which prints out hello world but will contain more in following work.

rdar://54550630

There's no reference in the Xcode project to this tool since it can't specify the dependency on swift-tools-support-core. So building the tool is only supported via SwiftPM.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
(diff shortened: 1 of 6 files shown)


BenchR267:
@swift-ci please smoke test

PR #555, opened 2019-09-09
Remove dependency on tools support core

BenchR267:
Having the dependency on TSC in llbuild's Swift package creates resolution problems for packages depending on SwiftPM because the targets have the same name.
This is a temporary solution until we find a long term solution for this problem.

SR-11437


BenchR267:
@swift-ci please smoke test

PR #556, opened 2019-09-10
Specify a local package for llbuild-analyze

BenchR267:
llbuild-analyze depends on tools support core and can therefore not be specified in the Package.swift of llbuild itself because SwiftPM depends on llbuild which defines targets with the same name.
To solve resolution errors in projects depending on SwiftPM we declare a local package for llbuild-analyze. Once the TSC targets in SwiftPM has been removed and are owned by tools support core we can merge the packages for llbuild and llbuild-analyze back.

SR-11437


BenchR267:
@swift-ci please smoke test

PR #559, opened 2019-09-12
Finalize all sql statements when database gets closed

BenchR267:
Without finalizing the sqlite database doesn't get closed accordingly.

rdar://55274300

### DIFF
--- a/lib/Core/SQLiteBuildDB.cpp
+++ b/lib/Core/SQLiteBuildDB.cpp
@@ -169,6 +169,7 @@ class SQLiteBuildDB : public BuildDB {
         clientVersion != clientSchemaVersion) {
       // Close the database before we try to recreate it.
       sqlite3_close(db);
+      db = nullptr;
       
       if (!recreateOnUnmatchedVersion) {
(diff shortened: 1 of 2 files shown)


BenchR267:
@swift-ci please smoke test

BenchR267:
I don't see a way to actually test for that because there is not method for checking if there're still not finalized prepared statements.
But `sqlite3_close` returns `SQLITE_BUSY` if resources are not closed, finalized, finished accordingly. I added an assert for that since it's a development error and a test which opens and closes a connection and should hit this assert if that happens again.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

ddunbar:
That’s exactly what I had in mind, thanks!

ddunbar:
@BenchR267 actually, just noticed: the close shouldn’t be inside the assert, no assert builds will not close the DB! Can you fix?

BenchR267:
Oh true! Yes, I'll fix

PR #560, opened 2019-09-12
Fix memory leak when copying strings

BenchR267:
StringRef doesn't deallocate it's char pointer so that needs to be done after the external command has been run.
Also return the actual exit status of the external command.

rdar://55195607


BenchR267:
Yeah, let's fix this for all commands in a different PR. I think the `const char **` then shouldn't be converted to `std::string` at all but directly to `StringRef`.

PR #564, opened 2019-09-19
Use local variable if we're building without asserts

BenchR267:
This otherwise triggers a warning of an unused variable that gets written to.

rdar://55447092


BenchR267:
@swift-ci please smoke test

PR #566, opened 2019-09-27
Absolutize node names

vlm:
The `clang` generates either absolute or relative dependencies (in `.d` files) depending on its flags. This makes `llbuild ninja build` miss dependencies sometimes since from `llbuild ninja`'s standpoint the relative path name is different from absolute (this matches OS `ninja` behavior).

This patch resolves this edge case by absolutizing the node names (dependencies and rules).

An unfortunate consequence of this patch is that the debug output from the `llbuild ninja build` tool may sometimes differ: the absolute pathnames could be displayed instead of relative ones.

rdar://55235219

### DIFF
--- a/lib/Commands/NinjaBuildCommand.cpp
+++ b/lib/Commands/NinjaBuildCommand.cpp
@@ -30,6 +30,8 @@
 #include "llbuild/Ninja/ManifestLoader.h"
 
 #include "llvm/ADT/SmallString.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
 
 #include "CommandLineStatusOutput.h"
@@ -336,6 +338,7 @@ public:
 };
 
 struct NinjaBuildEngineDelegate : public core::BuildEngineDelegate {
+  std::string workingDirectory;
   class BuildContext* context = nullptr;
 
   virtual core::Rule lookupRule(const core::KeyType& key) override;
@@ -343,11 +346,16 @@ struct NinjaBuildEngineDelegate : public core::BuildEngineDelegate {
   virtual void cycleDetected(const std::vector<core::Rule*>& items) override;
 
   virtual void error(const Twine& message) override;
+
+  NinjaBuildEngineDelegate(StringRef workingDirectory)
+    : workingDirectory(workingDirectory) { }
 };
 
 /// Wrapper for information used during a single build.
 class BuildContext : public ExecutionQueueDelegate {
 public:
+  const std::string workingDirectory;
+
   /// The build engine delegate.
   NinjaBuildEngineDelegate delegate;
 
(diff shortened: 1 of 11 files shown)


dmbryson:
@swift-ci please smoke test

lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -1204,33 +1214,47 @@ buildCommand(BuildContext& context, ninja::Command* command) {
...
 
           virtual void actOnRuleDependency(const char* dependency,
                                            uint64_t length,
-                                           const StringRef
-                                             unescapedWord) override {
-            context.engine.taskDiscoveredDependency(task, unescapedWord);
+                                           const StringRef unescapedWord) override {
+
+            StringRef path;
+            SmallString<256> absPathTmp;
```
dmbryson: Should this be the equivalent of  `PATH_MAX`?

lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -1774,14 +1798,22 @@ int commands::executeNinjaBuildCommand(std::vector<std::string> args) {
     }
   }
 
+  SmallString<128> current_dir;
```
dmbryson: same as above, `PATH_MAX`?

lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -1946,7 +1975,9 @@ int commands::executeNinjaBuildCommand(std::vector<std::string> args) {
 
     // If this is the first iteration, build the manifest, unless disabled.
     if (autoRegenerateManifest && iteration == 0) {
-      context.engine.build(manifestFilename);
+      SmallString<256> absManifestFilename = StringRef(manifestFilename);
```
dmbryson: ditto

lib/Commands/NinjaCommand.cpp
```diff
@@ -412,8 +413,16 @@ static int executeLoadManifestCommand(const std::vector<std::string>& args,
     filename = filename.substr(pos+1);
   }
 
+  SmallString<128> current_dir;
```
dmbryson: ditto

lib/Ninja/Manifest.cpp
```diff
@@ -1,4 +1,4 @@
-//===-- ManifestLoader.cpp ------------------------------------------------===//
+//===-- Manifest.cpp ------------------------------------------------------===//
```
dmbryson: 👍

lib/Commands/NinjaBuildCommand.cpp
ddunbar: Lev & I already discussed this, paths can exceed path max so I think 256 is fine above. This one here isn't perf sensitive so I think we are ok as is.... if we wanted to clean it up later we should move these constants out to a shared file, but not germane to this PR.

tests/Ninja/Build/absolute-dependencies.ninja
```diff
@@ -0,0 +1,31 @@
+# Check that touching absolute transitive dependencies affect relative targets.
```
ddunbar: We should probably include a note *somewhere* that we are striving to be "more correct" than the actual Ninja, in this regard, since it is otherwise a little confusing if anyone ever wonders why.

lib/Commands/NinjaBuildCommand.cpp
vlm: I don't think so:

1. PATH_MAX is 1k on macOS, which is **not enough** to fully contain the maximum path name on APFS and other file systems.
2. This thing gets cleared in the beginning, which is O(N) if we're not extremely lucky.

vlm:
@swift-ci please smoke test

dmbryson:

PR #568, opened 2019-10-01
build: prepare for Foundation conversion to CMake 3.15

compnerd:
Stage the path structure to match the output of Foundation's results
after the migration to 3.15.1.  This is migration work, with the
canonical layout changing over to this layout.  This will be cleaned up
in a subsequent PR.


dmbryson:
@swift-ci please smoke test

PR #570, opened 2019-10-01
Added support for escaped CRLF newlines in Makefile Parser

gmittert:


gmittert:
@swift-ci please smoke test

lib/Core/MakefileDepsParser.cpp
```diff
@@ -70,6 +70,12 @@ static void skipNonNewlineWhitespace(const char*& cur, const char* end) {
       continue;
     }
 
+    // Also skip \r\n newlines
+    if (c == '\\' && cur + 2 != end && cur[1] == '\r' && cur[2] == '\n') {
```
dmbryson: I think that needs to be something like `cur + 2 < end`.  If we have only 1 byte left, the +2 would miss that we're going to read past the end.

dmbryson: One way to show that would be to add a test for `input = "out: \\\r";`

gmittert: Good point, fixed.

gmittert:
@swift-ci please smoke test

gmittert:
@swift-ci please smoke test

PR #572, opened 2019-10-03
Shim llvm::sys::getDefaultTargetTriple() to Swift

brentdax:
The default target triple for the host machine can be useful information when constructing command line arguments for a tool, but this information isn’t in LLVM’s C API, so it’s inaccessible from Swift. This commit exposes it in libllbuild as `llb_get_default_target_triple()` and in llbuildSwift as `BuildSystem.defaultTargetTriple`.

### DIFF
--- a/products/libllbuild/C-API.cpp
+++ b/products/libllbuild/C-API.cpp
@@ -14,6 +14,7 @@
 #include <llbuild/llbuild.h>
 
 #include "llbuild/Basic/Version.h"
+#include "llvm/Support/Host.h"
 
 using namespace llbuild;
 
@@ -28,4 +29,11 @@ const char* llb_get_full_version_string(void) {
 
 int llb_get_api_version(void) {
     return LLBUILD_C_API_VERSION;
+}
+
+const char * llb_get_default_target_triple() {
+  // Use a static local to store the triple, to avoid lifetime issues.
+  static std::string tripleString = llvm::sys::getDefaultTargetTriple();
+
+  return tripleString.c_str();
 }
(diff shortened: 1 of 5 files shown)


brentdax:
@swift-ci please smoke test

products/libllbuild/C-API.cpp
```diff
@@ -29,3 +30,10 @@ const char* llb_get_full_version_string(void) {
 int llb_get_api_version(void) {
     return LLBUILD_C_API_VERSION;
 }
+
+const char * llb_get_default_target_triple() {
+  // Use a static local to store the triple, to avoid lifetime issues.
+  static std::string tripleString = llvm::sys::getDefaultTargetTriple();
```
aciidb0mb3r: does this really work? LLVM goes to some extreme lengths to determine the default triple. I am not sure if llbuild does that as well.

brentdax: I'm not quite sure what you mean. This ends up using the exact same code that would be called in the clang or Swift drivers. That code does sometimes invoke external tools or perform other heroics to determine the OS version, but all of that code looks like it's intended to be idempotent, so caching the result ought not to cause any problems.

aciidb0mb3r: I meant if the copy in llbuild actually does the heroics done by llvm. I checked and it seems like this is filled in at compile time using `LLVM_DEFAULT_TARGET_TRIPLE` defined to empty string in llbuild: https://github.com/apple/swift-llbuild/blob/master/include/llvm/Config/config.h#L356

brentdax: Damn, you're right. Is there a particular reason we do this, or is it just because we haven't wanted this until now?

brentdax:
Looks like this won't work without a larger change to the way we configure llbuild.

PR #573, opened 2019-10-03
Simulate ninja -t clean by removing the database.

vlm:
There's a belief that it is quite hard (if not impossible) to make a _correct_ `-t clean` simulation. The next best thing to at least avoid breaking scripts that request `ninja -t clean` is to remove the build database. That's what we do in this patch.

rdar://20224847

### DIFF
--- a/lib/Commands/NinjaBuildCommand.cpp
+++ b/lib/Commands/NinjaBuildCommand.cpp
@@ -1774,6 +1774,7 @@ int commands::executeNinjaBuildCommand(std::vector<std::string> args) {
 
   if (!customTool.empty()) {
     std::vector<std::string> availableTools = {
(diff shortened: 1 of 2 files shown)


lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -1869,6 +1870,18 @@ int commands::executeNinjaBuildCommand(std::vector<std::string> args) {
       return 0;
     }
 
+    // Emulate `-t clean` by removing the database.
+    if(!customTool.empty() && customTool == "clean") {
```
ddunbar: Nit: `if (`

vlm:
@swift-ci please smoke test

PR #575, opened 2019-10-05
Strip ANSI colors unless CLICOLOR_FORCE is in effect

vlm:
rdar://51822760

### DIFF
--- a/lib/Commands/CommandLineStatusOutput.cpp
+++ b/lib/Commands/CommandLineStatusOutput.cpp
@@ -23,6 +23,9 @@ struct CommandLineStatusOutputImpl {
   /// The output stream.
   FILE* fp{nullptr};
 
+  /// Whether the output device supports ANSI color.
+  bool stripColor{true};
+
   /// Whether the output stream honors '\r'.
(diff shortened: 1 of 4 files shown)


vlm:
@swift-ci please smoke test

vlm:
@swift-ci please smoke test

vlm:
@swift-ci please smoke test

vlm:
@swift-ci please smoke test

vlm:
The fix https://github.com/apple/swift-llbuild/pull/575/commits/aa08154e08ce4dd2c6ded783c2c4d5e4d38c43ab suggests that the CI uses `dash` for a shell (`/bin/sh`), since `dash` is strictly POSIX-compliant and can't use `\x1b` in the `echo` builtin.

dmbryson:
@swift-ci please smoke test

vlm:
@ddunbar doesn't think like `printf` is any better [0]: POSIX doesn't let it use hexadecimals. In any case, the hex vs oct choice is not at all important in this patch.

 * [0] https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html

PR #577, opened 2019-10-07
Modify test so it updates the timestamp on FSs with 1-second resolution

vlm:
Temporarily modify the test until a proper fix is applied.

rdar://56045388


vlm:
@swift-ci please smoke test

PR #578, opened 2019-10-09
Remove "'const' type qualifier on return type has no effect" warning (-W, -Wall)

vlm:
* Remove superfluous const qualifier warning.
 * Almost allows running with `-W -Wall -Werror` error-free.


dmbryson:
@swift-ci please smoke test

PR #582, opened 2019-10-11
[CMake] Export libllbuild target

aciidb0mb3r:
This will allow other projects to easily link against libllbuild.


aciidb0mb3r:
@swift-ci smoke test

PR #588, opened 2019-10-15
Update to a newer lit API

vlm:
See https://reviews.llvm.org/D69002


vlm:
@swift-ci please smoke test

PR #589, opened 2019-10-15
Add adjust-times as a test target dependency

vlm:
* Tests are invoked by `ninja test`, skipping the default build. The `adjust-times` utility has to be a dependency of the test target.


vlm:
@swift-ci please smoke test

PR #590, opened 2019-10-16
[Xcode] Add a missing dependency on `adjust-times`.

ddunbar:
- rdar://56270635 (DerivedData/Build/Products/Debug/adjust-times: No such file or directory)


ddunbar:
Bypassing merge, this code path isn't tested.

PR #593, opened 2019-10-29
Fix Path Normalization on Windows

gmittert:
Absolute paths on Windows do not begin with a leading slash. Defer to
llvm's is_absolute instead which handles cross platform paths.


gmittert:
cc @vlm 

@swift-ci please test

gmittert:
@swift-ci please smoke test

PR #596, opened 2019-10-31
Streamline file descriptor handling during spawning

vlm:
* Leak of the output pipe file descriptors if control pipe can't be opened.
  * Leak of posix file actions memory if the control pipes can't be created.
  * Descriptors leak of unrelated tasks into spawned task (linux, win).
  * Windows code improperly inheriting the control file descriptors.
  * Windows code not checking results of the system calls.
  * A race condition in the control channel parsing on windows resulting in  a potential double-close() (which can close someone else's handle).
    
rdar://56800263

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -22,9 +22,11 @@
 #include "llvm/Support/ConvertUTF.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/Program.h"
+#include "llvm/Support/Compiler.h"
 
 #include <atomic>
 #include <thread>
+#include <memory>
 
 #include <fcntl.h>
 #if !defined(_WIN32)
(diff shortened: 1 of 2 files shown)


vlm:
@swift-ci please smoke test

lib/Basic/Subprocess.cpp
```diff
@@ -395,14 +643,17 @@ void llbuild::basic::spawnProcess(
...
-  const bool shouldCaptureOutput = !attr.connectToConsole;
 
   // Don't use lane release feature for console workloads.
   if (attr.connectToConsole) {
     attr.controlEnabled = false;
   }
 
+#if defined(_WIN32)
+  // Control channel support is broken (thread-unsafe) on Windows.
+  attr.controlEnabled = false;
```
vlm: WaitForMultipleObject was not configured in a way that makes it possible to use the control channel without doing a stack smash as well.

lib/Basic/Subprocess.cpp
```diff
@@ -713,44 +906,45 @@ void llbuild::basic::spawnProcess(
         ProcessInfo info{ attr.canSafelyInterrupt };
         pgrp.add(std::move(guard), pid, info);
       }
-    }
-  }
+
+    // Close the child ends of the forwarded output and control pipes.
+    controlPipeChildEnd.close();
+    outputPipeChildEnd.close();
```
vlm: The child ends should be closed under the lock because they are potentially inheritable by other processes if we are to spawn one in a parallel thread.

lib/Basic/Subprocess.cpp
```diff
@@ -882,36 +1058,32 @@ void llbuild::basic::spawnProcess(
...
-                 completionFn=std::move(completionFn)
-                 ]() mutable {
-        if (shouldCaptureOutput) {
-          captureExecutedProcessOutput(delegate, outputFd, handle, ctx);
-        } else {
-          assert(outputFd == -1);
+      std::shared_ptr<ManagedDescriptor> outputFdShared
+        = std::make_shared<ManagedDescriptor>(std::move(outputPipeParentEnd));
+      std::shared_ptr<ManagedDescriptor> controlFdShared
+        = std::make_shared<ManagedDescriptor>(std::move(controlPipeParentEnd));
```
vlm: The shared_ptrs are necessary because we disabled the copy constructor on a `ManagedDescriptor`, which makes it impossible to copy lambdas around, which makes it impossible to capture it in a lambda with a purpose of forwarding it further.

vlm:
@gmittert if you have a chance to check the windows portion that'd be great!

gmittert:
@vlm Sure! Might take me a bit though, I'm out of office this week.

PR #598, opened 2019-11-07
Re-enable max-commands-progress

vlm:
* The test flakiness was likely related to FS timestamp resolution.

rdar://23089862


vlm:
@swift-ci please smoke test

PR #601, opened 2019-11-09
build: limit clang flags to C++ (NFC)

compnerd:
This adjusts the application of the flags to just the C++ parts of the code,
enabling moving to CMake 3.15.


compnerd:
@swift-ci please smoke test

PR #603, opened 2019-11-19
Repair Windows build

compnerd:


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

lib/Basic/LaneBasedExecutionQueue.cpp
```diff
@@ -241,7 +241,7 @@ class LaneBasedExecutionQueue : public ExecutionQueue {
   /// Returns the number of allowed foreground and background tasks.
   static auto estimateTaskLimits(unsigned numLanes) -> std::pair<unsigned, unsigned> {
     llbuild_rlim_t curOpenFileLimit = llbuild::basic::sys::getOpenFileLimit();
-    const unsigned reservedFileCount = (STDERR_FILENO+1) + 2 /* Database */
+    const unsigned reservedFileCount = (fileno(stderr) + 1) + 2 /* Database */
```
vlm: 1. This code estimates the number of file descriptors used up by the current process. The three standard file descriptors (stdin, stdout, stderr) utilize three file descriptors. I didn't want to just put 3 in there; wanted to hint at how the value `3` is computed by using the largest file descriptor and adding 1 (since STDIN_FILENO is 0).
2. We don't care about the value of the `stderr` file descriptor in this piece of resource estimation code. Even if it is 42 we still use a single file descriptor for each one of {stdin, stdout, stderr{.
3. https://codereview.qt-project.org/c/qt/qt/+/80277 (via @jakepetroules)

If the reason for this code not to compile on Windows is unavailability of `STDERR_FILENO` the best solution seems to just put `3` instead and comment appropriately.

lib/Basic/PlatformUtility.cpp
```diff
@@ -28,6 +28,11 @@
 #endif
 #include <stdio.h>
 
+#if defined(_WIN32)
+const HANDLE llbuild::basic::sys::FileDescriptorTraits<HANDLE>::InvalidDescriptor =
+    INVALID_HANDLE_VALUE;
```
vlm: I don't know if it would work but would it work to just define `static const ... = INVALID_HANDLE_VALUE` up there in the trait itself? I would prefer not to split this logic between different files. Is there a particular reason for the approach taken in this patch?

vlm:
Thank you for the patch!

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

lib/Basic/PlatformUtility.cpp
compnerd: It doesn't actually work in the inline constructor (that was what I tried first), hence the split up.  Putting the definition in the header results in the value being defined multiply which also doesn't work.

compnerd:
@swift-ci please smoke test

vlm:
@swift-ci please smoke test

PR #608, opened 2019-11-22
[build] move evaluation of CMAKE_OSX_SYSROOT

edymtt:
The evaluation of CMAKE_OSX_SYSROOT needs to be done very early,
otherwise CMake will not adjust its variable to consider the SDK
correctly and may pick up headers in incorrect locations.

Addresses rdar://problem/57300418


edymtt:
Please test with following pull request:
apple/swift#28440

@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #609, opened 2019-11-28
build: remove `FindSQLite3` cmake module (NFC)

compnerd:
This was introduced in CMake 3.13, and we require CMake 3.15.  Remove
the local copy of the module.


compnerd:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #610, opened 2019-11-28
build: clean up some vestigial support code (NFC)

compnerd:
This was used for the parallel compilation of Swift code which has since
been migrated to CMake.  Remove the vestigial code to clean up the
module.


compnerd:
@swift-ci please smoke test

PR #612, opened 2019-12-01
build: add another workaround for CMake <3.16

compnerd:
Linking against libraries from the system was broken on Windows with
CMake 3.15.  It would require explicitly adding the `-l` prefix which is
less than ideal.  This workaround allows building without a dispatch or
Foundation build tree available for Windows.


compnerd:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

PR #615, opened 2019-12-11
build: llbuildSwift requires libllbuild

compnerd:
This is so that export targets can work properly for
swift-package-manager.  llbuildSwift should bring along the include
paths to libllbuild which it exposes.


compnerd:
@swift-ci please smoke test

PR #619, opened 2019-12-20
Adds Swift support for discovered dependencies

sergiocampama:
Also includes a couple commits to work around assumptions that are no longer true for dynamic ExternalCommand subclasses

### DIFF
--- a/products/llbuildSwift/BuildSystemBindings.swift
+++ b/products/llbuildSwift/BuildSystemBindings.swift
@@ -51,6 +51,11 @@ public struct BuildSystemCommandInterface {
     public func commandNeedsInput(key: BuildKey, inputID: UInt) {
         llb_buildsystem_command_interface_task_needs_input(_bsci, _task, key.internalBuildKey, inputID)
     }
+
+    /// Marks a build key as a runtime found dependency for the command.
+    public func commandDiscoveredDependency(key: BuildKey) {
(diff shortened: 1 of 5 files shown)


sergiocampama:
@swift-ci please smoke test

I expect this test to fail since I need to update swift-package-manager for the change in signature of the `execute` method

sergiocampama:
@swift-ci please smoke test

sergiocampama:
@swift-ci Please smoke test

sergiocampama:
Updated to include an incremental approach to API adoption.

sergiocampama:
@swift-ci Please smoke test

sergiocampama:
@swift-ci Please smoke test

sergiocampama:
@swift-ci Please smoke test

products/llbuildSwift/BuildSystemBindings.swift
```diff
@@ -144,9 +149,32 @@ public protocol ExternalCommand: class {
 
     /// Called to execute the given command.
     ///
+    /// This method is deprecated in favor of the execute(Command, BuildSystemCommandInterface) method.
```
jakepetroules: If you want, you could add an explicit deprecation annotation:

    @available(*, deprecated, renamed: "execute(_:_:)")

This would also give you an automatic fix-it in IDEs.

sergiocampama: Nifty, but this would introduce a build warning on line 170, on the workaround to invoke the deprecated command...

PR #620, opened 2019-12-20
[Core] Simplify task registration

dmbryson:
Removes registerTask() method, takes ownership directly during the
rule action call.


dmbryson:
@swift-ci please smoke test

dmbryson:
We definitely should, but given my next change in flight, I think I'll do so in a followup PR.

PR #621, opened 2020-01-02
SR-11730 Improved GraphViz output

tbartelmess:
This changes the `critical-path` to to include more meaningful information in the vertices instead of just the BuildKey ID from the database.

It also adds the option to only include the critical build path or the entire graph.

https://bugs.swift.org/browse/SR-11730

### DIFF
--- a/products/llbuild-analyze/Sources/CriticalPathTool.swift
+++ b/products/llbuild-analyze/Sources/CriticalPathTool.swift
@@ -22,11 +22,17 @@ public final class CriticalPathTool: Tool<CriticalPathTool.Options> {
         public enum OutputFormat: String, CaseIterable, CommandLineArgumentChoices {
             case json, graphviz
         }
+
+        public enum GraphvizDisplay: String, CaseIterable, CommandLineArgumentChoices {
+            case criticalPath
+            case everything
+        }
         
         var database: AbsolutePath!
         var outputPath: AbsolutePath?
         var clientSchemaVersion = 9
         var outputFormat: OutputFormat = .json
+        var graphvizDisplay: GraphvizDisplay = .criticalPath
         var pretty = false
         var quiet = false
         
@@ -63,6 +69,11 @@ public final class CriticalPathTool: Tool<CriticalPathTool.Options> {
         binder.bind(
             option: parser.add(option: "--pretty", usage: "If output format is json, the output will be pretty printed"),
             to: { $0.pretty = $1 })
+
+        binder.bind(
+            option: parser.add(option: "--graphvizOutput", usage: "Graphviz display options. Possible values: \(Options.GraphvizDisplay.helpDescription)."),
+            to: { $0.graphvizDisplay = Options.GraphvizDisplay(rawValue: $1) ?? .everything }
+        )
         
         binder.bind(
             option: parser.add(option: "--quiet", shortName: "-q", usage: "There will be no output on stdout."),
@@ -117,18 +128,25 @@ public final class CriticalPathTool: Tool<CriticalPathTool.Options> {
     
     private func graphViz(_ path: CriticalBuildPath, buildKeyLookup: IdentifierFactory<BuildKey>) -> Data {
         var result = "digraph G {\n\tedge [style=dotted]\n"
-        
-        func string(for key: BuildKey) -> String {
-            // TODO: This is obviously just temporary - need a shorter version for visualization
-            return buildKeyLookup.identifier(element: key).description
-        }
-        
-        for element in path {
-            for dep in element.result.dependencies {
-                result += "\t\"\(string(for: dep))\" -> \"\(string(for: element.key))\"\n"
+
+        var edges = Set<DirectedEdge>()
+
+        if var last = path.first {
+            for item in path[1..<path.endIndex] {
+                edges.insert(DirectedEdge(a: last.key, b: item.key, isCritical: true))
+                last = item
             }
         }
-        
+
+        if options.graphvizDisplay == .everything {
+            for element in path {
+                for dep in element.result.dependencies {
+                    edges.insert(DirectedEdge(a: dep, b: element.key))
+                }
+            }
+        }
+
+        result += edges.map{ $0.graphVizString }.joined()
         result += "}"
         return result.data(using: .utf8) ?? Data()
     }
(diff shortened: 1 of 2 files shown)


products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -12,20 +12,27 @@ import llbuildAnalysis
...
     
     public struct Options: Option {
-        public enum OutputFormat: String, CaseIterable {
+        public enum OutputFormat: String, CaseIterable, CommandLineArgumentChoices {
             case json, graphviz
         }
+
+        public enum GraphvizDisplay: String, CaseIterable, CommandLineArgumentChoices {
+            case criticalPath
+            case everything
```
BenchR267: nit: I think `all` is a better wording here.

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -56,12 +63,17 @@ public final class CriticalPathTool: Tool<CriticalPathTool.Options> {
...
-            option: parser.add(option: "--outputFormat", shortName: "-f", usage: "The format of the output, default is json. Possible values: \(Options.OutputFormat.allCases.map({ $0.rawValue }))."),
+            option: parser.add(option: "--outputFormat", shortName: "-f", usage: "The format of the output, default is json. Possible values: \(Options.OutputFormat.helpDescription)."),
             to: { $0.outputFormat = Options.OutputFormat(rawValue: $1) ?? .json })
         
         binder.bind(
             option: parser.add(option: "--pretty", usage: "If output format is json, the output will be pretty printed"),
             to: { $0.pretty = $1 })
+
+        binder.bind(
+            option: parser.add(option: "--graphvizOutput", usage: "Graphviz display options. Possible values: \(Options.GraphvizDisplay.helpDescription)."),
```
BenchR267: Please add the default value here for the `--help` output.

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -272,3 +291,119 @@ struct EncodableRuleResult: Encodable {
         try container.encode(deps, forKey: .dependencies)
     }
 }
+
+
+protocol GraphVizNode {
```
BenchR267: Could we move all of that to a new file dedicated only for the graph viz output? We should do the same for the JSON serialization code, but that doesn't need to be in this PR.

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -272,3 +291,119 @@ struct EncodableRuleResult: Encodable {
...
+    var graphVizName: String { get }
```
BenchR267: Is the output format you chose a convention for GraphViz? If not, could we use the existing logic in `BuildKey.description`? That uses a xml-like representation of the object, e.g. for `BuildKey.Command`: `<BuildKey.\(type(of: self)) name=\(name)>` -> `<BuildKey.Command name=A>`.

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -272,3 +291,119 @@ struct EncodableRuleResult: Encodable {
...
+/// and gurantee that there is only one edge between two verticies.
+struct DirectedEdge: Hashable, Equatable {
+    /// Source `BuildKey`
+    let a: BuildKey
+
+    /// Destination `BuildKey`
+    let b: BuildKey
+
+    /// Flag if the edge is on critical build path.
+    var isCritical: Bool = false
```
BenchR267: Why is this mutable?

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -272,3 +291,119 @@ struct EncodableRuleResult: Encodable {
...
+        if isCritical {
+            return "[style=bold]"
+        }
+        return ""
+    }
+
+    /// GraphViz representation of the Edge.
+    var graphVizString: String {
+        guard let a = a as? GraphVizNode, let b = b as? GraphVizNode else {
+            fatalError("Both edges need to conform to GraphvizNode")
```
BenchR267: Please add some context to the error string. Something like: `Both edges need to conform to GraphVizNode to generate a graphVizString for DirectedEdge \(self)`.

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -272,3 +291,119 @@ struct EncodableRuleResult: Encodable {
...
+        }
+        return "\t\"\(a.graphVizName)\" -> \"\(b.graphVizName)\"\(style)\n"
+    }
+
+}
+
+private protocol CommandLineArgumentChoices: CaseIterable, RawRepresentable where RawValue == String {
```
BenchR267: Would it make sense to define the protocol without type constraints and have the extension use them so that other types could implement this as well? Would probably need a deeper implementation and a more sophisticated help description system to make sense.

BenchR267:
@swift-ci please smoke test

ddunbar:
Thanks @tbartelmess, this is awesome!

products/llbuild-analyze/Sources/CriticalPathTool.swift
tbartelmess: @BenchR267 I think that makes sense,  however, I am not sure if that should be in scope for SR-11730. Should I open a new ticket at bugs.swift.org?

products/llbuild-analyze/Sources/CriticalPathTool.swift
tbartelmess: It's not a convention, I thought the XML-ish syntax looked a bit odd in the generated PDF.

For now, I've made the graphVizName return the description, maybe we can add options what info to include in a future version.

BenchR267: The XML-ish description is not a convention either. We could also change that 🙂

products/llbuild-analyze/Sources/CriticalPathTool.swift
BenchR267: Yeah, that makes sense. Yes, please. Thanks!

products/llbuild-analyze/Sources/GraphViz.swift
```diff
@@ -0,0 +1,64 @@
...
+    private var style: String {
+        if isCritical {
+            return "[style=bold]"
+        }
+        return ""
+    }
+
+    /// GraphViz representation of the Edge.
+    var graphVizString: String {
+        guard let a = a as? GraphVizNode, let b = b as? GraphVizNode else {
```
BenchR267: This check isn't necessary anymore as every BuildKey is a GraphVizNode \o/
It also generates a warning due to that.

tbartelmess: Sorry, I've missed that.

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

PR #622, opened 2020-01-10
Refactor Rule to be a virtual base class

dmbryson:
In this change, BuildSystem, et al. still use std::function objects.
However, this change lays the foundation for eliminating those.

### DIFF
--- a/include/llbuild/Core/BuildEngine.h
+++ b/include/llbuild/Core/BuildEngine.h
@@ -214,7 +214,7 @@ public:
   /// state managed externally to the build engine. For example, a rule which
   /// computes something on the file system may use this to verify that the
   /// computed output has not changed since it was built.
-  virtual bool isResultValid(BuildEngine&, const ValueType&);
+  virtual bool isResultValid(BuildEngine&, const ValueType&) = 0;
 
   /// Called to indicate a change in the rule status.
   virtual void updateStatus(BuildEngine&, StatusKind);
--- a/include/llbuild/Core/BuildEngine.h
+++ b/include/llbuild/Core/BuildEngine.h
@@ -188,14 +188,24 @@ public:
     SupplyPriorValue = 1
   };
 
+public:
   /// The key computed by the rule.
-  KeyType key;
+  const KeyType key;
 
   /// The signature of the rule.
-  basic::CommandSignature signature;
+  const basic::CommandSignature signature;
+
+private:
+  Rule(const Rule&) LLBUILD_DELETED_FUNCTION;
+  void operator=(const Rule&) LLBUILD_DELETED_FUNCTION;
+
+public:
+  Rule(const KeyType& key, const basic::CommandSignature& signature = {})
+    : key(key), signature(signature) { }
+  virtual ~Rule() = 0;
 
   /// Called to create the task to build the rule, when necessary.
-  std::function<Task*(BuildEngine&)> action;
+  virtual Task* createTask(BuildEngine&) = 0;
 
   /// Called to check whether the previously computed value for this rule is
   /// still valid.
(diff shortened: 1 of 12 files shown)


include/llbuild/Core/BuildEngine.h
```diff
@@ -204,11 +214,10 @@ class Rule {
   /// state managed externally to the build engine. For example, a rule which
   /// computes something on the file system may use this to verify that the
   /// computed output has not changed since it was built.
-  std::function<bool(BuildEngine&, const Rule&,
-                     const ValueType&)> isResultValid;
+  virtual bool isResultValid(BuildEngine&, const ValueType&);
 
   /// Called to indicate a change in the rule status.
-  std::function<void(BuildEngine&, StatusKind)> updateStatus;
```
dmbryson: Substance of the change is here.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -1513,7 +1513,54 @@ convertStatusKind(core::Rule::StatusKind kind) {
   return BuildSystemDelegate::CommandStatusKind::IsScanning;
 }
 
-Rule BuildSystemEngineDelegate::lookupRule(const KeyType& keyData) {
+class BuildSystemRule : public Rule {
```
dmbryson: This is a temporary bridge between the old world (std::function) and the new world.

dmbryson:
@swift-ci please smoke test

include/llbuild/Core/BuildEngine.h
```diff
@@ -204,11 +214,10 @@ class Rule {
   /// state managed externally to the build engine. For example, a rule which
   /// computes something on the file system may use this to verify that the
   /// computed output has not changed since it was built.
-  std::function<bool(BuildEngine&, const Rule&,
-                     const ValueType&)> isResultValid;
+  virtual bool isResultValid(BuildEngine&, const ValueType&);
```
ddunbar: Should we make these required, just so that this is a purely abstract base class? If so, remember to give the vtable a home.

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -1526,7 +1573,7 @@ Rule BuildSystemEngineDelegate::lookupRule(const KeyType& keyData) {
     auto it = getBuildDescription().getCommands().find(key.getCommandName());
     if (it == getBuildDescription().getCommands().end()) {
       // If there is no such command, produce an error task.
-      return Rule{
+      return std::unique_ptr<Rule>(new BuildSystemRule(
```
ddunbar: We normally use make_unique, don't we?

dmbryson: Yeah. make_unique wasn't happy with these, though.  I think perhaps because of the default argument values?

include/llbuild/Core/BuildEngine.h
dmbryson: Possibly.  The `isResultValid` method does feel like it is something that rules should be explicit about.   The `updateStatus` method, on the other hand, is definitely optional, and as I look at it perhaps could/should be removed altogether with some more refactoring.

dmbryson:
```
/home/buildnode/jenkins/workspace/swift-llbuild-PR-Linux-smoke-test/llbuild/products/libllbuild/Core-C-API.cpp:78:12: error: no viable conversion from returned value of type 'unique_ptr<(anonymous namespace)::CAPIBuildEngineDelegate::CAPIRule, default_delete<(anonymous namespace)::CAPIBuildEngineDelegate::CAPIRule>>' to function return type 'unique_ptr<llbuild::core::Rule, default_delete<llbuild::core::Rule>>'
```

Really linux clang?

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #623, opened 2020-01-12
Fix remaining tests for Android

buttaface:
Used to build and test llbuild and SPM natively on Android in [the Termux app](https://termux.com/). I can modify lit.cfg to only check the `PREFIX` env var on Android if there's any worry that it might be spuriously set on other platforms.

`makeTmpDir` is only checked by the unit tests, which I noticed are not run by the CI anymore, maybe because those llbuild test binaries were moved around? I see the following warning on both the linux and mac CI for recent pulls:
```
lit.py: /home/buildnode/jenkins/workspace/swift-llbuild-PR-Linux-smoke-test/llvm-project/llvm/utils/lit/lit/discovery.py:210: warning: test suite 'llbuild-unit' contained no tests
```
Running them manually shows that no tests fail after this pull on Android.

### DIFF
--- a/lib/Basic/PlatformUtility.cpp
+++ b/lib/Basic/PlatformUtility.cpp
@@ -348,8 +348,17 @@ std::string sys::makeTmpDir() {
   CreateDirectoryW((LPCWSTR)wPath.data(), NULL);
   return std::string(path);
 #else
-  char tmpDirPathBuf[] = "/tmp/fileXXXXXX";
-  return std::string(mkdtemp(tmpDirPathBuf));
+  if (const char *tmpDir = std::getenv("TMPDIR")) {
+    char *tmpDirPath = nullptr;
+    asprintf(&tmpDirPath, "%s/fileXXXXXX", tmpDir);
+    auto tmpDirString = std::string(mkdtemp(tmpDirPath));
+    free(tmpDirPath);
+    return tmpDirString;
+  }
+  else {
+    char tmpDirPathBuf[] = "/tmp/fileXXXXXX";
+    return std::string(mkdtemp(tmpDirPathBuf));
+  }
 #endif
 }
 
--- a/tests/BuildSystem/Build/dep-scan-order.llbuild
+++ b/tests/BuildSystem/Build/dep-scan-order.llbuild
@@ -37,7 +37,7 @@
 # Check that a modification triggers both uses.
 #
 # RUN: echo "MOD" >> %t.build/input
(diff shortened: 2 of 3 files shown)


lib/Basic/PlatformUtility.cpp
```diff
@@ -348,7 +348,11 @@ std::string sys::makeTmpDir() {
   CreateDirectoryW((LPCWSTR)wPath.data(), NULL);
   return std::string(path);
 #else
-  char tmpDirPathBuf[] = "/tmp/fileXXXXXX";
+  char *tmpDirPathBuf = nullptr;
+  if (const char *tmpDir = std::getenv("TMPDIR"))
+    asprintf(&tmpDirPathBuf, "%s/fileXXXXXX", tmpDir);
```
dmbryson: The buffer allocated here needs to be `free()`d before we return.

buttaface: Done.

lib/Basic/PlatformUtility.cpp
```diff
@@ -348,8 +348,14 @@ std::string sys::makeTmpDir() {
...
 #else
-  char tmpDirPathBuf[] = "/tmp/fileXXXXXX";
-  return std::string(mkdtemp(tmpDirPathBuf));
+  char *tmpDirPath = nullptr;
+  if (const char *tmpDir = std::getenv("TMPDIR"))
+    asprintf(&tmpDirPath, "%s/fileXXXXXX", tmpDir);
+  else
+    tmpDirPath = "/tmp/fileXXXXXX";
+  auto tmpDirString = std::string(mkdtemp(tmpDirPath));
+  free(tmpDirPath);
```
dmbryson: 🤔 This may now attempt to free the inline `"/tmp/fileXXXXXX"`.

buttaface: I think this last iteration should finally work fine.

dmbryson:
@swift-ci please smoke test

buttaface:
Hmm, I didn't know if the PackageDescription version would make a difference like it did natively on Android, apple/swift-package-manager#2466, guess it does so removed that Android addition for ncurses till the manifest version supports it.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

buttaface:
Thanks, about the unit tests, I notice that `warning: test suite 'llbuild-unit' contained no tests` in both the linux and mac CI logs for this pull, but I now see that the mac CI circles back around at the end and runs the unit tests some other way, though they still don't appear to be run on the linux CI. Oversight or conscious choice?

PR #626, opened 2020-01-15
[Core] Refactor key into an encapsulated type

dmbryson:
Key remains a std::string under the covers, but clients must now be
explicit when they wish to use it as such.

### DIFF
--- a/include/llbuild/Core/BuildEngine.h
+++ b/include/llbuild/Core/BuildEngine.h
@@ -34,8 +34,26 @@
 namespace llbuild {
 namespace core {
 
-// FIXME: Need to abstract KeyType;
-typedef std::string KeyType;
+class KeyType {
+private:
+  std::string key;
+
+public:
+  KeyType() { }
+  KeyType(const std::string& key) : key(key) { }
+  KeyType(const char* key) : key(key) { }
+  KeyType(llvm::StringRef ref) : key(ref) { }
+  KeyType(const char* data, size_t length) : key(data, length) { }
+
+  bool operator==(const KeyType& rhs) const { return key == rhs.key; }
+  bool operator<(const KeyType& rhs) const { return key < rhs.key; }
+
+  const std::string& str() const { return key; }
+  const char* data() const { return key.data(); }
+  const char* c_str() const { return key.c_str(); }
+  size_t size() const { return key.size(); }
+};
+
 typedef std::vector<uint8_t> ValueType;
 
 class BuildDB;
@@ -429,6 +447,18 @@ public:
 };
 
 }
+}
+
+namespace std
+{
+  template<>
+  struct hash<llbuild::core::KeyType>
+  {
+    size_t operator()(const llbuild::core::KeyType& key) const
+    {
+      return hash<std::string>{}(key.str());
+    }
+  };
 }
 
 #endif
--- a/lib/Commands/BuildEngineCommand.cpp
+++ b/lib/Commands/BuildEngineCommand.cpp
@@ -90,7 +90,7 @@ struct AckermannKey {
 
   /// Create an Ackermann key from the encoded representation.
   AckermannKey(const core::KeyType& key) {
-      auto keyString = StringRef(key);
+      auto keyString = StringRef(key.str());
       assert(keyString.startswith("ack(") && keyString.endswith(")"));
       auto arguments = keyString.split("(").second.split(")").first.split(",");
       m = 0;
--- a/lib/Commands/NinjaBuildCommand.cpp
+++ b/lib/Commands/NinjaBuildCommand.cpp
@@ -1681,7 +1681,7 @@ std::unique_ptr<core::Rule> NinjaBuildEngineDelegate::lookupRule(const core::Key
   // FIXME: This is frequently a redundant lookup, given that the caller might
   // well have had the Node* available. This is something that would be nice
   // to avoid when we support generic key types.
-  ninja::Node* node = context->manifest->findOrCreateNode(workingDirectory, key);
+  ninja::Node* node = context->manifest->findOrCreateNode(workingDirectory, key.str());
 
   class NinjaInputRule: public core::Rule {
     BuildContext* context;
@@ -1716,7 +1716,7 @@ void NinjaBuildEngineDelegate::cycleDetected(
   for (const auto* rule: cycle) {
     if (!first)
       messageStream << " ->";
-    messageStream << " \"" << rule->key << '"';
+    messageStream << " \"" << rule->key.str() << '"';
     first = false;
   }
 
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -72,7 +72,7 @@ class BuildEngineImpl : public BuildDBDelegate {
 
   /// The mutex that protects the key table.
   std::mutex keyTableMutex;
-  
+
   /// The build database, if attached.
   std::unique_ptr<BuildDB> db;
 
@@ -87,7 +87,7 @@ class BuildEngineImpl : public BuildDBDelegate {
 
   /// Whether a build is currently running.
   std::atomic<bool> buildRunning{ false };
-  
+
   /// The queue of input requests to process.
   struct TaskInputRequest {
     /// The task making the request.
@@ -166,7 +166,7 @@ class BuildEngineImpl : public BuildDBDelegate {
 
     /// The rule this information describes.
     std::unique_ptr<Rule> rule;
-    
+
     /// The state dependent record for in-progress information.
     union {
       RuleScanRecord* pendingScanRecord;
@@ -208,7 +208,7 @@ class BuildEngineImpl : public BuildDBDelegate {
       return state == StateKind::Complete &&
         result.builtAt == engine->getCurrentEpoch();
     }
-    
+
     void setComputing(const BuildEngineImpl* engine) {
       assert(isInProgressWaiting());
       state = StateKind::InProgressComputing;
@@ -236,7 +236,7 @@ class BuildEngineImpl : public BuildDBDelegate {
       // longer valid.
       state = StateKind::Incomplete;
     }
-    
+
     RuleScanRecord* getPendingScanRecord() {
       assert(isScanning());
       return inProgressInfo.pendingScanRecord;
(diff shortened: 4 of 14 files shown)


dmbryson:
@swift-ci please smoke test

ddunbar:
After having wrote this patch, how convinced are you of whether or not this is the right direction?

include/llbuild/Core/BuildEngine.h
```diff
@@ -34,8 +34,39 @@
 namespace llbuild {
 namespace core {
 
-// FIXME: Need to abstract KeyType;
-typedef std::string KeyType;
+struct KeyType {
+  std::string function;
+  std::vector<uint8_t> args;
```
ddunbar: Should we do a custom allocator for this so we can be more efficient? We probably want these to be immutable, and could do a single heap allocation and end up with substantially less memory use than this (std::string and std::vector are probably each 3 words or more?)

include/llbuild/Core/BuildEngine.h
```diff
@@ -34,8 +34,39 @@
...
+    return function == rhs.function && args == rhs.args;
+  }
+
+  bool operator<(const KeyType& rhs) const {
+    return function <= rhs.function && args < rhs.args;
+  }
+
+  std::string describe() const {
+    // FIXME: encode args
+    return function;
```
ddunbar: This seems important? :)

ddunbar: What is `describe` for? If it is for debugging, LLVM has a convention of "dump" methods that write to a raw_ostream, I'd somewhat rather follow that than unnecessary heap allocations.

include/llbuild/Core/BuildEngine.h
```diff
@@ -34,8 +34,39 @@
...
+struct KeyType {
+  std::string function;
+  std::vector<uint8_t> args;
+
+  bool operator==(const KeyType& rhs) const {
+    return function == rhs.function && args == rhs.args;
+  }
+
+  bool operator<(const KeyType& rhs) const {
+    return function <= rhs.function && args < rhs.args;
```
ddunbar: This is not a correct `<` function. :)

include/llbuild/Core/BuildEngine.h
```diff
@@ -34,8 +34,39 @@
...
+  KeyType(const std::string& legacyKey) : function(legacyKey) { }
+  KeyType(llvm::StringRef legacyRef) : function(legacyRef) { }
+  KeyType(const char* str) : function(str) { }
+  KeyType(const unsigned char* const start, const unsigned char* const end) : function(start, end) { }
+  KeyType(const char* data, size_t length) : function(data, length) { }
+
+  const char* data() const { return function.data(); }
+  const char* c_str() const { return function.c_str(); }
+  size_t size() const { return function.size(); }
+  size_t length() const { return function.length(); }
```
ddunbar: Do we need all these? Are these also just shims we can eventually eliminate?

If so, while not particularly important, it would be cleaner to refactor KeyType to its own type with just a single string inside of it, then refactor the clients, then do the key + args refactor.

include/llbuild/Core/BuildEngine.h
```diff
@@ -34,8 +34,39 @@
...
+
+  operator llvm::StringRef() const { return llvm::StringRef(function); }
```
ddunbar: I have a mild preference for a `str()` method that returns the StringRef (similar to SmallString) unless LLVM also provides conversion operators on similar types.

include/llbuild/Core/BuildEngine.h
```diff
@@ -431,4 +462,16 @@ class BuildEngine {
...
 }
 
+namespace std
+{
+  template<>
+  struct hash<llbuild::core::KeyType>
+  {
+    size_t operator()(const llbuild::core::KeyType& key) const
+    {
+      return hash<std::string>{}(key.function);
```
ddunbar: This doesn't seem correct, to use function & args in == and < but not in the hash.

lib/Commands/BuildEngineCommand.cpp
```diff
@@ -90,7 +90,7 @@ struct AckermannKey {
 
   /// Create an Ackermann key from the encoded representation.
   AckermannKey(const core::KeyType& key) {
-      auto keyString = StringRef(key);
+      auto keyString = static_cast<StringRef>(key);
```
ddunbar: `key.str()` seems nicer here, for example

ddunbar: However... I find this patch unclear to understand the benefit without actually understand how code like this would use the args. I'm a little worried that we will land something like this, then actually go and use the args, and then maybe end up with questions about whether it was a good idea. I think I'd like to see how code like this uses the args before weighing in on this.

include/llbuild/Core/BuildEngine.h
ddunbar: From later, it looks like describe is not just for debugging, so definitely deserves a doc comment explaining it.

ddunbar: (and also, still, I think I would prefer that it write to a raw_ostream so we avoid a lot of heap allocations in using it).

This will require various refactors in the clients, which does make me think it might be worth doing a refactor to switch KeyType to its own struct, add a bunch of methods, and then stage in the function/args split later.

dmbryson:
As discussed offline, this definitely ended up looking a lot messier than we had thought.  Reworked this change to merely be about encapsulating KeyType so that clients must be explicit when using it as a std::string.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

include/llbuild/Core/BuildEngine.h
```diff
@@ -34,8 +34,26 @@
 namespace llbuild {
 namespace core {
 
-// FIXME: Need to abstract KeyType;
-typedef std::string KeyType;
+class KeyType {
```
ddunbar: Maybe good to have a comment here?

lib/Commands/NinjaBuildCommand.cpp
```diff
@@ -1681,7 +1681,7 @@ std::unique_ptr<core::Rule> NinjaBuildEngineDelegate::lookupRule(const core::Key
   // FIXME: This is frequently a redundant lookup, given that the caller might
   // well have had the Node* available. This is something that would be nice
   // to avoid when we support generic key types.
-  ninja::Node* node = context->manifest->findOrCreateNode(workingDirectory, key);
+  ninja::Node* node = context->manifest->findOrCreateNode(workingDirectory, key.str());
```
ddunbar: Should this method just take the key?

lib/Core/BuildEngine.cpp
```diff
@@ -1329,7 +1329,7 @@ class BuildEngineImpl : public BuildDBDelegate {
     // The RHS of the mapping is actually ignored, we use the StringMap's ptr
     // identity because it allows us to efficiently map back to the key string
     // in `getRuleInfoForKey`.
-    auto it = keyTable.insert(std::make_pair(key, KeyID::novalue())).first;
```
ddunbar: Should the key table just map based on the key?

products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -572,7 +572,7 @@ class CAPIBuildSystem {
     return getFrontend().initialize();
   }
 
-  bool build(const core::KeyType& key) {
+  bool build(llvm::StringRef key) {
```
ddunbar: Shouldn't this stay as KeyType?

unittests/Core/DepsBuildEngineTest.cpp
```diff
@@ -222,7 +222,7 @@ TEST(DepsBuildEngineTest, BogusConcurrentDepScan) {
     DynamicRule(const KeyType& key, std::vector<std::string>& builtKeys)
       : Rule(key), builtKeys(builtKeys) { }
     Task* createTask(BuildEngine&) override {
-      builtKeys.push_back(key);
+      builtKeys.push_back(key.str());
```
ddunbar: Feels like `builtKeys` should store KeyType.

lib/Commands/NinjaBuildCommand.cpp
dmbryson: In this case it won't get us much.  The code takes the StringRef that this is implicitly converted to and immediately moves it into a SmallString.  It also passes the StringRef as is to a Node().   Changing that all to use KeyTypes is probably a good idea, but out of scope here.

lib/Core/BuildEngine.cpp
dmbryson: keyTable is a StringMap, and as such, wants a StringRef for the key. Seems a reasonable choice of data structure for what we are doing with it right now.   We could split usages that want a StringRef out from a std::string&.  That could allow us to swap out the internals and differentiate performance for those that actually need a std::string from those that just need the ref.  That said, though, it can also be done later, if/when we want to do that.

ddunbar: Makes sense

products/libllbuild/BuildSystem-C-API.cpp
dmbryson: This is a case where we were relying on the fact that keyType was a std::string.   We were getting auto conversion to it and then to a stringref.   The build system layer this calls doesn't actually want a KeyType. It takes the string and constructs a Target build key out of it.

As such, I think it would actually be better to rename this method to buildTarget to describe what it actually does.   That is a much larger change, though.

unittests/Core/DepsBuildEngineTest.cpp
dmbryson: Sure. Probably worth to have a little more explicit coverage of keytype.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

dmbryson:
@swift-ci please smoke test osx

dmbryson:
sad panda @swift-ci

dmbryson:
@swift-ci please smoke test osx

dmbryson:

dmbryson:

PR #632, opened 2020-02-06
Fix TSan errors found via test execution

dmbryson:
rdar://problem/26356623


dmbryson:
@swift-ci please smoke test

PR #633, opened 2020-02-06
Unify adjust-times warnings with top level

dmbryson:
The adjust-times target was setting different levels of warnings, thus
triggering unexpected variation from the rest of the project.

rdar://problem/57171721&57171722&57171723


dmbryson:
@swift-ci please smoke test

PR #634, opened 2020-02-09
Fix broken symlink for Demangle's include

aciidb0mb3r:
These symlinks are used for SwiftPM support and some recent improvements
in SwiftPM caught this broken symlink.


aciidb0mb3r:
@swift-ci smoke test

PR #635, opened 2020-02-13
[Core] Refactor Task Interface

dmbryson:
Lays the foundation for build environment support, bringing task,
job processing, and process execution into a unified interface.


dmbryson:
@swift-ci please smoke test

include/llbuild/Core/BuildEngine.h
```diff
@@ -101,6 +98,121 @@ struct Result {
...
+/// non-ideal systems to still gain some of the benefits.
+///
+/// The engine stores references to environments and manages merging them on
+/// behalf of tasks executing during a build. The actual value of an
+/// environment, however, is regarded as opaque and not directly interpreted by
+/// the engine.
+typedef void* Environment;
+
+
+class TaskInterface {
```
dmbryson: This is the core of the change.  Tasks now get this interface instead of a build engine reference.   Behind the scenes this will allow the engine to track environment state across each of the operations on it.

include/llbuild/Core/BuildEngine.h
```diff
@@ -259,6 +371,24 @@ class BuildEngineDelegate {
   /// that the client can translate into an error.
   virtual std::unique_ptr<Rule> lookupRule(const KeyType& key) = 0;
 
+  /// Provide a new environment object representing the starting/null state.
+  ///
+  /// The engine implicitly retains a reference to the environment object until
+  /// it calls \see releaseEnvironment().
+  virtual Environment nullEnvironment() = 0;
+
```
ddunbar: I think the initial environment should be an arg to build()?

ddunbar:
I think we should write up a proposal of “llbuild environments” support that we can post to llbuild-dev on forums. Even if just you and I read it, it will be a good reference to have on how/why/when this support was added.

This would likely be a useful reference to link to from the “content hashing” thread going on, as we start to talk about how llbuild would integrate with CAS technology.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

PR #636, opened 2020-02-16
test: use `lit` to provide `which`

compnerd:
Rather than attempting to implement the utility finding by hand, which
does not account for builtin utilities, use `lit` properly.


compnerd:
@swift-ci please smoke test

PR #637, opened 2020-02-16
test: use idiomatic python for conditional evaluation

compnerd:
Use the more traditional `if` condition checks for the flag computation.
This will become more useful as more platforms are added.


compnerd:
@swift-ci please smoke test

PR #638, opened 2020-02-16
tests: do not assume `cc` is valid

compnerd:
On Windows, the default compiler is spelt `cl` as MSVC is the proffered
compiler in that environment.  `cc` is a unicism for the C compiler.
Explicitly use `clang` as the compiler as the test uses GNU command line
options.

This replicates the `swiftInferBinary` function from Swift to replicate
the environment handling and logging.


compnerd:
@swift-ci please smoke test

PR #640, opened 2020-02-18
test: enable using lit's internal shell parser

compnerd:
Not all platforms have a full featured shell or a UNIX shell.  Enable
the use of the lit interpreter instead as a shell.


compnerd:
@swift-ci please smoke test

PR #641, opened 2020-02-18
Improve modification detection by adding content comparison

igor-makarov:
This PR revises the way the `BuildSystem` detects file modification for rebuild. 

Prior to this, the definition of a modified files was through `fstat` properties: `device`, `inode`, `mode`, `size` and `mtime`.

This approach is very robust, but has false positives in several common cases:

* git operations - when you move between branches, sometimes you end up with exactly the same file tree content, but git will have touched the files several times.
* Scripts generating code - the contents would be the same, but the file might be completely different by `mtime`, `inode` etc.
* Accidental edits to files - you type in the wrong window and fix it a moment later, but the file was modified.
* Compiler generated files - you rebuild some module, causing the ObjC interface header to be rebuilt, which cascades the other parts of the project regardless of whether the file is actually different.

The current PR changes the way file modification is defined:
* `device`, `inode`, `mode` are no longer used.
* The file is compared by `size` and then by `mtime`.
* If and only if the `mtime` comparison fails, the MD5 hash of the file's contents is compared to the previous value.

This follows a [forum discussion](https://forums.swift.org/t/using-md5-of-the-file-contents-as-a-proxy-for-modification/33858). 
Issues raised in the discussion but not addressed by the PR:

* MD5 performance - while this is a reasonable concern, the calculation will be performed lazily and not affect build system performance. I couldn't find a better hashing function in the confines of the project, and one can be swapped in later.
* Non-content attribute cases.

### DIFF
--- a/include/llbuild/Basic/FileInfo.h
+++ b/include/llbuild/Basic/FileInfo.h
@@ -29,6 +29,7 @@
 
 #include <cstdint>
 #include <string>
+#include "llvm/Support/FileSystem.h"
 
 namespace llbuild {
 namespace basic {
@@ -60,6 +61,28 @@ struct FileTimestamp {
   }
 };
 
+struct LazyMD5Digest {
+  std::string path = "";
+  mutable std::pair<uint64_t, uint64_t> digest;
+  mutable bool needsCalculation;
+  
+  bool operator==(const LazyMD5Digest& rhs) const {
+    ensureCalculated();
+    rhs.ensureCalculated();
+    return digest == rhs.digest;
+  }
+  bool operator!=(const LazyMD5Digest& rhs) const {
+    return !(*this == rhs);
+  }
+
+  void ensureCalculated() const {
+    if (needsCalculation && !path.empty()) {
+      digest = llvm::sys::fs::md5_contents(Twine(path))->words();
+      needsCalculation = false;
+    }
+  }
+};
+
 /// File information which is intended to be used as a proxy for when a file has
 /// changed.
 ///
@@ -76,7 +99,7 @@ struct FileInfo {
   /// The modification time of the file.
   FileTimestamp modTime;
   /// The MD5 hash of the file.
-  std::pair<uint64_t, uint64_t> digest;
+  LazyMD5Digest digest;
 
   /// Check if this is a FileInfo representing a missing file.
   bool isMissing() const {
@@ -91,7 +114,7 @@ struct FileInfo {
   
   bool operator==(const FileInfo& rhs) const {
     return (size == rhs.size &&
-            digest == rhs.digest);
+            (modTime == rhs.modTime || digest == rhs.digest));
   }
   bool operator!=(const FileInfo& rhs) const {
     return !(*this == rhs);
@@ -121,6 +144,19 @@ struct BinaryCodingTraits<FileTimestamp> {
 };
 
 template<>
+struct BinaryCodingTraits<LazyMD5Digest> {
+  static inline void encode(const LazyMD5Digest& value, BinaryEncoder& coder) {
+    value.ensureCalculated();
+    coder.write(value.digest.first);
+    coder.write(value.digest.second);
+  }
+  static inline void decode(LazyMD5Digest& value, BinaryDecoder& coder) {
+    coder.read(value.digest.first);
+    coder.read(value.digest.second);
+  }
+};
+
+template<>
 struct BinaryCodingTraits<FileInfo> {
   static inline void encode(const FileInfo& value, BinaryEncoder& coder) {
     coder.write(value.device);
@@ -128,8 +164,7 @@ struct BinaryCodingTraits<FileInfo> {
     coder.write(value.mode);
     coder.write(value.size);
     coder.write(value.modTime);
-    coder.write(value.digest.first);
-    coder.write(value.digest.second);
+    coder.write(value.digest);
   }
   static inline void decode(FileInfo& value, BinaryDecoder& coder) {
     coder.read(value.device);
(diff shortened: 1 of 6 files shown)


ddunbar:
Thanks for the PR, sorry for the delay in review!

A couple high level comments:
1. We should probably model this a bit differently. Rather than have a LazyMD5 abstraction, we should leverage the power of the build engine so that what we actually do is alter the build graph so that when a task depends on file X, it’s graph is rewritten to depend on a node “hash(X)” which depends on X. Then the build engine will handle only bringing the hash up to date when the inode is modified (and there shouldn’t be any changes to existing FileInfo struct.

2. We will want clients to be able to selectively choose whether or not to use this behavior. To start with, I’d suggest we maybe make it so that tasks can opt in or out of this behavior at the BuildSystem manifest level (maybe with a way to toggle the global default, in order to keep configs small).

WDYT?

ddunbar:
As far as the hash algorithm goes, for similar use cases we’ve lately been using BLAKE2 for this kind of use case, and I think it’s a very reasonable place to start. We can however start with MD5 and just anticipate that it is pluggable and we will swap it out...

https://blake2.net/acns/slides.html
https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2P

igor-makarov:
@ddunbar thanks for the comments!
For the first part, I do think this is a great suggestion, I wish I'd have thought of it myself! I'd like some more guidance on how to do this - do I create a new rule? Task? 

For the second part - I'm not even sure how to approach this. What should I do about it? Do you think it's necessary for this PR to include it?

include/llbuild/BuildSystem/BuildValue.h
```diff
@@ -113,14 +113,8 @@ class BuildValue {
...
-  union {
-    /// The file info for the rule output, for existing inputs, successful
-    /// commands with a single output, and directory contents.
-    FileInfo asOutputInfo;
-
-    /// The file info for successful commands with multiple outputs.
-    FileInfo* asOutputInfos;
-  } valueData = { {} };
+  /// The file info for successful commands with multiple outputs or a single output.
+  FileInfo* outputInfos;
```
dmbryson: I like the simplification here removing the union, but perhaps it would be best in a separate PR?

include/llbuild/Basic/FileInfo.h
```diff
@@ -88,10 +113,8 @@ struct FileInfo {
   bool isDirectory() const;
   
   bool operator==(const FileInfo& rhs) const {
-    return (device == rhs.device &&
-            inode == rhs.inode &&
-            size == rhs.size &&
-            modTime == rhs.modTime);
+    return (size == rhs.size &&
```
dmbryson: Although this does make the behavior automatic, I'm not sure that changing operator== in this way is correct.   I think a user would rightly expect that comparing a FileInfo object would yield a deep comparison of all fields.   The contents comparison, as written here, should probably be a separate, explicit method.  That would also enable us to more cleanly implement @ddunbar 's comment about making this an opt-in feature.

include/llbuild/BuildSystem/BuildValue.h
dmbryson: Hmm, a counterpoint for the simplification, I think the reason for the union is to avoid the allocation overhead on the many single output build values we might have floating around.

ddunbar:
@igor-makarov 

For the first part, take a look at `DirectoryTreeSignatureTask` in BuildSystem.

For the second part, we would probably do it by adding attributes to the BuildSystem manifest, one attribute for the tool, to switch the default, and one attribute for the task, to override the tool default. Search for `can-safely-interrupt` for an example of a ShellTask attribute (although it doesn't have an example of the global default.

Unfortunately, yes, the second part is definitely required because this is a potential behavior changing thing that clients (like SwiftPM or Xcode Build System) would need to incrementally adopt. In SwiftPM, we can probably move somewhat quickly to adopt it and prove it out, but other clients might not have that freedom. From that sense, the second part is honestly the most important (safety of changes).

I love that you are working on this, by the way, its been something we have wanted for ages!

include/llbuild/Basic/FileInfo.h
```diff
@@ -75,6 +98,8 @@ struct FileInfo {
   uint64_t size;
   /// The modification time of the file.
   FileTimestamp modTime;
+  /// The MD5 hash of the file.
+  LazyMD5Digest digest;
```
dmbryson: Similar to @ddunbar's comment, I think we'll want this to be selectable by the client.  Probably from a layering perspective, we should have a separate digest build value that holds an identifier and a digest value (blob?).  The lazy calculation is a good idea, but likely should be done outside of FileInfo, using a task in the build graph.   That will separate the concerns better and allow the system to be more explicit in when it is calculated.  Right now there is no way for the system to be sure that we don't accidentally calculate the digest of the file at a much later time when the rest of the info has possibly changed.

Given the last point, I think the digest task should probably be formulated such that it outputs two build values, the FileInfo of the file at digest calculation and the Digest value itself.  i.e. `fstat()` the same `fd` we use.  That will make sure we guard against outside changes impacting our view of the external (file system) environment.

drewcrawford:
We're interested in behavior "like this patch" in the xcode buildsystem, in order to do incremental builds in a CI like GitHub Actions.

Relying on the inode for incremental builds implies that a DerivedData copied from "somewhere else" (even the same machine) will do a full rebuild.  As a result, we can't cache actions with [actions/cache](https://github.com/actions/cache) even with quite sophisticated attr-preserving methods.

FB8440337

dmbryson:
You might try `-IgnoreFileSystemDeviceInodeChanges=YES` in your `xcodebuild` arguments.   That should propagate through to llbuild and do what it says, ignore the device and inode fields of stat info when performing checks.

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create pull request with `main` branch.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

PR #642, opened 2020-02-25
[Swift/C-API] Add getFileInfo support

dmbryson:
Allow tasks to defer to llbuild to get file info, specifically for use
in generating custom build values.

rdar://problem/59433735


dmbryson:
@swift-ci please smoke test

PR #643, opened 2020-02-28
Use ArgumentParser for cli command definition

BenchR267:
ArgumentParser is a new library from Apple to parse CLI arguments. It enables llbuild-analyze to define its commands, options and flags in a declarative fashion.
More information here: https://swift.org/blog/argument-parser/

rdar://59846733


BenchR267:
@swift-ci please test

BenchR267:
@swift-ci please smoke test

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -6,95 +6,58 @@
...
+import ArgumentParser
 
 import struct Foundation.Data
-import class Foundation.ListFormatter
 import class Foundation.FileManager
 import class Foundation.JSONEncoder
 
-public final class CriticalPathTool: Tool<CriticalPathTool.Options> {
+struct CriticalPathTool: ParsableCommand {
+    static var configuration = CommandConfiguration(commandName: "critical-path", shouldDisplay: true)
```
natecook1000: You can omit the `shouldDisplay` parameter here.
```suggestion
    static var configuration = CommandConfiguration(commandName: "critical-path")
```

products/llbuild-analyze/Sources/CriticalPathTool.swift
```diff
@@ -6,95 +6,58 @@
...
-            option: parser.add(option: "--graphvizOutput", usage: "Graphviz display options. Possible values: \(Options.GraphvizDisplay.helpDescription). Default is all"),
-            to: { $0.graphvizDisplay = Options.GraphvizDisplay(rawValue: $1) ?? .all }
-        )
-        
-        binder.bind(
-            option: parser.add(option: "--quiet", shortName: "-q", usage: "There will be no output on stdout."),
-            to: { $0.quiet = $1 })
-        
-    }
+    @Option(default: 9)
```
natecook1000: This looks like it needs a couple more parameters:
```suggestion
    @Option(name: [.customLong("clientSchemaVersion"), .customLong("version", withSingleDash: true)], default: 9, help: "The version of the client schema.")
```

BenchR267: Thanks, that makes it 100% compatible. I don't expect infrastructure to rely on this yet but it's still better to keep it compatible.

BenchR267:
@swift-ci please smoke test

.gitignore
```diff
@@ -21,6 +21,10 @@
 /.build
 /Packages
 
+# Commonly used name for build products for llbuild-analyze.
+/products/build
```
ddunbar: Why is this one here?

.gitignore
```diff
@@ -21,6 +21,10 @@
...
+/products/llbuild-analyze/.build
```
ddunbar: Why don't we just change the `/.build` a couple lines up to be `.build`

products/llbuild-analyze/Sources/json.swift
```diff
@@ -0,0 +1,151 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+
+import llbuildAnalysis
+import llbuildSwift
+
+// MARK: - JSON Serialization
```
ddunbar: `json.swift` is a bad name for this file, IMHO

.gitignore
BenchR267: It got created when building llbuild-analyze in the IDE. We could either re-configure that or ignore the `build` dir, I decided to ignore it. Would you suggest to do the former?

.gitignore
BenchR267: I just had a case where a similar situation caused a lot of confusion but I guess it would be fine due to the `.` prefix.

products/llbuild-analyze/Sources/json.swift
BenchR267: Would `JSONSerialization` be better?

ddunbar: I think `PathOutput` would be better, since that is the key top it defines, I think?

jakepetroules: `CriticalBuildPathSerialization`? :D

BenchR267: It just contains the serialization in json format though. We support a different one as well: GraphViz. And we might add more in the future (there's an open task for Chromium tracing for example).

BenchR267:
@swift-ci please smoke test

products/llbuild-analyze/Sources/json.swift
BenchR267: I've renamed it to `PathJSONSerialization` and the GraphViz.swift file to `PathGraphVizSerialization`. I think using `Path` instead of `CriticalPath` in this context is fine since the whole tool is built for finding the critical path - although the serialization is generic and can visualize any path theoretically.

BenchR267:
@swift-ci please smoke test

products/llbuild-analyze/Sources/json.swift
ddunbar: 👍

PR #644, opened 2020-03-08
Call echo with -e in strip-colors.ninja

drexin:
On most systems echo will not interpret backslash escapes unless called with -e

### DIFF
--- a/tests/Ninja/Build/strip-colors.ninja
+++ b/tests/Ninja/Build/strip-colors.ninja
@@ -17,7 +17,7 @@
 #
 
 rule SHOW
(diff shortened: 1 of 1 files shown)


drexin:
I ran into this when trying to build Swift on CentOS 7.

ddunbar:
@swift-ci please smoke test

ddunbar:
Does macOS support -e? If not we might need some lit magic to make a platform specific substitution

drexin:
@ddunbar macOS seems to just ignore it.

drexin:
Meaning the output on macOS is the same, with and without -e

PR #647, opened 2020-03-12
[SwiftNinja] Add a simple API for accessing Ninja manifests from Swift.

ddunbar:
- This is not an elegant implementation (it leverages the JSON support rather
   than bridging to all the C++ APIs), but it is useful for bootstrapping
   experiments with working with Ninja manifests in Swift.

### DIFF
--- a/products/swift-Ninja/NinjaManifest.swift
+++ b/products/swift-Ninja/NinjaManifest.swift
@@ -1,0 +1,118 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import Foundation
+
+/// An individual rule in a Ninja manifest.
+public struct Rule: Codable, Equatable {
+    /// The command to run for this rule, if present.
+    public let command: String?
+
+    /// The description of this rule, if present.
+    public let description: String?
+
+    public init(
+        command: String? = nil,
+        description: String? = nil
+    ) {
+        self.command = command
+        self.description = description
+    }
+}
+
+/// An individual command in a Ninja manifest.
+public struct Command: Codable, Equatable {
+    /// The name of the rule which produced this command.
+    public let rule: String
+
+    /// The list of inputs to the command.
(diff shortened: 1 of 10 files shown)


ddunbar:
This is a WIP, not really ready for full review but open to suggestions on naming or other things. Main thing left to clean up is move some of the copy-paste utility code to a common location.

ddunbar:
Ready now, even with the .xcodeproj changes which make me sad vs the Package.swift changes.

ddunbar:
@swift-ci please smoke test

products/swift-Ninja/NinjaManifest.swift
```diff
@@ -0,0 +1,118 @@
...
+        p.standardOutput = pipe
+        p.launch()
+        var data: Data!
+        let sema = DispatchSemaphore(value: 0)
+        DispatchQueue.global().async {
+            data = pipe.fileHandleForReading.readDataToEndOfFile()
+            sema.signal()
+        }
+        p.waitUntilExit()
+        sema.wait()
```
jakepetroules: Are we certain this version is safe against deadlocks? Lily was helping me with a similar issue earlier...

/CC @millenomi

dmbryson: I'd be very interested to know how this could deadlock.  The calling thread is blocked until process exit.  The pipe reader will continue to read until the pipe is closed, aka process exit.  It isn't waiting on the caller thread for anything.   The caller thread doesn't wait on the semaphore until after the process has exited, but at that point we should reasonably expect that the pipe read will be getting the EOF soon, if it hasn't already.

dmbryson:
@swift-ci please smoke test

products/swift-Ninja/NinjaManifest.swift
ddunbar: Exactly, certainly without the async it can easily deadlock

ddunbar:
I guess those Info.plist files are required, sigh.

jakepetroules:
> I guess those Info.plist files are required, sigh.

If you set `GENERATE_INFOPLIST_FILE=YES`, then you don't have to provide them :)

ddunbar:
@jakepetroules thanks! I thought we had something... is that setting new or XCBuild only?

ddunbar:
@swift-ci please smoke test

jakepetroules:
It's XCBuild only, new since Xcode 11.0.

PR #652, opened 2020-03-21
Backport: Call echo with -e in strip-colors.ninja

drexin:
On most systems echo will not interpret backslash escapes unless called with -e


ddunbar:
@swift-ci please smoke test

PR #653, opened 2020-03-25
[Engine] Ensure build() cleans under early exit

dmbryson:
Under early cancellation and other early exit conditions, build() may
leave the db transaction open or rule scan records allocated. Move those
cleanup actions into scope defers for all exits.

rdar://problem/60856571


dmbryson:
@swift-ci please smoke test

PR #654, opened 2020-03-26
Add Core capabilities to configure a lane base scheduler and use `spawn`

sergiocampama:


products/llbuildSwift/CoreBindings.swift
```diff
@@ -374,7 +420,7 @@ public class BuildEngine {
     /// The number of rules which have been defined.
     public var numRules: Int = 0
 
-    public init(delegate: BuildEngineDelegate) {
+    public init(delegate: BuildEngineDelegate, serial: Bool = false, schedulerAlgorithm: SchedulerAlgorithm = .commandNamePriority, schedulerLanes: UInt32 = 0) {
```
sergiocampama: This is a change in functionlity, since before Core was getting the serial scheduler. At the same time, spawn APIs were not available before, so it won't change how current users process their rules.

products/libllbuild/Core-C-API.cpp
```diff
@@ -240,6 +273,36 @@ void llb_buildengine_task_is_complete(llb_task_interface_t* ti_p,
...
+  }
+
+  virtual void getVerboseDescription(SmallVectorImpl<char> &result) const {
+      llvm::raw_svector_ostream(result) << getOrdinalName();
+  }
+};
+
+void llb_buildengine_spawn(llb_task_interface_t* ti_p, void* context, void (*action)(void* context, llb_task_interface_t* ti_p)) {
+    auto ti = ((TaskInterface*) ti_p);
+    auto ti_cp = TaskInterface(*ti);
```
sergiocampama: this line right here...  it will haunt me forever, I hate C++

dmbryson: Ah but the grass on the other side could be something like:

```
ti_p.withMemoryRebound(to: UnsafePointer<TaskInterface>.self, capacity: 1) { ti in
  let ti_cp = TaskInterface(copying: ti)
  ...
}
```

dmbryson:
@swift-ci please smoke test

sergiocampama:
@swift-ci please smoke test

PR #655, opened 2020-03-29
[Package] Add a potentially necessary linked library on Linux.

ddunbar:
- In some cases, the LLVM support library ends up needing to be linked against
   libm.


ddunbar:
This only impacts the package build, which isn't currently covered in tests... bypassing merge. I have independently tested this on an impact Linux system to validate the fix.

PR #658, opened 2020-04-04
Fix NinjaManifest when used in xctest on linux

dmbryson:


dmbryson:
Not tested by swift-ci, bypassing.

PR #659, opened 2020-04-06
Add 'Evo' engine API evolution experiment

dmbryson:
Evo is an exploration of a new engine <-> task/rule API. It actually
predates and informed some of the new TaskInterface APIs. The current
implementation doesn't scale particularly well, as it burns
a thread per executing rule. However, it is worthwhile having in the
codebase for a bit as we consider future directions.


dmbryson:
@swift-ci please smoke test

dmbryson:

PR #660, opened 2020-04-07
Add experimental cevobuild package

dmbryson:
`cevobuild` is an exploration of evolving concepts in the core build
engine of llbuild.


experimental/cevobuild/Tests/CevoCoreTests/XCTestManifests.swift
```diff
@@ -0,0 +1,18 @@
+#if !canImport(ObjectiveC)
```
aciidb0mb3r, line 1: You can kill this and use `swift test --enable-test-discovery`

dmbryson: Sweet!

PR #663, opened 2020-04-19
[Unittests] Fix lit integration of googletest.

ddunbar:
- It turns out that #532 refactored the output location of unit tests binaries,
   which broke the lit integration with googletest. Thus, all our unit tests
   haven't been running for a couple months 🤦‍♂️.


ddunbar:
@swift-ci please smoke test

compnerd:
Wow, thanks for fixing that.  I wonder if it makes sense to consider something like lit.site.cfg that is preprocessed to ensure that whatever information about the build tree is needed is passed from CMake to lit.

ddunbar:
We do have a lit site cfg, I probably just don’t know how to ask cmake for that info

PR #664, opened 2020-04-19
[CAS] Stub out initial CASDatabase.

ddunbar:

PR #665, opened 2020-04-20
[CAS] Add DenseMapInfo support for DataID.

ddunbar:
* [ ] Depends on #664

### DIFF
--- a/include/llbuild/CAS/DataID.h
+++ b/include/llbuild/CAS/DataID.h
@@ -13,6 +13,7 @@
 #ifndef LLBUILD_CAS_DATAID_H
 #define LLBUILD_CAS_DATAID_H
 
+#include "llvm/ADT/DenseMapInfo.h"
 #include "llvm/ADT/StringRef.h"
 
 #include <cstdint>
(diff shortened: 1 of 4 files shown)


ddunbar:
@swift-ci please smoke test

include/llbuild/CAS/CASDatabase.h
```diff
@@ -0,0 +1,70 @@
...
+  /// Check if the database contains the given.
+  virtual auto contains(const DataID& id) ->
+    std::future<llvm::ErrorOr<bool>> = 0;
+
+  /// Get the given object, if present.
+  virtual auto get(const DataID& id) ->
+    std::future<llvm::ErrorOr<std::unique_ptr<CASObject>>> = 0;
+
+  /// Write the given object into the database.
+  virtual auto put(std::unique_ptr<CASObject> object) ->
```
dmbryson, line 53: Reference?   That way the client doesn't have to relinquish its copy of the object on put.

include/llbuild/CAS/CASDatabase.h
```diff
@@ -0,0 +1,70 @@
...
+};
+
+/// An individual object in the CAS.
+struct CASObject {
+public:
+  /// The list of references in this object.
+  llvm::SmallVector<DataID, 8> refs;
+
+  /// The data in this object.
+  llvm::SmallVector<uint8_t, 256> data;
```
dmbryson, line 64: /me is not intimately familiar with llvm::SmallVector.  Is this hard capping the size at 256 bytes, or is that just the size under which it will perform most optimally?

include/llbuild/CAS/DataID.h
```diff
@@ -0,0 +1,93 @@
...
+
+private:
+  /// Used for sentinels.
+  DataID(uint8_t size, char markerByte) : size(size) {
+    id[0] = markerByte;
+  }
+  
+public:
+  /// Create a DataID from the given string.
+  DataID(llvm::StringRef id_) {
```
dmbryson, line 42: explicit?

dmbryson:
Oops, left my comments on the wrong PR.

include/llbuild/CAS/CASDatabase.h
ddunbar: The big problem with that is that any implementation which wants to be async is going to have to copy the data, which might be a double copy if the client had to create an owning copy anyway?

This is part of what I meant with being a not very well though out API.

What I think we really want here is some kind of optionally-owning vector, like an ArrayRef but which can be a unique owner. I didn't want to write that, so I went with this for now.

Thoughts?

include/llbuild/CAS/CASDatabase.h
ddunbar: It is the size which is allocated inline with the SmallVector itself, as opposed to needing a separate heap allocation.

I'm not really sure this is a good use case for SmallVector, as above I think we want a maybe owning data reference, but didn't want to make one.

include/llbuild/CAS/DataID.h
ddunbar: indeed, thanks

ddunbar:
@swift-ci please smoke test

ddunbar:
@swift-ci please smoke test

PR #667, opened 2020-04-24
Remove iOS as supported platform

BenchR267:
llbuild-analyze is not meant to be run on iOS, it's an executable target to run on macOS for analyzing llbuild databases.
rdar://problem/62300693

### DIFF
--- a/products/llbuild-analyze/Package.swift
+++ b/products/llbuild-analyze/Package.swift
@@ -8,7 +8,7 @@ import PackageDescription
 let package = Package(
     name: "llbuild-analyze",
     platforms: [
-        .macOS(.v10_10), .iOS(.v9),
+        .macOS(.v10_10)
     ],
     products: [
         .executable(


BenchR267:
@swift-ci please smoke-test

BenchR267:
@swift-ci please smoke test

jakepetroules:
What's driving this? All it does is change the deployment target from iOS 9 to iOS 8, SwiftPM doesn't have any notion of supporting or not supporting a particular platform.

BenchR267:
Ha, that is actually news to me. I thought that every platform that is not specified is not supported. Isn't that what the `.all` case is for?
Anyway, it doesn't have any effect but this tool is written to be run on macOS, therefore it's technically incorrect to specify iOS as a supported platform in the package manifest.

jakepetroules:
> Ha, that is actually news to me. I thought that every platform that is not specified is not supported. Isn't that what the .all case is for?

Swift packages always support all platforms; there's currently no way to opt out (and there's no `.all` case).

What the `platforms` property does at the moment is allow you to control the deployment target for that platform; if left unspecified, the default is used, which is:

- macOS: 10.10
- iOS: 8.0
- tvOS: 9.0
- watchOS: 2.0

PR #671, opened 2020-05-11
Adjust TaskInferface passing for easier use

dmbryson:
Passing a reference to a stack allocated TaskInterface object complicated safe usage of it in clients.  In particular, the Swift/C-API interface made it unsafe to perform work on another thread. This changes to pass-by-value semantics for the small, two pointer class/struct, enabling simpler/safer use of it on other threads.

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -360,7 +360,7 @@ std::unique_ptr<basic::ExecutionQueue> BuildSystemEngineDelegate::createExecutio
 
 #pragma mark - Task implementations
 
-static BuildSystemImpl& getBuildSystem(TaskInterface& ti) {
+static BuildSystemImpl& getBuildSystem(TaskInterface ti) {
   return static_cast<BuildSystemEngineDelegate*>(ti.delegate())->getBuildSystem();
 }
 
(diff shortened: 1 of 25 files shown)


dmbryson:
@swift-ci please smoke test

dmbryson:

dmbryson:
sigh `/home/buildnode/jenkins/workspace/swift-llbuild-PR-Linux-smoke-test/swift-corelibs-foundation/CoreFoundation/String.subproj/CFString.c:3552:100: error: use of undeclared identifier 'UCHAR_EMOJI_MODIFIER'`

dmbryson:

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -650,18 +650,18 @@ class DirectoryStructureInputNodeTask : public Task {
...
                                  const ValueType& value) override {
   }
 
-  virtual void provideValue(TaskInterface&, uintptr_t inputID,
+  virtual void provideValue(TaskInterface, uintptr_t inputID,
                             const ValueType& value) override {
   }
 
-  virtual void inputsAvailable(TaskInterface& ti) override {
+  virtual void inputsAvailable(TaskInterface ti) override {
```
ddunbar, line 664: Was this problem also an issue in C++? How come it never came up?

dmbryson: Basically all of the internal  C++ uses of it where we actually hop threads used lambdas and were implicitly capturing copies.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

dmbryson:

PR #672, opened 2020-05-20
[Core] Record dependencies in order requested

dmbryson:
This changes the core engine processing order to preserve the order in which tasks/the client have requested input dependencies. This ensures that subsequent builds scan those dependencies in a similar ordering.

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -28,6 +28,7 @@
 #include <cassert>
 #include <cstdio>
 #include <condition_variable>
+#include <deque>
 #include <memory>
 #include <mutex>
 #include <thread>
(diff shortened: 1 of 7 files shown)


dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
```diff
@@ -109,9 +110,13 @@ class BuildEngineImpl : public BuildDBDelegate {
     /// Force the use of a prior value
     bool forcePriorValue = false;
   };
-  std::vector<TaskInputRequest> inputRequests;
+  std::deque<TaskInputRequest> inputRequests;
```
ddunbar, line 113: Should we worry about this having a perf impact?

dmbryson: Existing perf tests show +/- 3-5%, but those can go in either direction in repeated test runs.

dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
ddunbar: Cool, thanks for checking.

dmbryson:
@swift-ci please smoke test

PR #673, opened 2020-06-04
Enhance fixup target to use plutil and get correct xctest path

edymtt:
There are two reason to use plutil
* `defaults` will no longer support this in the future, as per man page

> WARNING: The defaults command will be changed in an upcoming major
release to only operate on preferences domains. General plist
manipulation utilities will be folded into a different command-line
program.
* `plutil` prints more meaningful errors that help in troubleshooting CI issues

Indeed this switch helped uncover an issue with the logic inferring the path of `XcodePerfTests.xctest`.

Addresses rdar://63984731

### DIFF
--- a/perftests/Xcode/PerfTests/CMakeLists.txt
+++ b/perftests/Xcode/PerfTests/CMakeLists.txt
@@ -43,7 +43,7 @@ target_link_libraries(XcodePerfTests PRIVATE
 #
 # http://www.cmake.org/Bug/view.php?id=15485
 add_custom_command(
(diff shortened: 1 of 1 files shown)


edymtt:
@swift-ci please smoke test

edymtt:
Thanks for catching my typo 😄

edymtt:
@swift-ci please smoke test

edymtt:
Rerequesting reviews since the PR scope is slightly expanded now.

edymtt:
@swift-ci please smoke test

PR #674, opened 2020-06-10
[5.3] Enhance fixup target to use plutil and get correct xctest path

edymtt:
There are two reason to use plutil

* defaults will no longer support this in the future, as per man page

> WARNING: The defaults command will be changed in an upcoming major
release to only operate on preferences domains. General plist
manipulation utilities will be folded into a different command-line
program.

* plutil prints more meaningful errors that help in troubleshooting CI
issues

Indeed this switch helped uncover an issue with the logic inferring the
path of XcodePerfTests.xctest.

Addresses rdar://63984731

(cherry picked from commit 145c113cde37318f3cef38843f9ce6a678aaa4fd,  #673)


edymtt:
@swift-ci please smoke test

edymtt:
Please test with following pull request:
https://github.com/apple/swift/pull/32316

@swift-ci Please smoke test Linux platform

PR #677, opened 2020-07-29
Fix perf test ProcessAttributes initializer

dmbryson:
rdar://problem/66291052


dmbryson:
@swift-ci please smoke test

PR #681, opened 2020-08-18
build: define `libllbuild_EXPORTS` on Windows

compnerd:
This is required to decorate the public symbols with
`__declspec(dllexport)` as we do not currently support static linking on
Windows with SwiftPM.


compnerd:
@swift-ci please smoke test

compnerd:
CC: @dmbryson @tomerd

PR #684, opened 2020-08-19
Disable ninja multiple-outputs test temporarily

dmbryson:
rdar://problem/67415097


dmbryson:
@swift-ci please smoke test

PR #689, opened 2020-08-27
Base the install paths of targets on DEVELOPER_INSTALL_DIR

jakepetroules:
This is to simplify some other build configuration elsewhere.

Related: rdar://67568955


jakepetroules:
@swift-ci please smoke test

PR #690, opened 2020-08-27
Remove doc comment for non-existent function param

dmbryson:
rdar://problem/67817536


dmbryson:
@swift-ci please smoke test

PR #692, opened 2020-08-28
[llbuildSwift/BuildKey] Fix memory leak coming from `BuildKey.construct(data:)`

akyrtzi:
Inside the `BuildKey.construct(data:)` function, `llb_build_key_make` allocates a new `CAPIBuildKey` object which gets leaked because of `weakPointer` set to true.


akyrtzi:
@swift-ci smoke test

PR #694, opened 2020-08-29
[BuildSystem-C-API] Fix memory leak of `CAPIBuildKey` object from `CAPITool.createCustomCommand()`

akyrtzi:


akyrtzi:
@swift-ci smoke test

PR #670, opened 2020-05-06
[CMake][llbuildSwift] fix runpath for ELF platforms

buttaface:
Remove the absolute path to the host toolchain's stdlib from libllbuildSwift.so.

Otherwise, you see the following in the latest official 5.3 snapshot release for linux:
```
> readelf -d swift-5.3-DEVELOPMENT-SNAPSHOT-2020-05-04-a-ubuntu18.04/usr/lib/swift/pm/llbuild/libllbuildSwift.so | ag runpath
0x000000000000001d (RUNPATH)            Library runpath: [/home/buildnode/jenkins/workspace/oss-swift-5.3-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:$ORIGIN:$ORIGIN/../../linux]
```

PR #695, opened 2020-08-29
[5.3][CMake][llbuildSwift] fix runpath for ELF platforms

buttaface:
Remove the absolute path to the host toolchain's stdlib from libllbuildSwift.so.

It has been suggested to me by some Swift devs that this would be good to get in before the 5.3 release, a cherry-pick from #670.

This should be completely safe for the 5.3 branch, as it's just removing the build toolchain's stdlib rpath, which should be unused by this library.


dmbryson:
@swift-ci please smoke test

buttaface:
Thanks for the quick review.

PR #696, opened 2020-09-01
[Python3] Update scripts to support Python3

tbkka:
Most of the changes here just convert `print` statements (which are no longer supported in Python3) to `print()` function calls (which exists in both Python 2 and 3).

The only other change of note is to strip the `o` character from the output of `oct()`.  In Python 3, `oct()` outputs octal values with a leading `0o` as 0o123 and these scripts expect C-style octal format `0123` instead.


dmbryson:
@swift-ci please smoke test

PR #698, opened 2020-09-07
LLVMSupport: make `HandleAbort` static

compnerd:
When building SourceKit-LSP on Windows with swift-package-manager, we
end up with multiple definitions of `HandleAbort` due to multiple copies
of `LLVMSupport` being used due to the shared linking model of
swift-package-manager.  Declare the function as static to make it be
COMDATed.

### DIFF
--- a/lib/llvm/Support/Windows/Signals.inc
+++ b/lib/llvm/Support/Windows/Signals.inc
@@ -404,11 +404,19 @@ AvoidMessageBoxHook(int ReportType, char *Message, int *Return) {
 
 #endif
 
-extern "C" void HandleAbort(int Sig) {
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

dmbryson (approved):
Have you considered upstreaming this change to llvm as well?    Though infrequent, we do update these files from time to time.  This change might be lost if it is not changed at the source.

compnerd:
AIUI, this only impacts the build with s-p-m.  I'll try to see if I can figure out the difference between cmake and s-p-m so we understand exactly what's going on.

compnerd:
I'll look into the upstreaming aspect of this separately I think.  Lets get this merged since this should make it easier for others to start contributing to the efforts for improving Windows support.

PR #701, opened 2020-09-11
Handle interrupts properly on Windows

egorzhdan:
Currently SwiftPM ignores Ctrl-C on Windows, this change makes it stop the build process when Ctrl-C is pressed


egorzhdan:
cc @compnerd

compnerd:
Yeah, I think it makes sense for the processes to be a single process group.

compnerd:
@swift-ci please smoke test

compnerd (approved):

PR #702, opened 2020-09-19
llbuild should disable Swift Objective-C compatibility headers

jakepetroules:
llbuild has no @objc APIs, so the generated header is not needed.

<rdar://69185419>


jakepetroules:
@swift-ci please smoke test

dmbryson (approved):

PR #706, opened 2020-10-14
llbuildSwift: replace unnecessary import of `MSVCRT`

compnerd:
Replace the use of `MSVCRT` with `ucrt` as only the standard C
interfaces are needed.  This is in preparation to move away from
`MSVCRT`.


compnerd:
@swift-ci please smoke test

dmbryson (approved):

PR #707, opened 2020-11-04
Fix a warning that occurs when building with -Wdocumentation

abertelrud:
This is currently the only warning when building swift-package-manager using its bootstrap script.

Since this is a declaration in a source file and not a header, it seemed reasonable to just remove it.  The include/llbuild/Basic/FileInfo.h header already has the proper comments.


abertelrud:
@swift-ci please smoke test

dmbryson (approved):

PR #712, opened 2021-01-07
[Test] Relax ordering assumption in max-commands-progress.ninja

edymtt:
While most of the times `output-2` is copied before `output-1`, in some
rare cases the opposite can happen (which seems reasonable since there
is no ordering requirement between the two).

Addresses rdar://72898516


edymtt:
@swift-ci please smoke test

edymtt:
My change as is would consider the following valid

```
[1/3] cp input-1 output-1
[1/3] cp input-2 output-2
```

but I think the main focus of those checks is comparing the ordering between the `cp` command and the `cat` one

edymtt:

edymtt:
@swift-ci please smoke test

dmbryson (approved):

jakepetroules:
Thanks!

PR #720, opened 2021-03-25
Fix the llbuild tests that are broken/flaky and blocking CI

jakepetroules:
BuildDBBindingsTests: Swift variables are not guaranteed to live to the
end of their declarating scope; extend the lifetime of the database
object in a couple of the test methods to ensure they are not
prematurely deallocated.

buildsystem-capi.llbuild: macOS uses DYLD_LIBRARY_PATH,
not LD_LIBRARY_PATH. Pass the proper environment variable to the test
so that libllbuild.dylib can be found. This only manifested locally, not
on the CI, for some reason.

incremental.ninja: relax the ordering assumption. This is the same fix
done for the similar max-commands-progress.ninja test case back in
f339c9e115816b72bafad1b8a2927a6eccca66b5.


jakepetroules:
@swift-ci please smoke test

dmbryson (approved):

jakepetroules:
@swift-ci please smoke test

PR #726, opened 2021-05-17
Fix the llbuild umbrella header

jakepetroules:
This was changed in #538 without a clear reason, and it breaks
framework-style includes in some cases. Return to the original style.


jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please smoke test

dmbryson (approved):

PR #727, opened 2021-05-21
Scope build debugging trace file to the build operation

jakepetroules:
Clients who cache and re-use a BuildSystem instance to serve multiple subsequent build operations are unable to get a trace file for multiple builds.

This is because the trace file pointer is scoped to and kept open for the lifetime of the BuildSystem instance, rather than being scoped to the build operation.

This change scopes the trace file writing to the build operation. This means that multiple build operations by the same BuildSystem will no longer be able to "append" to the same trace file (e.g. having multiple build-started/build-ended pairs), but this responsibility can be shifted to clients who can preserve the file between builds if desired.

rdar://47133775

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -81,6 +81,9 @@ class BuildEngineImpl : public BuildDBDelegate {
   /// The tracing implementation, if enabled.
   std::unique_ptr<BuildEngineTrace> trace;
 
+  /// Path of the trace file to write to, if set.
+  std::string traceFile;
+
   /// Mutex for access to execution queue.
   std::mutex executionQueueMutex;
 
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci please smoke test

lib/Core/BuildEngine.cpp
```diff
@@ -1534,6 +1532,23 @@ class BuildEngineImpl : public BuildDBDelegate {
...
+
+      this->trace = std::move(trace);
+    }
+
+    llbuild_defer {
+      std::string error;
+      if (!trace->close(&error)) {
+        delegate.error(error);
+      }
+    };
```
dmbryson, line 1550: Should we be clearing out/storing nullptr to `this->trace` here too?

dmbryson (approved):

lib/Core/BuildEngine.cpp
jakepetroules: Good point; done!

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please smoke test

PR #730, opened 2021-06-17
Try to fix non-deterministic output in test

dmbryson:
Cherry-pick the fix for rdar://77849636

rdar://79468954


varungandhi-apple:
LGTM, lmk when this is ready to merge. Running CI won't work right now because of issues with fetching from GitHub on macOS.

PR #731, opened 2021-06-29
Add a C API and corresponding Swift interface for the Ninja component

bnbarham:
There's currently 2 commits in this PR and I'm happy to split them if you'd prefer - just thought it'd be worth giving the whole context initially:
  - A refactor to use MemoryBuffer and StringRef rather than passing unique_ptr, string, const char *, length, etc. This started off from `ManifestLoaderActions`, so isn't *everything*, but does refactor a significant chunk.
  - The addition of a C API (`products/libllbuild/include/llbuild/ninja.h`) for the Ninja component and modifications to `NinjaManifest.swift` to use it instead of executing a llbuild process.

I went back and forth between a couple different C API implementations, so quite interested on any thoughts there. It currently trades a bit of allocation for a simpler API with actual types, rather than handing back only opaque types and having functions for all the details.

I've also renamed what was `Command` in the Ninja component to `BuildStatement`, which I believe matches up with the actual Ninja documentation a little better (also makes it easier not to conflate with llbuild's `Command`).


bnbarham:
@swift-ci please test

ddunbar:
@dmbryson ?

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

tests/Ninja/Loader/include.ninja
```diff
@@ -24,7 +24,7 @@ build b2: testrule input ${input2_var} xx${should_be_empty}xx
 
 # Check that we diagnose broken includes.
 #
-# CHECK-ERR: include.ninja:[[@LINE+1]]:8: error: unable to open input
+# CHECK-ERR: include.ninja:[[@LINE+1]]:8: error: unable to read input
```
bnbarham, line 27: Is the exact error important here? The open/read aren't separate any more (just uses MemoryBuffer::getFile) and "read" seemed the more generic of the two.

dmbryson: “read” seems reasonable to me.

dmbryson (approved):

bnbarham:
Tests didn't start for some reason?

@swift-ci please test

dmbryson:
@swift-ci please smoke test

bnbarham:
I'm not sure why the bin/CoreTests build doesn't fail on macOS 🤔 - I have a `emplace_back(someString, 0)` for a vector whose elements are `std::pair<std::string, std::vector<std::string>>`. So the 0 there is definitely wrong... 🤷. I've fixed that and the warning about unused variables in `NinjaCommand.cpp` (missed removing them).

@swift-ci please test

dmbryson:
@swift-ci please smoke test

bnbarham:
I hadn't added ninja.h to the llbuild header group in the Xcode project, done now.

@swift-ci please smoke test

products/swift-Ninja/NinjaManifest.swift
```diff
@@ -1,169 +1,154 @@
...
-    public let command: String
-
-    /// The description of the string, or empty.
-    public let description: String
-
-    /// Whether or not the command is a generator.
-    public var generator: Bool {
-        return _generator ?? false
+    if cManifest.error.length > 0 {
+      self.error = ownedString(cManifest.error)
```
akyrtzi, line 83: Should this part throw an error instead? And is there any concern with continuing even after an error?

bnbarham: My thought was that it depends on the client and that it may still be useful to have the rest of the manifest, even if there were errors. Ie. maybe they only care about a particular build statement.

I considered adding a `throwOnError` argument, but then errors would have to be handled regardless since the signature would be `throws`. Perhaps that's not too bad since `try!` could be used, thoughts?

akyrtzi: > it may still be useful to have the rest of the manifest, even if there were errors. Ie. maybe they only care about a particular build statement.

Maybe have a separate creation function for this? I feel that the initializer that will be commonly used should just throw the error (that would be "Swifty") and if a client has a use case to try to get info after errors then it could use another function.

bnbarham: Sounds good, thanks!

bnbarham:
So I had been working off of Package.swift rather than the xcodeproject. In fixing it up, I've realised that having Ninja-C-API.cpp in libllbuild and then a separate Swift Ninja framework is a little odd.

There's a couple options here:
1. Move NinjaManifest.swift into llbuildSwift, have it be part of the llbuild Swift interface, and remove the various Ninja targets
2. Create another lib for the Ninja C API, remove it from libllbuild, and keep the separate Ninja Swift lib

I've already done (1) in my local branch, but thought I'd check what we actually want first. @ddunbar was the separate library just to make it clear that it wasn't using the llbuild API, or did you actually want it as a separate library?

bnbarham:
I went ahead and pushed the changes for (1) above.

Also note that I've made `llbuildTestSupport` a `target` rather than `testTarget` as the former doesn't work as a dependency from what I can tell. Though `llbuildTestSupport` is technically not needed any more as the NinjaManifest interface is now in `llbuild`, so we could remove it entirely and move the extension into `llbuildSwiftTests`.

bnbarham:
@swift-ci please smoke test

bnbarham:
All tests pass now :)

@dmbryson are you happy with the latest changes? Ie. `NinjaManifest` is now a part of `llbuildSwift` rather than its own `Ninja` target.

jakepetroules:
This broke llbuild2, which I've fixed in apple/swift-llbuild2#125.

bnbarham:
> This broke llbuild2, which I've fixed in [apple/swift-llbuild2#125](https://github.com/apple/swift-llbuild2/pull/125).

@jakepetroules sorry about that (and thanks for fixing) - I knew llbuild2 was using the Ninja API, but I had thought it wouldn't require changes until updating llbuild2 since `Package.resolved` specified `53d12bc6e70e4ae6c0dec634fd48ec0cfff17f2d` (ie. a commit before these changes). Does `Package.resolved` not actually lock to a specific version?

jakepetroules:
Yes, Package.resolved locks a specific version, but that doesn't apply if llbuild2 is not itself the root package (I used `swift package update` on a package that depended on llbuild2). So the version constraints were such that the package dependency graph still allowed incompatible versions to be used with each other.

bnbarham:
Thanks, makes sense 🙇

PR #733, opened 2021-07-09
Suffix for building on OpenBSD.

3405691582:
sqlite3 is required but is not part of the base system. This means the
usual `-I/usr/local/include` and `-L/usr/local/lib` flags need to be
supplied on this platform.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -228,3 +228,16 @@ package.targets.first{ $0.name == "libllbuild" }?.cxxSettings = [
 ]
 
 #endif
+
+// FIXME: when the SupportedPlatforms availability directive is updated and
+// the platform port is in sync with this directive, these conditions can
+// be folded up with .when(platforms:_) clauses.
+#if os(OpenBSD)
+if let target = package.targets.first(where: { $0.name == "llbuildCore"}) {
+    target.cSettings = [.unsafeFlags(["-I/usr/local/include"])] 
+    target.linkerSettings = [
+        .linkedLibrary("sqlite3"),
+        .unsafeFlags(["-L/usr/local/lib"])
+    ]
+}
+#endif


3405691582:
ping.

dmbryson:
@swift-ci please smoke test

dmbryson (approved):

jakepetroules:
Does OpenBSD not have pkg-config? This shouldn't be necessary (SwiftPM can pick up the dependency via pkg-config), and  we should aim to avoid unsafe flags.

3405691582:
We do have pkg-config, but it doesn't seem obvious what part of the API we need to address to get swiftpm to look at the right pc file. What exists looks like if we want to create an entire wrapper package for sqlite3 (which seems extraneous given for other platforms we don't actually seem to need to do this) or if sqlite3 was part of the package (which it isn't).

So I'm not very fluent in swiftpm but I'm not exactly sure if swiftpm can actually do what we want here? We just want to add the alternate search paths, and swiftpm doesn't seem to provide affordances to do that at first glance.

jakepetroules:
> Are you suggesting the resolution of the pc file should work out of the box?

It should, yes. Where is the pkg-config binary located on OpenBSD, and is that directory in the PATH? If it's in a non-standard location, that may be why SwiftPM is not finding it..

You can see how SwiftPM invokes pkg-config to locate the .pc search paths here: https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCUtility/PkgConfig.swift#L57

jakepetroules:
> Are you suggesting the resolution of the pc file should work out of the box?

If the sqlite3 dependency is expressed as a `systemLibrary` target, then it should, yes. However we're presently just linking to the library directly.

You can see how SwiftPM invokes pkg-config to locate the .pc search paths here: https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCUtility/PkgConfig.swift#L57 If the pkg-config binary is in the PATH, then SwiftPM should find the right linker flags and header search paths to pass, in conjunction with specifying the dependency as a system library target.

3405691582:
So it sounds like I need to commit a modulemap in the project for sqlite3 (because there isn't one on the platform), append a sqlite3 `systemLibrary` target to the target list and patch the deps on `llbuildCore` to include this library, correct?

3405691582:
Unfortunately, doing that means you have to update the deps, which means some of the pkg-config flags have to be included in the parent targets in the dependency tree, which causes build problems. 

Specifically, we get `-I /usr/local/include` high up in the flag list for `swift-build-tool` because it includes the llbuildCore modulemap first, then the `sqlite3` modulemap, `-I /usr/local/include`, then everything else (llbuildBasic, llvmSupport, llvmDemangle, etc). This means if the system has the llvm package installed (needed for certain binutils used elsewhere in the Swift build), the llvmSupport libs will conflict with those installed on the system.

If the sqlite3 dep gets pushed to the end, that all ends up working okay, but we really just want to isolate the sqlite3 pkg-config flags to _just_ the llvmCore target, and we can't do that...

PR #734, opened 2021-07-22
Invoke subclasses of ExternalCommand unfiltered

BenchR267:
Before this change, subclasses of ExternalCommand were only invoked if the build value has been produced by a command (success or failure). With this change, the subclass gets invoked for all build values to keep track of its dependencies. It can still distinguish between the different inputs by looking for the requested inputID.

rdar://57125025

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -1896,11 +1896,7 @@ public:
       BuildSystem&,
       TaskInterface,
       uintptr_t inputID,
-      const BuildValue& value) override {
-    // Should never get here, since we're not requesting inputs in start.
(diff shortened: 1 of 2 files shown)


BenchR267:
@dmbryson I know there's a part missing of our discussion about this (the boolean return of provideValueExternalCommand). I think it's

a) not necessary since the existing logic still holds for external commands
b) adding more complexity due to the breakage in API in the C and Swift interfaces. We would need to add a new version with the return value and deprecate the old one to remove it later. That's a lot of (imo) unneeded complexity given a).

Looking forward to your thoughts on this though!

BenchR267:
@swift-ci please smoke test

dmbryson (approved):

PR #738, opened 2021-08-10
Increase version to 12

BenchR267:
With the behavior changes in https://github.com/apple/swift-llbuild/pull/734 clients need a way to make assumptions on how llbuild will call into ExternalCommands. Increasing the API gives that evidence.

rdar://57125025


BenchR267:
@swift-ci please smoke test

jakepetroules (approved):

BenchR267:
@swift-ci please smoke test

PR #739, opened 2021-08-30
Add high priority support for job queue

dmbryson:
rdar://44684511

### DIFF
--- a/include/llbuild/Basic/ExecutionQueue.h
+++ b/include/llbuild/Basic/ExecutionQueue.h
@@ -77,6 +77,11 @@ namespace llbuild {
       void execute(QueueJobContext* context) { work(context); }
     };
 
+    enum QueueJobPriority {
+      Normal,
+      High
+    };
+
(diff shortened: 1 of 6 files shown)


dmbryson:
@swift-ci please smoke test

include/llbuild/Basic/ExecutionQueue.h
```diff
@@ -77,6 +77,11 @@ namespace llbuild {
       void execute(QueueJobContext* context) { work(context); }
     };
 
+    enum QueueJobPriority {
```
akyrtzi, line 80: Just a suggestion, for improved code readability it would be better to make that an `enum class` so at use site you'd use `basic::QueueJobPriority::High` instead of `basic::High`.
`Hight` and `Normal` as names do not have enough context on their own.

dmbryson: Ah yes, my modern c++ is rusty.  So much swift these days.

PR #741, opened 2021-10-09
Add llbuildNinja to list of llvm targets

pcbeard:
Since swift-package-manager depends on this target, it needs to have LLVM_ON_WIN32 defined in the Windows build.

Define llvmTargets as a set to avoid O(n^2) search of package.targets.

This is a fix for SR-15295.

I tested these changes on Windows, by turning on the environment variable `SWIFTCI_USE_LOCAL_DEPS=1`, cloning local copies of:  swift-tools-support-core, swift-argument-parser, swift-driver, swift-crypto, and Yams, and checking out the necessary tags (see below). Since there seems to be a bug in SPM in the 5.5 toolchain, when processing relative paths, I changed all of the ../ relative paths to S:/ absolute Windows paths, and this made everything build.

Here were the tags/branches I used:

swift-tools-support-core:  .branch("main")
swift-argument-parser: .upToNextMinor(from: "0.4.3")
swift-driver: .branch("main")
swift-crypto: .upToNextMinor(from: "1.1.4")
Yams:  .upToNextMinor(from: "4.0.0")

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -198,10 +198,11 @@ let package = Package(
 // FIXME: Conditionalize these flags since SwiftPM 5.3 and earlier will crash for platforms they don't know about.
 #if os(Windows)
 
-["llvmSupport", "llvmDemangle", "llbuildBuildSystem", "llbuildBasic"].forEach { target in
-    package.targets.first(where: { $0.name == target })?.cxxSettings = [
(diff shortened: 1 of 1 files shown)


pcbeard:
Reviewers:  are the `.when(platforms: [.windows])` qualifiers really needed? These are only being defined if `os(Windows)` is true.

pcbeard:
👍

PR #744, opened 2021-11-01
Implement update status for rules representing custom tasks

BenchR267:
Also make 'Checking Swift Compiler Version' not show its status (SwiftPM expects it that way).

rdar://82651879


BenchR267:
@swift-ci please test

jakepetroules (approved):

PR #745, opened 2021-11-01
Allow clients to indicate that createCommand has failed

jakepetroules:
Some clients may not be able to create commands on-demand and may look
them up in an existing store when requested, which may not have said
command.

rdar://55989748
(cherry picked from commit 47445303f4eeb81183ae64bc9b02413154100a56)


jakepetroules:
@swift-ci please test

PR #749, opened 2021-11-11
[Xcode] Set an explicit iOS deployment target for llbuild.

cwakamo:
iOS 9.0 is the oldest version supported by Xcode 13, so this commit sets the deployment target to that.

In support of this, added a missing availability check since iOS now back-deploys instead of deploying to the latest OS only.

This is for <rdar://problem/84568181>.


cwakamo:
@swift-ci Please smoke test

include/llbuild/Basic/Tracing.h
```diff
@@ -47,17 +47,17 @@ static os_signpost_id_t signpost_id() {
 
 /// Begin an interval if tracing is enabled.
 #define LLBUILD_TRACE_INTERVAL_BEGIN(name, ...) { \
-if (__builtin_available(macOS 10.14, *)) os_signpost_interval_begin(getLog(), signpost_id(), name, ##__VA_ARGS__); \
+if (__builtin_available(macOS 10.14, iOS 12.0, *)) os_signpost_interval_begin(getLog(), signpost_id(), name, ##__VA_ARGS__); \
```
jakepetroules, line 50: Can you add tvOS/watchOS to these three as well? There's also a __builtin_available in adjust-times.cpp that could use adjustment.

jakepetroules (changes requested):

include/llbuild/Basic/Tracing.h
cwakamo: OK, can do -- I'll also set TVOS_DEPLOYMENT_TARGET to 9.0 and WATCHOS_DEPLOYMENT_TARGET to 2.0 to match the iOS deployment target when I fix this.

cwakamo:
@swift-ci Please smoke test

jakepetroules (approved):

PR #750, opened 2021-11-15
[Xcode] Updated the "Install in CommandLineTools" shell script build …

cwakamo:
…phase to not do anything meaningful if `COMMANDLINETOOLS_OUTPUT_DIR` is unset.

`COMMANDLINETOOLS_OUTPUT_DIR` is a new build setting indicating where in the CommandLineTools package llbuild.framework should be installed.
This is configured in Common.xcconfig, allowing clients to customize this location via CommonOverrides.xcconfig.
Additionally, if set to the empty string, the "Install in CommandLineTools" build phase will effectively do nothing.

As part of this, marked the "Install in CommandLineTools" build phase so it only runs during install builds.

This is for <rdar://problem/84568181>.


cwakamo:
@swift-ci Please smoke test

cwakamo:
@swift-ci Please smoke test

jakepetroules (approved):

PR #751, opened 2021-11-24
On Windows, use 2>NUL to ignore stderr output

pcbeard:
This fixes a nagging bug on Windows; the persistent error message generated whenever swift-build runs:

'The system cannot find the path specified.'

(Coincidentally, it was also creating a file called "null" in my "C:/dev" directory.)


pcbeard:
@swift-ci please smoke test

compnerd (approved):
The new redirection is correct for Windows; `NUL` exists in all directories as a null device.

neonichu (approved):

dmbryson (approved):

PR #752, opened 2021-11-30
Expose TaskInterface's spawn API to clients

BenchR267:
TaskInterface offers an API to spawn sub-processes for ExternalCommands' executeExternalCommand method. This API was not exposed to Swift clients before. With this change, Swift (and C) clients can spawn sub-processes from within the ExternalCommand context enabling custom commands with custom dependency and lifecycle logic while still using llbuild for the execution of processes.

rdar://58130781


BenchR267:
@swift-ci please smoke test

jakepetroules (approved):

PR #754, opened 2021-12-05
Revert "Expose TaskInterface's spawn API to clients"

compnerd:
Reverts apple/swift-llbuild#752

This regressed the Windows build: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/418/consoleText

PR #755, opened 2021-12-05
[Windows Build] Change processStarted(pid:) to use llbuild_pid_t?

pcbeard:
This is needed because Windows models llbuild_pid_t as a HANDLE, which is an implicitly unwrapped pointer, so the value can be nil, INVALID_HANDLE_VALUE, or a valid HANDLE value.

### DIFF
--- a/products/llbuildSwift/BuildSystemBindings.swift
+++ b/products/llbuildSwift/BuildSystemBindings.swift
@@ -104,7 +104,7 @@ public struct BuildSystemCommandInterface {
     /// - commandLine: All command line arguments
     /// - environment: The environment the process will be executed in
     /// - workingDirectory: The path to the directory the process will use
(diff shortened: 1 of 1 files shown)


compnerd (approved):
This should work fine for `HANDLE`.

compnerd:
@swift-ci please smoke test

jakepetroules:
This seems fine. @compnerd If we take this patch, we don't need #754, correct?

jakepetroules (approved):

compnerd:
@jakepetroules yes, if we get this merged, I believe that this should fix the issue and make #754 unnecessary.

compnerd:
Lets merge this then to get the CI back to green.

BenchR267:
Thank you! I'll contact @shahmishal to figure out how to get a Windows build in the Swift-CI to catch this pre-merge next time. Sorry for the trouble and thanks for the fast fix.

PR #757, opened 2022-01-12
Add API to return CommandResult for ExternalCommands

BenchR267:
ExternalCommands mimic a process that runs in-process but could express their result only as a bool before.
With this change, an ExternalCommand can also set itself as cancelled or skipped on their own logic.
That allows clients to cancel commands if their dependencies exit with an unwanted BuildValue.

rdar://87451647


BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

BenchR267:
@swift-ci please smoke test

jakepetroules (approved):

dmbryson (approved):

PR #761, opened 2022-01-27
Implement `withCArrayOfStrings` non-recursive

BenchR267:
For very large inputs the recursive implemenation might hit a stack overflow.

rdar://88132779

### DIFF
--- a/products/llbuildSwift/Internals.swift
+++ b/products/llbuildSwift/Internals.swift
@@ -46,39 +46,46 @@ internal func stringFromData(_ data: llb_data_t) -> String {
 }
 
 extension Array where Element == String {
-    internal func withCArrayOfStrings<T>(_ body: @escaping (UnsafePointer<UnsafePointer<CChar>>) -> T) -> T {
-        func appendPointer(_ index: Array.Index, to target: inout Array<UnsafePointer<CChar>>) -> T {
-            if index == self.endIndex {
-                return body(&target)
-            } else {
-                return self[index].withCString { cStringPtr in
-                    target.append(cStringPtr)
-                    return appendPointer(self.index(after: index), to: &target)
-                }
-            }
+    private func withTemporaryBuffer<T>(_ body: @escaping (UnsafePointer<UnsafePointer<CChar>>) -> T) -> T {
+        // The buffer is in the form "a1\0a2\0a3\0…"
+        let totalLength = reduce(0) { $0 + $1.utf8.count + 1 }
+        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: totalLength)
+        let buffer = UnsafeMutableBufferPointer(start: pointer, count: totalLength)
+        let elementPointers = UnsafeMutablePointer<UnsafePointer<CChar>>.allocate(capacity: count)
+        
+        defer {
+            pointer.deallocate()
+            elementPointers.deallocate()
         }
-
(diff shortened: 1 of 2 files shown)


products/llbuildSwift/Internals.swift
```diff
@@ -46,39 +46,39 @@ internal func stringFromData(_ data: llb_data_t) -> String {
...
-                return body(&target)
-            } else {
-                return self[index].withCString { cStringPtr in
-                    target.append(cStringPtr)
-                    return appendPointer(self.index(after: index), to: &target)
-                }
+    private func withTemporaryBuffer<T>(_ body: @escaping (UnsafePointer<UnsafePointer<CChar>>) -> T) -> T {
+        // The buffer is in the form "a1\0a2\0a3\0…"
+        let totalLength = reduce(0) { $0 + $1.utf8.count + 1 }
+        let buffer = UnsafeMutableRawPointer.allocate(byteCount: totalLength, alignment: 1)
```
jakepetroules, line 52: Shouldn't alignment be something more principled than 1?

products/llbuildSwift/Internals.swift
```diff
@@ -46,39 +46,39 @@ internal func stringFromData(_ data: llb_data_t) -> String {
...
+                bufferCursor.copyMemory(from: ptr, byteCount: element.utf8.count)
+                bufferCursor = bufferCursor.advanced(by: element.utf8.count)
+                bufferCursor.copyMemory(from: "\0", byteCount: 1)
+                bufferCursor = bufferCursor.advanced(by: 1)
+                return result
             }
+            elementPointers[index] = elementPtr
+        }
+        let result = body(UnsafePointer(elementPointers))
+        buffer.deallocate()
```
jakepetroules, line 67: Suggest moving this up to a `defer` right below buffer... less risky we leak if things are restructured later.

Or, can you use one of the `with` methods that takes a block?

products/llbuildSwift/Internals.swift
```diff
@@ -46,39 +46,39 @@ internal func stringFromData(_ data: llb_data_t) -> String {
...
-            } else {
-                return self[index].withCString { cStringPtr in
-                    target.append(cStringPtr)
-                    return appendPointer(self.index(after: index), to: &target)
-                }
+    private func withTemporaryBuffer<T>(_ body: @escaping (UnsafePointer<UnsafePointer<CChar>>) -> T) -> T {
+        // The buffer is in the form "a1\0a2\0a3\0…"
+        let totalLength = reduce(0) { $0 + $1.utf8.count + 1 }
+        let buffer = UnsafeMutableRawPointer.allocate(byteCount: totalLength, alignment: 1)
+        let elementPointers = UnsafeMutablePointer<UnsafePointer<CChar>>.allocate(capacity: count)
```
jakepetroules, line 53: A deallocate() is missing for elementPointers

products/llbuildSwift/Internals.swift
BenchR267: Changing this to be `UnsafeMutablePointer<CChar>` instead, which defines the alignment based on the generic type.

products/llbuildSwift/Internals.swift
BenchR267: I added a `defer` after allocation of the pointers.

BenchR267:
@swift-ci please smoke test

dmbryson (approved):

PR #774, opened 2022-04-08
Add option to mark dependencies as single use

BenchR267:
A dependency marked as singleUse will only be added to the dependencies of the requesting task for the current build iteration.
This allows tasks to request other tasks for running once and not get invalidated for following incremental builds.

rdar://91494744

### DIFF
--- a/include/llbuild/Core/DependencyKeyIDs.h
+++ b/include/llbuild/Core/DependencyKeyIDs.h
@@ -1,4 +1,4 @@
-//===- AttributedKeyIDs.h ---------------------------------------*- C++ -*-===//
+//===- DependencyKeyIDs.h ---------------------------------------*- C++ -*-===//
 //
 // This source file is part of the Swift.org open source project
 //
@@ -26,19 +26,48 @@ namespace core {
 /// compact form.
 // FIXME: At some point, figure out the optimal representation for this
 // data structure, which is likely to be a lot of the resident memory size.
-class AttributedKeyIDs {
+class DependencyKeyIDs {
   // Dependencies to be kept track of.
   std::vector<KeyID> keys;
 
-  /// Flags associated with the keys.
-  std::vector<bool> flags;
-
+  /// Flags about the dependency relation, by bit.
+  ///
+  /// value & 1: Flag indicating if the dependency invalidates the downstream task
+  /// (value >> 1) & 1: Flag indicating if the dependency is valid for only the current build and discarded for incremental builds
+  std::vector<uint8_t> flags;
+  
+private:
+  bool singleUse(int index) const {
+    return (flags[index] >> 1) & 1;
+  }
+  
+  bool orderOnly(int index) const {
+    return flags[index] & 1;
+  }
+  
 public:
 
   /// Clear the contents of the set.
   void clear() {
     keys.clear();
     flags.clear();
+  }
+  
+  /// Removes entries that are flagged as `singleUse`.
+  void cleanSingleUseDependencies() {
+    bool shouldClean = size() > 0;
+    if (!shouldClean) {
+      return;
+    }
+    
+    // Go in reverse order to make
+    for (int i = (int)size(); i > 0; i--) {
+      int index = i - 1;
+      if (singleUse(index) == true) {
+        keys.erase(keys.begin() + index);
+        flags.erase(flags.begin() + index);
+      }
+    }
   }
 
   /// Check whether the set is empty.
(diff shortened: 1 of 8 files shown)


BenchR267:
@swift-ci please smoke test

include/llbuild/Core/DependencyKeyIDs.h
```diff
@@ -0,0 +1,154 @@
...
+/// A data structure representing a set of tuples (KeyID, flag) in a somewhat
+/// compact form.
+// FIXME: At some point, figure out the optimal representation for this
+// data structure, which is likely to be a lot of the resident memory size.
+class DependencyKeyIDs {
+  // Dependencies to be kept track of.
+  std::vector<KeyID> keys;
+
+  /// Flag indicating if the dependency invalidates the downstream task
+  std::vector<bool> orderOnly;
```
owenv, line 34: Based on the FIXME above, I wonder if this should be a vector of flag bitfields instead of two vectors of bools as an optimization

include/llbuild/Core/DependencyKeyIDs.h
```diff
@@ -0,0 +1,154 @@
...
+  
+  /// Removes entries that are flagged as `singleUse`.
+  void cleanSingleUseDependencies() {
+    bool shouldClean = size() > 0;
+    if (!shouldClean) {
+      return;
+    }
+    
+    std::vector<size_t> indicesToRemove;
+    for (size_t i = 0; i < size(); i++) {
```
owenv, line 57: If you iterate over the vector in reverse here, you can remove the entries as you go without a second loop below

owenv (approved):
LGTM! I left a couple notes in places I think could be optimized a little

include/llbuild/Core/DependencyKeyIDs.h
BenchR267: Ah right, even better

include/llbuild/Core/DependencyKeyIDs.h
BenchR267: Yeah, that sounds reasonable. Let me change that.

BenchR267: ✅

BenchR267:
@swift-ci please smoke test

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -614,17 +616,17 @@ class SQLiteBuildDB : public BuildDB {
...
       // FIXME: This is naively mapping all keys with no caching at this point,
       // thus likely to perform poorly.  Should refactor this into a bulk
       // query or a DB layer cache.
-      auto dbKeyID = getKeyID(keyIDAndFlag.keyID, error_out);
+      auto dbKeyID = getKeyID(dependency.keyID, error_out);
       if (!error_out->empty()) {
         return false;
       }
-      encoder.write((dbKeyID.value << 1) + keyIDAndFlag.flag);
+      encoder.write((dbKeyID.value << 2) + (dependency.singleUse << 1) + dependency.orderOnly);
```
dmbryson, line 629: Should this really be writing single use dependencies into the DB?   Wouldn't it be more appropriate to return an error here?   Single use entries by definition shouldn't ever be written to the db.

include/llbuild/Core/DependencyKeyIDs.h
```diff
@@ -26,20 +26,49 @@ namespace core {
...
-  std::vector<bool> flags;
-
+  /// Flags about the dependency relation, by bit.
+  ///
+  /// value & 1: Flag indicating if the dependency invalidates the downstream task
+  /// (value >> 1) & 1: Flag indicating if the dependency is valid for only the current build and discarded for incremental builds
+  std::vector<uint8_t> flags;
+  
+private:
+  bool singleUse(int index) const {
```
dmbryson, line 40: These accessors should really be `size_type` (which I think is `size_t` here) not `int`.

include/llbuild/Core/DependencyKeyIDs.h
```diff
@@ -26,20 +26,49 @@ namespace core {
...
+  
+  /// Removes entries that are flagged as `singleUse`.
+  void cleanSingleUseDependencies() {
+    bool shouldClean = size() > 0;
+    if (!shouldClean) {
+      return;
+    }
+    
+    // Go in reverse order to make
+    for (int i = (int)size(); i > 0; i--) {
```
dmbryson, line 64: Casting to `(int)` here being the red flag about sign conversion.

dmbryson (commented):
A couple bits of cleanup that might be worthwhile here.

PR #824, opened 2022-06-02
Delete forwarding delegate after use

BenchR267:
rdar://93555189

### DIFF
--- a/products/libllbuild/BuildSystem-C-API.cpp
+++ b/products/libllbuild/BuildSystem-C-API.cpp
@@ -1235,9 +1235,12 @@ bool llb_buildsystem_command_interface_spawn(llb_task_interface_t ti, llb_builds
     }
 
   };
-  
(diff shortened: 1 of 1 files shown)


products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -1235,9 +1235,12 @@ bool llb_buildsystem_command_interface_spawn(llb_task_interface_t ti, llb_builds
     }
 
   };
-  
-  coreti->spawn((QueueJobContext*)job_context, ArrayRef<StringRef>(arguments), ArrayRef<std::pair<StringRef, StringRef>>(environment), {true, false, StringRef((const char*)working_dir->data, working_dir->length)}, {commandCompletionFn}, new ForwardingProcessDelegate(delegate));
-  
+
+  auto forwardingDelegate = new ForwardingProcessDelegate(delegate);
```
jakepetroules, line 1239: nit: you could also use this to manage lifetime automatically:

```c++
auto forwardingDelegate = std::make_shared<ForwardingProcessDelegate>(delegate);
```

(but you'd have to pass `forwardingDelegate.data()` to the function unless you want to refactor it to take a shared ptr)

jakepetroules (approved):

products/libllbuild/BuildSystem-C-API.cpp
BenchR267: Good to know, thanks! I'll leave it as is for now.

PR #826, opened 2022-06-08
Guard against deletion of an in-use build engine

dmbryson:
rdar://85827825


dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please test

PR #827, opened 2022-06-09
[next] Add missing include for ratio

bnbarham:
`Chrono.h` uses `std::ratio`, but was relying on a transitive include
from `<chrono>`. Include `<ratio>` directly.


bnbarham:
@swift-ci please test

dmbryson (approved):

dmbryson:
@swift-ci please smoke test

PR #831, opened 2022-06-15
Add an API to notify the BuildEngine/BuildSystem delegate when it is determined that a rule needs to run

owenv:
…, and why

For example, because it has never been built, or because one of its inputs was rebuilt

rdar://95208406

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -47,6 +47,8 @@ void Rule::updateStatus(BuildEngine&, StatusKind) {}
 
 BuildEngineDelegate::~BuildEngineDelegate() {}
 
+void BuildEngineDelegate::determinedRuleNeedsToRun(Rule* ruleNeedingToRun, Rule::RunReason, Rule* inputRule) {}
+
 bool BuildEngineDelegate::shouldResolveCycle(const std::vector<Rule*>& items,
                                              Rule* candidateRule,
                                              Rule::CycleAction action) {
@@ -460,6 +462,7 @@ private:
       if (trace)
         trace->ruleNeedsToRunBecauseNeverBuilt(ruleInfo.rule.get());
       ruleInfo.state = RuleInfo::StateKind::NeedsToRun;
+      delegate.determinedRuleNeedsToRun(ruleInfo.rule.get(), Rule::RunReason::NeverBuilt, nullptr);
       return true;
     }
 
@@ -467,6 +470,7 @@ private:
       if (trace)
         trace->ruleNeedsToRunBecauseSignatureChanged(ruleInfo.rule.get());
       ruleInfo.state = RuleInfo::StateKind::NeedsToRun;
+      delegate.determinedRuleNeedsToRun(ruleInfo.rule.get(), Rule::RunReason::SignatureChanged, nullptr);
       return true;
     }
 
@@ -479,6 +483,7 @@ private:
       if (trace)
         trace->ruleNeedsToRunBecauseInvalidValue(ruleInfo.rule.get());
       ruleInfo.state = RuleInfo::StateKind::NeedsToRun;
+      delegate.determinedRuleNeedsToRun(ruleInfo.rule.get(), Rule::RunReason::InvalidValue, nullptr);
       return true;
     }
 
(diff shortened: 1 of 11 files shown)


owenv:
@swift-ci please test

products/libllbuild/include/llbuild/buildsystem.h
```diff
@@ -203,6 +203,24 @@ typedef enum LLBUILD_ENUM_ATTRIBUTES {
...
+    llb_rule_run_reason_signature_changed LLBUILD_SWIFT_NAME(signatureChanged) = 1,
+
+    /// The rule has an invalid value.
+    llb_rule_run_reason_invalid_value LLBUILD_SWIFT_NAME(invalidValue) = 2,
+
+    /// An input of the rule has been rebuilt.
+    llb_rule_run_reason_input_rebuilt LLBUILD_SWIFT_NAME(inputRebuilt) = 3,
+
+    /// The rule has never been built.
+    llb_rule_run_reason_forced = 4,
```
dmbryson, line 221: LLBUILD_SWIFT_NAME for consistency?

dmbryson (approved):

products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -471,6 +471,39 @@ class CAPIBuildSystemFrontendDelegate : public BuildSystemFrontendDelegate {
     uint64_t count() { return rules.size(); }
   };
 
+  static llb_rule_run_reason_t
+  convertRunReason(core::Rule::RunReason reason) {
```
BenchR267, line 475: Maybe ad a commend to both types and do a static cast? They're based on the same raw value.

lib/Core/BuildEngine.cpp
```diff
@@ -479,6 +483,7 @@ class BuildEngineImpl : public BuildDBDelegate {
       if (trace)
         trace->ruleNeedsToRunBecauseInvalidValue(ruleInfo.rule.get());
```
BenchR267, line 484: Would be cool if the trace would also use the new enum

BenchR267 (approved):

products/libllbuild/BuildSystem-C-API.cpp
owenv: I think I prefer the current approach to maintain consistency with the other enums, so we have a clearer boundary between the C & C++ APIs

lib/Core/BuildEngine.cpp
owenv: That's a good idea. I'll file a follow-up bug for the refactor

owenv:
@swift-ci please smoke test

products/libllbuild/BuildSystem-C-API.cpp
jakepetroules: Agree with @owenv, despite the verbosity, forced casts based on assumptions about the underlying integer value can be dangerous.

products/libllbuild/include/llbuild/buildsystem.h
```diff
@@ -203,6 +203,24 @@ typedef enum LLBUILD_ENUM_ATTRIBUTES {
     llb_quality_of_service_unspecified = 4
 } llb_quality_of_service_t LLBUILD_SWIFT_NAME(QualityOfService);
 
+/// Rule run reasons.
+typedef enum LLBUILD_ENUM_ATTRIBUTES {
+    /// The rule has never been built.
+    llb_rule_run_reason_never_built LLBUILD_SWIFT_NAME(neverBuilt) = 0,
```
jakepetroules, line 209: LLBUILD_SWIFT_NAME 👍

jakepetroules (approved):

jakepetroules:
@swift-ci please smoke test

owenv:
It turns out tests sometimes use a BuildSystemDelegate which is not a BuildSystemFrontendDelegate, so the `static_cast<BuildSystemFrontendDelegate*>(&getBuildSystem().getDelegate())` was unsafe. I've added `determinedRuleNeedsToRun` as an abstract method on BuildSystemDelegate, which lets me drop the unsound cast and generally seems reasonable.

owenv:
@swift-ci please smoke test and merge

owenv:
@swift-ci please smoke test

PR #832, opened 2022-06-19
Parallel  Incremental Builds

erikolofsson:
This is my try at solving FB8394316 that I reported. I found no good way to know when it's safe to scan dependencies in parallel. If you scan all of them at once you end up with incorrect scan results. I ended up solving this with barriers between batches of dependencies. This passes all the tests and seems to work ok in Xcode. The barriers are created when a result is reported to a task as that should prevent calculations on stale values.

rdar://20248283

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -111,6 +111,8 @@ class BuildEngineImpl : public BuildDBDelegate {
     uintptr_t inputID;
     /// The rule for the input which was requested.
     RuleInfo* inputRuleInfo;
+    /// The dependency batch sequence when the input was requested
+    unsigned dependencyBatchSequence = 0;
     /// Whether this rule is to be executed as order-only.
     bool orderOnly = false;
     /// Force the use of a prior value
@@ -130,18 +132,26 @@ class BuildEngineImpl : public BuildDBDelegate {
     /// The rule making the request.
     RuleInfo* ruleInfo;
     /// The input index being considered.
-    unsigned inputIndex;
+    unsigned inputIndex = 0;
+    /// The input index that the below cache fields are valid for
+    unsigned cachedInputIndex = -1;
     /// The input being considered, if already looked up.
     ///
     /// This is used when a scan request is deferred waiting on its input to be
     /// scanned, to avoid a redundant hash lookup.
-    RuleInfo* inputRuleInfo;
-    /// Whether the rule is executed in order-only way.
-    bool orderOnly;
-    /// Whether the rule is cleaned from dependencies after execution.
+    RuleInfo* inputRuleInfo = nullptr;
+
+    /// The inputs that we are waiting for to finish for the current barrier
+    std::vector<unsigned> pendingInputs;
+
+    /// Whether the cached rule is executed in order-only way.
+    bool orderOnly = false;
+    /// Whether the cached rule is cleaned from dependencies after execution.
     bool singleUse = false;
+    /// Whether the rule scan request is already in the scan queue
+    bool isInScanQueue = true;
   };
-  std::vector<RuleScanRequest> ruleInfosToScan;
+  std::vector<std::shared_ptr<RuleScanRequest>> ruleInfosToScan;
 
   struct RuleScanRecord {
     /// The vector of paused input requests, waiting for the dependency scan on
@@ -149,7 +159,7 @@ class BuildEngineImpl : public BuildDBDelegate {
     std::vector<TaskInputRequest> pausedInputRequests;
     /// The vector of deferred scan requests, for rules which are waiting on
     /// this one to be scanned.
-    std::vector<RuleScanRequest> deferredScanRequests;
+    std::vector<std::shared_ptr<RuleScanRequest>> deferredScanRequests;
   };
 
   /// Wrapper for information specific to a single rule.
@@ -315,12 +325,17 @@ class BuildEngineImpl : public BuildDBDelegate {
     /// this one to be scanned.
     //
     // FIXME: As above, this structure has redundancy in it.
-    std::vector<RuleScanRequest> deferredScanRequests;
+    std::vector<std::shared_ptr<RuleScanRequest>> deferredScanRequests;
     /// The rule that this task is computing.
     RuleInfo* forRuleInfo = nullptr;
+    /// Used to sequence batches of inputs that can be scanned in parallel
+    std::atomic<unsigned> dependencyBatchSequence{0};
     /// The number of outstanding inputs that this task is waiting on to be
     /// provided.
     unsigned waitCount = 0;
+    /// Keep track of last dependency batch sequence to know when to insert
+    /// barriers in the rule info result
+    unsigned lastDependencyBatchSequence = 0;
     /// The list of discovered dependencies found during execution of the task.
     DependencyKeyIDs discoveredDependencies;
 
@@ -505,7 +520,7 @@ private:
       trace->ruleScheduledForScanning(ruleInfo.rule.get());
     ruleInfo.state = RuleInfo::StateKind::IsScanning;
     ruleInfo.setPendingScanRecord(newRuleScanRecord());
-    ruleInfosToScan.push_back({ &ruleInfo, /*InputIndex=*/0, nullptr, false });
+    ruleInfosToScan.push_back(std::make_shared<RuleScanRequest>(RuleScanRequest{ &ruleInfo }));
 
     return false;
   }
@@ -546,7 +561,7 @@ private:
 
     // register the task
     taskInfosMutex.lock();
-    auto result = taskInfos.emplace(task, TaskInfo(task));
+    auto result = taskInfos.emplace(std::piecewise_construct, std::forward_as_tuple(task), std::forward_as_tuple(task));
     assert(result.second && "task already registered");
     auto taskInfo = &(result.first)->second;
     taskInfosMutex.unlock();
@@ -596,7 +611,9 @@ private:
   ///
   /// This will process all of the inputs required by the requesting rule, in
   /// order, unless the scan needs to be deferred waiting for an input.
-  void processRuleScanRequest(RuleScanRequest request) {
+  void processRuleScanRequest(std::shared_ptr<RuleScanRequest> const &requestPtr) {
+    auto& request = *requestPtr;
+
     auto& ruleInfo = *request.ruleInfo;
 
     // With forced builds in cycle breaking, we may end up being asked to scan
@@ -605,14 +622,16 @@ private:
     if (!ruleInfo.isScanning())
       return;
 
-    // Process each of the remaining inputs.
-    do {
+    std::vector<unsigned> newPendingInputs;
+
+    auto processInput = [&](unsigned inputIndex) -> bool {
       // Look up the input rule info, if not yet cached.
-      if (!request.inputRuleInfo) {
-        const auto& keyAndFlag = ruleInfo.result.dependencies[request.inputIndex];
+      if (!request.inputRuleInfo || request.cachedInputIndex != inputIndex) {
+        const auto& keyAndFlag = ruleInfo.result.dependencies[inputIndex];
         request.inputRuleInfo = &getRuleInfoForKey(keyAndFlag.keyID);
         request.orderOnly = keyAndFlag.orderOnly;
         request.singleUse = keyAndFlag.singleUse;
+        request.cachedInputIndex = inputIndex;
       }
 
       auto& inputRuleInfo = *request.inputRuleInfo;
@@ -627,8 +646,10 @@ private:
           trace->ruleScanningDeferredOnInput(ruleInfo.rule.get(),
                                              inputRuleInfo.rule.get());
         inputRuleInfo.getPendingScanRecord()
-          ->deferredScanRequests.push_back(request);
-        return;
+          ->deferredScanRequests.push_back(requestPtr);
+
+        newPendingInputs.push_back(inputIndex);
+        return false;
       }
 
       if (trace)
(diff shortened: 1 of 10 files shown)


dmbryson:
This may take me a bit to review, but I am looking and will respond.

dmbryson:
FYI, I haven't forgotten about this.   Overall I really like the approach so far.  Just digging through the build engine logic itself, amidst other distractions.

lib/Core/BuildEngine.cpp
```diff
@@ -687,8 +740,13 @@ class BuildEngineImpl : public BuildDBDelegate {
     auto scanRecord = inputRuleInfo.getPendingScanRecord();
 
     // Wake up all of the pending scan requests.
-    for (const auto& request: scanRecord->deferredScanRequests) {
-      ruleInfosToScan.push_back(request);
+    for (auto& request: scanRecord->deferredScanRequests) {
+      if ((--request->numPending) == 0 && !request->isInScanQueue) {
```
dmbryson, line 744: 🤔 This means that we are going to wait for _all_ pending scan requests in the current batch before we process the deferred scan again.   Can we short circuit that for cases where the current scan has been determined as needs to run?

dmbryson (commented):
Overall I really like the approach and I think this looks solid.  My one concern, noted above, is that we may be holding tasks waiting for more scan results than we actually need, thus delaying the start of tasks.

lib/Core/BuildEngine.cpp
erikolofsson: If we short circuit we cannot use a simple numPending to keep track of when all are finished as we will be reprocessing all pending requests and some old might still be deferred. 

I committed a simpler version that just re-processes after each result is available. This has the drawback that we will end up with the same request in the deferredScanRequests several times as well as reprocessing unnecessarily.

To fix this I added another commit that uses an unordered_set instead. This has the drawback that it will make the order of deferredScanRequests random. Is this ok, or should I use both a vector and unordered_set?

Finally I added back so we have the same logic but with the short circuit. This with an unordered_set instead of a counter to keep track of when to requeue.

I don't have access to a benchmark so I don't know which of these is the best performing.

dmbryson: I agree that always reprocessing seems suboptimal.  Indeed, it is not clear if the random ordering of deferredScanRequests would be problematic.  I think where you have landed with the combination of unordered set  and the short circuit seems like a good balance.

dmbryson:
@swift-ci please smoke test

dmbryson:
Apologies that this has sat in the queue for so long. It is not forgotten. I think it is ready to go, but would like to run some extra qualification runs before integrating.

erikolofsson:
> Apologies that this has sat in the queue for so long. It is not forgotten. I think it is ready to go, but would like to run some extra qualification runs before integrating.

No problem. I think it needs more testing as I have been running this in Xcode for over a month now and I have noticed an issue that it can get into a state where it's not rebuilding things that have changed. I have observed the following:
* I think that it can only only happen after a build error (I have "Continue building after errors" unchecked)
* Restarting Xcode will let it discover and build the change files correctly only once.

I have not had time to debug it, but maybe the symtoms give you a clue for what could be causing this?

chiragramani:
Hey @erikolofsson, can you share the setup that you are using to run this in Xcode? I will be interested in debugging the issue that you described. cc: @dmbryson if you have any leads that can help with the above.

brentleyjones:
@erikolofsson @dmbryson @chiragramani It seems this has stalled out. Is there anything we can do to make progress on it?

lib/Core/BuildEngine.cpp
```diff
@@ -959,12 +1035,22 @@ class BuildEngineImpl : public BuildDBDelegate {
...
         // Add all of the task's discovered dependencies.
         //
         // FIXME: We could audit these dependencies at this point to verify that
         // they are not keys for rules which have not been run, which would
         // indicate an underspecified build (e.g., a generated header).
-        ruleInfo->result.dependencies.append(taskInfo->discoveredDependencies);
+        if (!taskInfo->discoveredDependencies.empty()) {
+          ruleInfo->result.dependencies.addBarrier();
+          ruleInfo->result.dependencies.append(taskInfo->discoveredDependencies);
+        }
```
tapthaker, line 1054: @erikolofsson: The builds are not triggering correctly because this piece of code doesn't seem to be executed in case of build failures. In the subsequent builds, the `result.dependencies` is empty, resulting in `ruleDoesNotNeedToRun`. A not-so-nice solution would be to re-add the following near L904
```
request.taskInfo->forRuleInfo->result.dependencies.push_back(
            request.inputRuleInfo->keyID, request.orderOnly, request.singleUse);
```

Ideally, we should move the `sequencedDependencyBatches` inside `Result` but I find it risky since `Result` is persisted inside the Sqlite DB.

tapthaker:
Here's the patch on top of this PR if someone wants to try my not-so-nice solution 

[llbuild-fix-incremental-after-build-failure-patch.txt](https://github.com/apple/swift-llbuild/files/10079546/llbuild-fix-incremental-after-build-failure-patch.txt)


In order to make it work inside Xcode 14.1 you need to replace `llbuild.framework` inside and then re-sign:
- `llbuild.framework`
- `XCBBuildService.framework`
- `XCBBuildService`

lib/Core/BuildEngine.cpp
erikolofsson: Thanks. I refactored so `sequencedDependencyBatches` isn't needed and it adds to the result directly (See [branch](https://github.com/erikolofsson/swift-llbuild/tree/incremental-parallel-20248283-2)). I will be doing some longer term testing and if it holds up I will push the fix here.

tapthaker: Cool. We'll test this out here at Uber as well 😃

erikolofsson: I noticed that the incremental build wasn't parallell when it should have been so I fixed this by recording the batch sequence at the time the input was requested instead of when it was popped off the queue and pushed the fix to the [branch](https://github.com/erikolofsson/swift-llbuild/tree/incremental-parallel-20248283-2).

tapthaker:
@erikolofsson : Can you update this PR with the changes in your branch https://github.com/erikolofsson/swift-llbuild/tree/incremental-parallel-20248283-2

tapthaker:
@dmbryson: Please re-review this diff. We have been testing this feature here at Uber and it is giving us an improved performance due to parallel module builds. 

We did some local analysis to see the difference. The bug was that llbuild doesn’t scan rules until all the lanes are free; so it would not queue in more jobs even if all the inputs for a rule are available

Before: rules are not scanned until all lanes are free

After: Rulesmaite now scanned after every task finisha.
Resulting.sonminerallel builds

erikolofsson:
Since the code was reviewed the last time three bugs were fixed:
* Cancelling builds didn't record the dependencies
* Dependency barriers were applied at the time of processing instead of at request, this caused lost build parallelism depending on timing
* When dependencies were checked in parallel outputs that were deleted in a dependency weren't detected

You can see differences [here](https://github.com/apple/swift-llbuild/compare/ae8c36e0ef0d659c4b4e09eaab605879a1e009b1..618ddfb841bda02b8a5a8bb4bb8cb496ed68d3f6).

I have been running the latest code for two weeks without finding any new issues so I believe this is ready for review again.

dmbryson:
@swift-ci please smoke test

tapthaker:
The latest changes in the PR crash for us. https://github.com/erikolofsson/swift-llbuild/tree/incremental-parallel-20248283-2 works fine. 
```
Thread 8 Crashed::  Dispatch queue: XCBBuildSystem.BuildManager.buildQueue
0   libsystem_pthread.dylib       	       0x1a8d43570 pthread_mutex_lock + 12
1   libc++.1.dylib                	       0x1a8c989e8 std::__1::mutex::lock() + 16
2   llbuild                       	       0x102af9970 std::__1::lock_guard<std::__1::mutex>::lock_guard(std::__1::mutex&) + 8 (__mutex_base:90) [inlined]
3   llbuild                       	       0x102af9970 std::__1::lock_guard<std::__1::mutex>::lock_guard(std::__1::mutex&) + 8 (__mutex_base:90) [inlined]
4   llbuild                       	       0x102af9970 (anonymous namespace)::BuildEngineImpl::getTaskInfo(llbuild::core::Task*) + 36 (BuildEngine.cpp:1551)
5   llbuild                       	       0x102af98c0 (anonymous namespace)::BuildEngineImpl::addTaskInputRequest(llbuild::core::Task*, llbuild::core::KeyType const&, unsigned long, bool, bool) + 52 (BuildEngine.cpp:1815)
6   llbuild                       	       0x102a76874 llb_buildsystem_command_interface_task_needs_input + 52 (BuildSystem-C-API.cpp:1215)
7   XCBBuildSystem                	       0x10234d5c8 protocol witness for DynamicTaskExecutionDelegate.requestInputNode(node:nodeID:) in conformance OperatorSystemAdaptorDynamicContext + 100
8   XCBTaskExecution              	       0x102ddde44 SwiftDriverJobSchedulingTaskAction.taskSetup(_:dynamicExecutionDelegate:) + 528
9   XCBBuildSystem                	       0x10235e0f0 specialized InProcessCommand.start(_:_:) + 152
10  llbuild                       	       0x102a89b68 specialized CommandWrapper.configure(_:_:_:_:) + 1296 (BuildSystemBindings.swift:409)
11  llbuild                       	       0x102a84c50 CommandWrapper.configure(_:_:_:_:) + 16 [inlined]
12  llbuild                       	       0x102a84c50 closure #1 in ToolWrapper.buildCommand(_:_:) + 92 (BuildSystemBindings.swift:202)
13  llbuild                       	       0x102a76694 (anonymous namespace)::CAPIExternalCommand::CAPIExternalCommand(llvm::StringRef, llb_buildsystem_external_command_delegate_t_) + 200 (BuildSystem-C-API.cpp:1059) [inlined]
14  llbuild                       	       0x102a76694 (anonymous namespace)::CAPIExternalCommand::CAPIExternalCommand(llvm::StringRef, llb_buildsystem_external_command_delegate_t_) + 200 (BuildSystem-C-API.cpp:1039) [inlined]
15  llbuild                       	       0x102a76694 llb_buildsystem_external_command_create + 264 (BuildSystem-C-API.cpp:1171)
16  llbuild                       	       0x102a84a54 ToolWrapper.buildCommand(_:_:) + 320 (BuildSystemBindings.swift:221)
17  llbuild                       	       0x102a88c48 closure #1 in ToolWrapper.createCommand(_:) + 20 (BuildSystemBindings.swift:176) [inlined]
18  llbuild                       	       0x102a88c48 specialized Optional.map<A>(_:) + 20 [inlined]
19  llbuild                       	       0x102a88c48 ToolWrapper.createCommand(_:) + 32 (BuildSystemBindings.swift:176) [inlined]
20  llbuild                       	       0x102a88c48 closure #1 in BuildSystem.lookupTool(_:) + 172 (BuildSystemBindings.swift:1192)
21  llbuild                       	       0x102a77ee4 (anonymous namespace)::CAPITool::createCommand(llvm::StringRef) + 36 (BuildSystem-C-API.cpp:685)
22  llbuild                       	       0x102b0f260 (anonymous namespace)::BuildFileImpl::parseCommandsMapping(llvm::yaml::MappingNode*) + 528 (BuildFile.cpp:815) [inlined]
23  llbuild                       	       0x102b0f260 (anonymous namespace)::BuildFileImpl::parseRootNode(llvm::yaml::Node*) + 6580 (BuildFile.cpp:434)
24  llbuild                       	       0x102b0cd7c (anonymous namespace)::BuildFileImpl::load() + 228 (BuildFile.cpp:1015) [inlined]
25  llbuild                       	       0x102b0cd7c llbuild::buildsystem::BuildFile::load() + 272 (BuildFile.cpp:1062)
26  llbuild                       	       0x102b1288c (anonymous namespace)::BuildSystemImpl::loadDescription(llvm::StringRef) + 76 (BuildSystem.cpp:303) [inlined]
27  llbuild                       	       0x102b1288c llbuild::buildsystem::BuildSystem::loadDescription(llvm::StringRef) + 104 (BuildSystem.cpp:4053)
28  llbuild                       	       0x102b06978 (anonymous namespace)::BuildSystemFrontendImpl::initialize() + 268 (BuildSystemFrontend.cpp:390)
29  llbuild                       	       0x102b06dac (anonymous namespace)::BuildSystemFrontendImpl::build(llvm::StringRef) + 8 (BuildSystemFrontend.cpp:468) [inlined]
30  llbuild                       	       0x102b06dac llbuild::buildsystem::BuildSystemFrontend::build(llvm::StringRef) + 36 (BuildSystemFrontend.cpp:819)
31  llbuild                       	       0x102a88b6c BuildSystem.build(target:) + 152 (BuildSystemBindings.swift:1096)
32  XCBBuildSystem                	       0x102349afc BuildOperation.build() + 9180
33  XCBBuildSystem                	       0x10236215c closure #2 in BuildManager.startBuild(_:) + 60
34  XCBBuildSystem                	       0x1023616d0 thunk for @escaping @callee_guaranteed () -> () + 28
35  libdispatch.dylib             	       0x1a8bd163c _dispatch_block_async_invoke2 + 148
36  libdispatch.dylib             	       0x1a8bc2504 _dispatch_client_callout + 20
37  libdispatch.dylib             	       0x1a8bc5994 _dispatch_continuation_pop + 504
38  libdispatch.dylib             	       0x1a8bc4ffc _dispatch_async_redirect_invoke + 584
39  libdispatch.dylib             	       0x1a8bd3f94 _dispatch_root_queue_drain + 396
40  libdispatch.dylib             	       0x1a8bd47c0 _dispatch_worker_thread2 + 164
41  libsystem_pthread.dylib       	       0x1a8d450c4 _pthread_wqthread + 228
42  libsystem_pthread.dylib       	       0x1a8d43e20 start_wqthread + 8
```

erikolofsson:
> 

I don't think the other commits on main branch are compatible with released Xcode. There shouldn't be any differences in my code between those branches. If you delete the commits after efb176d866e3e99855a5b799da224a9f95058f2e and only include the pull request commits it should work.

dmbryson (approved):

PR #834, opened 2022-06-30
Run ShellCommand dependency parsing at high priority

owenv:
This ensures dependency file parsing runs as soon as possible after a task finishes, in FIFO order when using the LaneBasedExecutionQueue. Previously, it could run significantly later when many tasks were enqueued, delaying "task finished" callbacks despite the fact that very little work remained to be done.

rdar://96151636

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -2169,8 +2169,6 @@ public:
 
       // Otherwise, collect the discovered dependencies, if used.
       if (!depsPath.empty()) {
-        // FIXME: Really want this job to go into a high priority fifo queue
-        // so as to not hold up downstream tasks.
         ti.spawn({ this, [this, ti, completionFn, result](QueueJobContext* context) mutable {
           if (!processDiscoveredDependencies(ti, context)) {
             // If we were unable to process the dependencies output, report a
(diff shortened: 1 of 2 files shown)


dmbryson:
This could have unintended consequences on build performance.  Pushing it to high priority means we'll inject discovered dependency scan tasks potentially ahead of other build work.   It probably isn't a big deal, but also is this really necessary?  The client should already be getting a process finished callback.

owenv:
> This could have unintended consequences on build performance. Pushing it to high priority means we'll inject discovered dependency scan tasks potentially ahead of other build work.

I understand the concern, but I do think it's unlikely to have a meaningful impact, since parsing a dependency file is cheap enough to not significantly delay any enqueued tasks that might unblock additional downstream work, and the current prioritization of tasks by build key is often arbitrary.

> The client should already be getting a process finished callback.

It will receive one for the process execution, but there aren't any callbacks for the start of dependency parsing, so there's no way for the client to know if/when a command has 'yielded' its execution lane while waiting for the second job to be scheduled. We could add new API for this, but I'm not sure it's worth it for this one specific situation.

dmbryson:
Ah, yes I see it.  These do block the result because dep scanning can fail.   I retract my comment and instead suggest we also fix the ClangShellCommand's dep processing too, for consistency.    That way all three ShellCommand are doing the same thing.

dmbryson (approved):

owenv:
@dmbryson sounds good to me! I'd forgotten to check ClangShellCommand, thanks for pointing that out

owenv:
@swift-ci please smoke test

PR #836, opened 2022-07-18
Bump db schema version for filtered directory structure nodes

dmbryson:
rdar://96873217


dmbryson:
@swift-ci please smoke test

PR #843, opened 2022-09-07
Allow configuring the working directory of C-API external commands

owenv:
Some external commands shell out to tools which may produce dependency files. Configuring them with a working directory allows relative paths which may appear in these files to be resolved correctly. The resolution logic added here is identical to ShellCommands'

rdar://99284779


owenv:
@swift-ci please smoke test

jakepetroules (approved):

owenv:
@swift-ci please smoke test

PR #845, opened 2022-10-29
Don't strip symbols in llbuild's static libraries

jakepetroules:
rdar://101713756


jakepetroules:
@swift-ci please smoke test

owenv (approved):

PR #846, opened 2022-10-29
Set the file type of .inc files to "C++ Header"

jakepetroules:
This is necessary for syntax highlighting and breakpoints to work correctly in Xcode.


jakepetroules:
@swift-ci please smoke test

dmbryson (approved):

PR #847, opened 2022-10-30
Add error handling for DirectoryContentsTask::getContents and FilteredDirectoryContentsTask::getFilteredContents

jakepetroules:
rdar://101717159

### DIFF
--- a/lib/BuildSystem/BuildSystem.cpp
+++ b/lib/BuildSystem/BuildSystem.cpp
@@ -846,7 +846,10 @@ class DirectoryContentsTask : public Task {
     }
 
     std::vector<std::string> filenames;
-    getContents(path, filenames);
+    if (getContents(path, filenames)) {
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci please smoke test

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -846,15 +846,18 @@ class DirectoryContentsTask : public Task {
     }
 
     std::vector<std::string> filenames;
-    getContents(path, filenames);
+    if (getContents(path, filenames)) {
+      ti.complete(BuildValue::makeFailedInput().toData());
```
jakepetroules, line 850: @dmbryson Is `makeFailedInput` a reasonable `BuildValue` to use for the failure case here?

dmbryson: Seems reasonable to me.

dmbryson (approved):

PR #848, opened 2022-10-31
Vendor new llvmSupport sources from swift/release/5.7

owenv:
Based on apple/llvm-project at 65411f5b90f6f72550341b4856e9b678f4162d83

The import script is also updated, so if it's rerun there will be no diffs on top of this commit. It does apply two small patches to remove debug options and mixup one import. Most of the changes to llbuild's sources are mechanical, replacing llvm:: with std:: or adding .str() now that the implicit StringRef -> std::string conversion has been removed.

I've tested the xcodeproj and SwiftPM builds with this change, and speculatively updated CMake. It looks like I may need to reapply some BSD/Android patches to the config.h header as well


owenv:
@swift-ci please smoke test

owenv:
TODO: Update CMake sources lists

include/llvm-c/BLAKE3.h
```diff
@@ -0,0 +1,79 @@
+/*===-- llvm-c/blake3.h - BLAKE3 C Interface ----------------------*- C -*-===*\
```
johari, line 1: Seems like this file should be renamed to `llvm-c/blake3.h`.. Could you double check why the name is capitalized in the commit?

It seems like in `lib/llvm/Support/BLAKE3/blake3_impl.h` and `include/llvm/Support/BLAKE3.h` we have the lowercased version `#include "llvm-c/blake3.h"`

PR #849, opened 2022-11-29
Add ProducedDirectoryNodeTask

johari:
rdar://102208116


lib/BuildSystem/BuildSystem.cpp
```diff
@@ -1753,6 +1857,21 @@ std::unique_ptr<Rule> BuildSystemEngineDelegate::lookupRule(const KeyType& keyDa
...
+        },
+        /*IsValid=*/ [node](BuildEngine& engine, const Rule& rule,
+                            const ValueType& value) -> bool {
+          return ProducedDirectoryNodeTask::isResultValid(
+              engine, *node, BuildValue::fromData(value));
+        }
+      ));
+    }
+
     // Otherwise, create a task for a produced node.
```
owenv, line 1875: Can you move this comment above the additions so we differentiate between produced and non-produced nodes?

lib/BuildSystem/ExternalCommand.cpp
```diff
@@ -398,6 +398,9 @@ void ExternalCommand::execute(BuildSystem& system,
       //
       // FIXME: Need to use the filesystem interfaces.
       auto parent = llvm::sys::path::parent_path(node->getName());
+      if (node->getName().endswith("/")) {
```
owenv, line 401: Does this need a special case for the root path?

lib/BuildSystem/BuildNode.cpp
```diff
@@ -133,7 +133,15 @@ bool BuildNode::configureAttribute(
 
 FileInfo BuildNode::getFileInfo(basic::FileSystem& fileSystem) const {
   assert(!isVirtual());
-  return fileSystem.getFileInfo(getName());
+
+  // Drop the trailing slash
+  // otherwise non-directory paths that end with "/" will be reported as missing
+  StringRef path = getName();
+  if (path.endswith("/") && path != "/") {
```
owenv, line 140: Even with the explanatory comments, these checks are very subtle. I wonder if there's a way we could model this better (maybe there isn't)

owenv:
@swift-ci please smoke test

lib/BuildSystem/BuildNode.cpp
johari: I'm not certain, but this might be unavoidable if the high-level build system is encoding all input/output nodes as directory in the manifest.

When high-level build system treats all nodes as directories (regardless of whether user typed in a trailing slash or not), it introduces ambiguity in user's intent, without providing any opportunity for disambiguation.

At the time `ExternalCommand::computeCommandResult` is calling `node->getFileInfo(..)` for each of its outputs, llbuild can't distinguish whether the output node was truly intended to be a file or a directory.

If we want to avoid this check, the high-level build system needs to allow user to express their intent unambiguously. (i.e. high-level build system should only encode nodes with explicit trailing slash as directory node)..

@m-i-r-z-a what do you think?

lib/BuildSystem/BuildSystem.cpp
```diff
@@ -585,13 +588,34 @@ class DirectoryInputNodeTask : public Task {
...
-  virtual void provideValue(TaskInterface, uintptr_t inputID,
-                            const ValueType& value) override {
-    directorySignature = value;
+  virtual void provideValue(TaskInterface ti, uintptr_t inputID,
+                            const ValueType& valueData) override {
+    if (inputID == 0) {
+      directorySignature = valueData;
+    } else {
+      auto value = BuildValue::fromData(valueData);
+      // FIX ME (critical before merge): must check provided values are not erroneous
```
johari, line 612: In `DirectoryContentsTask::inputsAvailable` we have the following checks for corner cases:

```
    if (directoryValue.isMissingInput()) {
      ti.complete(BuildValue::makeMissingInput().toData());
      return;
    }

    if (directoryValue.isMissingOutput()) {
      // Rewrite a missing output as a missing input build value for this node, both conceptually and as a hedge against violating the downstream expectations of other rules.
      ti.complete(BuildValue::makeMissingInput().toData());
      return;
    }

    if (directoryValue.isFailedInput()) {
      ti.complete(BuildValue::makeFailedInput().toData());
      return;
    }

    // The input directory may be a 'mkdir' command, which can be cancelled or
    // skipped by the engine or the delegate. rdar://problem/50380532
    if (directoryValue.isSkippedCommand()) {
      ti.complete(BuildValue::makeSkippedCommand().toData());
      return;
    }
```

johari: - [ ] It'd be nice to have a test for erroneous cases too (e.g. "what happens when subpath of a directory input node is produced by a command that exits in error?")

lib/BuildSystem/BuildFile.cpp
```diff
@@ -94,6 +101,47 @@ static void dumpNode(llvm::yaml::Node* node, unsigned indent) {
...
 
+class OwnershipAnalysis {
+  std::map<BuildNode*, ExternalCommand*> paths;
+
+public:
+  const bool isUnownedNode(const BuildNode* node) {
+    StringRef curPath = node->getName();
+
+    while (!curPath.empty()) {
+      if (std::find_if(paths.begin(), paths.end(), [curPath](const std::pair<BuildNode*, ExternalCommand*>& pair) -> bool {
```
johari, line 112: nit: use a more descriptive name than "pair"

lib/BuildSystem/BuildFile.cpp
```diff
@@ -873,6 +921,61 @@ class BuildFileImpl {
...
+    });
+
+    OwnershipAnalysis ownershipAnalysis;
+
+    for (auto pair: outputNodesAndCommands) {
+      if (pair.first->isDirectory()) {
+        if (ownershipAnalysis.isUnownedNode(pair.first)) {
+          ownershipAnalysis.setOwner(pair.first, pair.second);
+        } else {
+          error("Multiple commands produce conflicting output.");
```
johari, line 955: Ensure the error message provides more context (e.g. which commands are in conflict?)

johari:
@swift-ci please smoke test

lib/BuildSystem/BuildFile.cpp
```diff
@@ -94,13 +100,158 @@ static void dumpNode(llvm::yaml::Node* node, unsigned indent) {
...
+
+  /// Establish ownerships
+  bool establishOwnerships() {
+    for (auto outputNodeAndCommand: outputNodesAndCommands) {
+      if (outputNodeAndCommand.first->isDirectory()) {
+        Command *owner = ownerOf(outputNodeAndCommand.first->getName());
+        if (owner == nullptr) {
+          setOwner(outputNodeAndCommand.first, outputNodeAndCommand.second);
+        } else if (owner == outputNodeAndCommand.second) {
+          // A path and some o fits subpaths are listed as output dependencies of a task.. Do nothing.
```
owenv, line 148:
```suggestion
          // A path and some of its subpaths are listed as output dependencies of a task.. Do nothing.
```

owenv (approved):

tests/BuildSystem/Build/directory-input-and-output.llbuild
```diff
@@ -0,0 +1,140 @@
...
+#  |   libZ.fake-h
+#  v
+# [TaskB] (modified copy; ALL CAPS)
+#  |
+#  v
+# headers-copy/
+#      libX.fake-h (will be automatically amended to outputs of TaskB)
+#  ,-- libY.fake-h
+#  |   libZ.fake-h
+#  v
```
johari, line 28: `libY.fake-h` will be automatically amended, not `libX.fake-h`

johari:
Linux/macOS CI failing with

```
/home/build-user/llbuild/lib/Commands/BuildSystemCommand.cpp:421:13: error: no matching constructor for initialization of 'llbuild::buildsystem::BuildFile'
  BuildFile buildFile(filename, delegate);
            ^         ~~~~~~~~~~~~~~~~~~
/home/build-user/llbuild/include/llbuild/BuildSystem/BuildFile.h:142:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 2 were provided
class BuildFile {
      ^
/home/build-user/llbuild/include/llbuild/BuildSystem/BuildFile.h:150:12: note: candidate constructor not viable: requires 3 arguments, but 2 were provided
  explicit BuildFile(StringRef mainFilename,
           ^
1 error generated.
```

- [ ] Trying to investigate CI failure.

owenv:
@johari  ^ I believe BuildSystemCommand.cpp is not built as part of llbuild.framework, but is part of the command line tool and various parts of the CMake build

johari:
@swift-ci please smoke test

dmbryson (approved):

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

include/llbuild/BuildSystem/BuildNode.h
```diff
@@ -84,6 +84,12 @@ class BuildNode : public Node {
     return exclusionPatterns;
   }
 
+  std::vector<StringRef> mustScanAfterPaths;
```
jakepetroules, line 87: This declaration should move up above with the other ivars so it doesn't have public visibility.

PR #851, opened 2022-12-02
rdar://99877341 (Check contents instead of modTime to determine if rule needs rerun)

johari:
This PR lands an opt-in support for content checksum comparison in FileInfo.

Normally a rule is marked for `NeedsToRun` if the timestamp of its input is different from the result from previous build.

In contrast to how llbuild currently works, using `ChecksumOnlyFileSystem` should cause llbuild to not mark a rule as `NeedsToRun` when modification time of input is different but checksum of the contents of the path that FileInfo is referring to matches the checksum from previous build.

### DIFF
--- a/include/llbuild/Basic/FileInfo.h
+++ b/include/llbuild/Basic/FileInfo.h
@@ -30,6 +30,12 @@
 #include <cstdint>
 #include <string>
 
+#ifdef __APPLE__
+#include <CommonCrypto/CommonDigest.h>
+#else
+#include "llvm/Support/MD5.h"
+#endif
+
 namespace llbuild {
 namespace basic {
 
@@ -60,6 +66,20 @@ struct FileTimestamp {
   }
 };
 
+struct FileChecksum {
+  uint8_t bytes[32] = {0};
+
+  bool operator==(const FileChecksum& rhs) const {
+    return (memcmp(bytes, rhs.bytes, sizeof(bytes)) == 0);
+  }
+
+  bool operator!=(const FileChecksum& rhs) const {
+    return !(*this==rhs);
+  }
+
+  static FileChecksum getChecksumForPath(const std::string& path);
+};
+
 /// File information which is intended to be used as a proxy for when a file has
 /// changed.
 ///
@@ -75,6 +95,8 @@ struct FileInfo {
   uint64_t size;
   /// The modification time of the file.
   FileTimestamp modTime;
+  /// The checksum of the file.
+  FileChecksum checksum = {};
 
   /// Check if this is a FileInfo representing a missing file.
   bool isMissing() const {
@@ -86,13 +108,15 @@ struct FileInfo {
 
   /// Check if the FileInfo corresponds to a directory.
   bool isDirectory() const;
-  
+
   bool operator==(const FileInfo& rhs) const {
     return (device == rhs.device &&
             inode == rhs.inode &&
             size == rhs.size &&
-            modTime == rhs.modTime);
+            modTime == rhs.modTime &&
+            checksum == rhs.checksum);
   }
+
   bool operator!=(const FileInfo& rhs) const {
     return !(*this == rhs);
   }
@@ -121,6 +145,20 @@ struct BinaryCodingTraits<FileTimestamp> {
 };
 
 template<>
+struct BinaryCodingTraits<FileChecksum> {
+  static inline void encode(const FileChecksum& value, BinaryEncoder& coder) {
+    for(int i=0; i<32; i++) {
+      coder.write(value.bytes[i]);
+    }
+  }
+  static inline void decode(FileChecksum& value, BinaryDecoder& coder) {
+    for(int i=0; i<32; i++) {
+      coder.read(value.bytes[i]);
+    }
+  }
+};
+
+template<>
 struct BinaryCodingTraits<FileInfo> {
   static inline void encode(const FileInfo& value, BinaryEncoder& coder) {
     coder.write(value.device);
@@ -128,6 +166,7 @@ struct BinaryCodingTraits<FileInfo> {
     coder.write(value.mode);
     coder.write(value.size);
     coder.write(value.modTime);
+    coder.write(value.checksum);
   }
   static inline void decode(FileInfo& value, BinaryDecoder& coder) {
     coder.read(value.device);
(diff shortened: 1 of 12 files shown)


johari:
@swift-ci please smoke test

include/llbuild/Basic/FileInfo.h
```diff
@@ -60,6 +60,24 @@ struct FileTimestamp {
   }
 };
 
+struct FileChecksum {
+  uint8_t bytes[32] = {0};
+
+  bool operator==(const FileChecksum& rhs) const {
+    for(int i=0; i<32; i++) {
```
owenv, line 67: replacing this with memcmp may be more efficient, even if these comparisons aren't a huge bottleneck

include/llbuild/Basic/FileSystem.h
```diff
@@ -153,6 +163,71 @@ class DeviceAgnosticFileSystem : public FileSystem {
...
+
+public:
+  explicit ChecksumOnlyFileSystem(std::unique_ptr<FileSystem> fs)
+    : impl(std::move(fs))
+  {
+  }
+
+  ChecksumOnlyFileSystem(const FileSystem&) LLBUILD_DELETED_FUNCTION;
+  void operator=(const ChecksumOnlyFileSystem&) LLBUILD_DELETED_FUNCTION;
+  DeviceAgnosticFileSystem &operator=(ChecksumOnlyFileSystem&& rhs) LLBUILD_DELETED_FUNCTION;
```
owenv, line 179:
```suggestion
  ChecksumOnlyFileSystem &operator=(ChecksumOnlyFileSystem&& rhs) LLBUILD_DELETED_FUNCTION;
```

include/llbuild/Basic/FileSystem.h
```diff
@@ -153,6 +163,71 @@ class DeviceAgnosticFileSystem : public FileSystem {
...
+
+    info.checksum = impl->getFileChecksum(path);
+
+    return info;
+  }
+
+  virtual FileInfo getLinkInfo(const std::string& path) override {
+    auto info = impl->getLinkInfo(path);
+
+    info.device = 0;
```
owenv, line 220: Right now, symlinks aren't using checksums. Should this avoid clearing the device and node numbers, so that if the underlying implementation is not a device-agnostic filesystem, we don't impose that behavior?

lib/Basic/FileInfo.cpp
```diff
@@ -63,3 +72,57 @@ FileInfo FileInfo::getInfoForPath(const std::string& path, bool asLink) {
...
   return result;
 }
+
+FileChecksum FileChecksum::getChecksumForPath(const std::string& path) {
+  FileChecksum result;
+
+  FileInfo fileInfo = FileInfo::getInfoForPath(path);
+  if (fileInfo.isMissing()) {
+    for(int i=0; i<32; i++) {
+      result.bytes[i] = 0;
```
owenv, line 82: nit: maybe prefer memset

lib/Basic/FileInfo.cpp
```diff
@@ -63,3 +72,57 @@ FileInfo FileInfo::getInfoForPath(const std::string& path, bool asLink) {
...
+    size_t bytesRead = 0;
+
+#ifdef __APPLE__
+    CC_SHA256_CTX ctx = {};
+    unsigned char temp_digest[CC_SHA256_DIGEST_LENGTH] = { 0 };
+    CC_SHA256_Init( &ctx );
+
+    if (file != NULL) {
+      while ((bytesRead = fread(buffer, 1, sizeof(buffer), file)) > 0) {
+        CC_SHA256_Update(&ctx, buffer, bytesRead );
```
owenv, line 99:
```suggestion
        CC_SHA256_Update(&ctx, buffer, bytesRead);
```

lib/Basic/FileInfo.cpp
```diff
@@ -63,3 +72,57 @@ FileInfo FileInfo::getInfoForPath(const std::string& path, bool asLink) {
...
+      }
+      fclose(file);
+    } else {
+      abort();
```
owenv, line 103: We should do something better than crashing here

products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -101,6 +101,21 @@ class CAPIFileSystem : public basic::FileSystem {
...
+  virtual basic::FileChecksum getFileChecksum(const std::string& path) override {
+    if (!cAPIDelegate.fs_get_file_checksum) {
+      return localFileSystem->getFileChecksum(path);
+    }
+
+    llb_fs_file_checksum_t file_checksum;
+    cAPIDelegate.fs_get_file_checksum(cAPIDelegate.context, path.c_str(), &file_checksum);
+
+    basic::FileChecksum result{};
+    for (int i=0; i<16; i++) {
```
owenv, line 113: Here a 16 byte checksum is assumed, but other areas assume 32 bytes

include/llbuild/Basic/FileSystem.h
johari: I can't see a blocker for symlinks to calculate checksums (but we do need to write tests for them). `ChecksumOnlyFileSystem::getLinkInfo` should look very similar to `ChecksumOnlyFileSystem::getFileInfo` and request for the checksum of the path.

Later on if we wanted to implement a FileInfoCache, we might want to resolve the symlink first to avoid multiple checksumming.

johari:
Todo:
* CAPI tests via gtest
* Swift tests via `$ swift test`
* Error handling for getChecksumForPath

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

include/llbuild/Basic/FileSystem.h
johari: I updated `getLinkInfo` so that on Unix we `readlink` and checksum the result of that.

On Windows.. I don't know.. I'm not even sure if `getLinkInfo` will be called at all or not. In absence of the notion of symlinks on that platform, I relied on `fread`ing the path and checksumming the content of that. We can revisit the Windows case later.

johari:
@swift-ci please smoke test

dmbryson (approved):

johari:
@swift-ci please smoke test

lib/Core/SQLiteBuildDB.cpp
```diff
@@ -77,7 +78,7 @@ class SQLiteBuildDB : public BuildDB {
   /// * 6: Added `ordinal` field for dependencies.
   /// * 5: Switched to using `WITHOUT ROWID` for dependencies.
   /// * 4: Pre-history
-  static const int currentSchemaVersion = 14;
+  static const int currentSchemaVersion = 15;
```
johari, line 81: We need to coordinate merging this PR with https://github.com/apple/swift-llbuild/pull/832

johari: Owen mentioned some support tools might need to be updated to accommodate database schema changes.

johari: Changed this to 16 and this is resolved now.

johari:
@swift-ci please smoke test

johari:
@swift-ci please smoke test

PR #853, opened 2023-01-06
Check error code when processing directories

dmbryson:
Under some circumstances, directory iterators encountering errors will not advance, leading to infinite looping. This change adds a check in the loop condition for getContents() and getFilteredContents() usages and propagates the error_code out to the caller.

Current callers/system behavior expect those errors to be silently ignored. Added a comment making that explicitly documented.

rdar://101717159


dmbryson:
@swift-ci please smoke test

dmbryson:
@swift-ci please smoke test macos

PR #855, opened 2023-01-24
Handle EINTR and EAGAIN when we `poll`

FranzBusch:
# Motivation
We are currently making calls to `poll` which could get interrupted or the resource could be temporarily not available. We should handle these cases.

# Modification
Wrap the calls to `poll` in a `wile` loop and try again if we get either `EINTR` or `EAGAIN`.

# Result
We are now handling signals while we are polling correctly.

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -1056,11 +1056,16 @@ void llbuild::basic::spawnProcess(
     assert(readfds[0].fd == outputPipeParentEnd.unsafeDescriptor());
     assert(readfds[1].fd == controlPipeParentEnd.unsafeDescriptor());
 
-    if (poll(readfds, nfds, -1) == -1) {
-      int err = errno;
-      delegate.processHadError(ctx, handle,
-          Twine("failed to poll (") + strerror(err) + ")");
-      break;
+    while (poll(readfds, nfds, -1) == -1) {
+        int err = errno;
+
+        if (err == EAGAIN || err == EINTR) {
+          continue;
+        } else {
+          delegate.processHadError(ctx, handle,
+            Twine("failed to poll (") + strerror(err) + ")"); 
+          return;
+        }
     }
 
     for (int i = 0; i < nfds; i++) {


lib/Basic/Subprocess.cpp
```diff
@@ -1056,11 +1056,20 @@ void llbuild::basic::spawnProcess(
     assert(readfds[0].fd == outputPipeParentEnd.unsafeDescriptor());
     assert(readfds[1].fd == controlPipeParentEnd.unsafeDescriptor());
 
-    if (poll(readfds, nfds, -1) == -1) {
-      int err = errno;
-      delegate.processHadError(ctx, handle,
-          Twine("failed to poll (") + strerror(err) + ")");
-      break;
+    while (true) {
+      if (poll(readfds, nfds, -1) == -1) {
```
Lukasa, line 1060: You can move this `if` condition into the `while` condition, and avoid writing the `else break` and `while true`.

weissi:
Please test

weissi:
test this please

dmbryson:
@swift-ci please smoke test

lib/Basic/Subprocess.cpp
```diff
@@ -1056,11 +1056,16 @@ void llbuild::basic::spawnProcess(
...
-      break;
+    while (poll(readfds, nfds, -1) == -1) {
+        int err = errno;
+
+        if (err == EAGAIN || err == EINTR) {
+          continue;
+        } else {
+          delegate.processHadError(ctx, handle,
+            Twine("failed to poll (") + strerror(err) + ")"); 
+          goto pollError;
```
dmbryson, line 1067: Can we just return here?   Why the `goto` ?

FranzBusch: I guess we can. I just wanted to have the exact same behaviour as before where we returned to after the while loop

dmbryson: I think the more direct return is better in this case and matches the coding style more cleanly.

dmbryson:
@swift-ci please smoke test

dmbryson (approved):

jakepetroules (approved):

PR #859, opened 2023-02-02
libllbuild: correct the macro usage for DLL builds

compnerd:
`_DLL` indicates that the MSVCRT version is the shared form rather than static.  This corresponds to `/MD` (and the lack to `/MT`).  `_WINDLL` is used to indicate that we are building statically or dynamically. This change is required to experiment with static builds of libllbuild with CMake.


compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):
LGTM

PR #861, opened 2023-02-06
llbuild: add missing parenthesis to macros

compnerd:
The value itself is the result of the macro, which was not parenthesized which made the usage of it incorrect in a few places where the negation was being checked and that bound more tightly than the equality check.


compnerd:
@swift-ci please smoke test

owenv (approved):

PR #862, opened 2023-02-07
Report when a task has a missing dynamically-requested input

owenv:
Currently, we don't have the necessary information to report which input is missing, but higher-level build systems can use other delegate callbacks to reconstruct that information. If no missing input is reported at all, we can easily end up in a situation where tasks are unexpectedly and silently cancelled, making debugging these sorts of issues difficult.

rdar://58130864

### DIFF
--- a/products/libllbuild/BuildSystem-C-API.cpp
+++ b/products/libllbuild/BuildSystem-C-API.cpp
@@ -309,8 +309,11 @@ public:
   virtual void commandCannotBuildOutputDueToMissingInputs(Command* command,
                Node* outputNode, SmallPtrSet<Node*, 1> inputNodes) override {
     if (cAPIDelegate.command_cannot_build_output_due_to_missing_inputs) {
-      auto str = outputNode->getName().str();
-      auto output = new CAPIBuildKey(BuildKey::makeNode(str));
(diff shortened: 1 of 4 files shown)


products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -309,8 +309,11 @@ class CAPIBuildSystemFrontendDelegate : public BuildSystemFrontendDelegate {
   virtual void commandCannotBuildOutputDueToMissingInputs(Command* command,
                Node* outputNode, SmallPtrSet<Node*, 1> inputNodes) override {
     if (cAPIDelegate.command_cannot_build_output_due_to_missing_inputs) {
-      auto str = outputNode->getName().str();
-      auto output = new CAPIBuildKey(BuildKey::makeNode(str));
+      CAPIBuildKey *output = nullptr;
```
owenv, line 312: All the changes outside of ExternalCommand.cpp are to account for the fact that the outputs list of a dynamic task will be empty

owenv:
@swift-ci please smoke test

johari (approved):

dmbryson (approved):

PR #867, opened 2023-02-22
rdar://105755236 (Make ProducedDirectoryNode propagate special cases similar to ProducedNode)

johari:
Fixes a downstream test failure that was due to `ProducedDirectoryNode` not propagating special cases similar to `ProducedNode`


johari:
@swift-ci please smoke test

jakepetroules (approved):

PR #868, opened 2023-02-23
[CMake][llbuildSwift] change the runpath to an architecture-specific directory on ELF platforms 

buttaface:
This is needed for apple/swift#63782, which changes the Unix toolchain to look for libraries in architecture-specific directories.

`get_swift_host_arch()` was copied over from the libdispatch CMake config and slightly updated and condensed.

### DIFF
--- a/products/llbuildSwift/CMakeLists.txt
+++ b/products/llbuildSwift/CMakeLists.txt
@@ -1,3 +1,37 @@
+# Returns the current architecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
+# with the sanitized host architecture name derived from CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
+  if("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "amd64|x86_64|s390x|wasm32")
+    set("${result_var_name}" "${CMAKE_SYSTEM_PROCESSOR}" PARENT_SCOPE)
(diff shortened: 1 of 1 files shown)


buttaface:
@etcwilde, now that this passed the CI on the linked compiler pull, looking for review so I can get this in before the upcoming 5.9 branch. Please don't merge though, as all 11 pulls will have to be merged together.

products/llbuildSwift/CMakeLists.txt
```diff
@@ -1,3 +1,37 @@
+# Returns the current architecture name in a variable
```
compnerd, line 1: Why replicate this rather than from `SwiftSupport.cmake` ?

etcwilde: LLBUILD doesn't have a `SwiftSupport.cmake` yet, but we should probably pull that from somewhere.
@compnerd is there a canonical version somewhere? The one in swift-system is a few lines longer from a quick glance, so probably handles more architectures; https://github.com/apple/swift-system/blob/main/cmake/modules/SwiftSupport.cmake.

buttaface:
Updated this pull to account for my just-merged #870.

products/llbuildSwift/CMakeLists.txt
buttaface: > The one in swift-system is a few lines longer from a quick glance, so probably handles more architectures;

I just checked and the only difference is that that one doesn't handle `wasm32`, whereas this one is missing the all-caps `ARM64`. Not particularly important, can always add it later if wanted, as I'd like to get this in with the other pulls before the 5.9 branch.

buttaface:
Closing since [the Swift devs want to take a different approach](https://github.com/apple/swift/pull/63782#pullrequestreview-1353429744).

PR #852, opened 2022-12-28
libllbuild: make the library static and internalise it

compnerd:
Convert libllbuild into static and avoid exporting the library for use in other projects.  The expected clients are only reliant on llbuildSwift.  This overall saves some space (~20K).  Additionally, it will simplify the redistribution in the toolchain by reducing the number of shared libraries while saving space.

PR #870, opened 2023-03-04
[CMake] Don't install libllbuild in the Swift toolchain now that it's statically linked

buttaface:
Now that `libllbuildSwift` is statically linked against `libllbuild` with #852, remove the runpath and don't install it. Also, the Android build started failing, so set the appropriate flag for it to work again.

Previously, we installed `libllbuild.so` in the Swift toolchain alongside `libllbuildSwift.so` and set the ELF runpath so it knew it was in the same directory:
```
> ls swift-5.7.3-RELEASE-ubuntu20.04/usr/lib/swift/pm/llbuild/
libllbuild.so  libllbuildSwift.so
> readelf -d swift-5.7.3-RELEASE-ubuntu20.04/usr/lib/swift/pm/llbuild/libllbuildSwift.so |ag runpath
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../../linux]
```
Now that @compnerd made it link statically, we're wrongly shipping the static libllbuild and still have the now unneeded runpath:
```
> ls swift-DEVELOPMENT-SNAPSHOT-2023-03-01-a-ubuntu20.04/usr/lib/swift/pm/llbuild/
libllbuild.a  libllbuildSwift.so
> readelf -d swift-DEVELOPMENT-SNAPSHOT-2023-03-01-a-ubuntu20.04/usr/lib/swift/pm/llbuild/libllbuildSwift.so | ag runpath
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../../linux]
```
Also, I started seeing [cross-compilation failures on my daily Android CI](https://github.com/buttaface/swift-android-sdk/actions/runs/4250967892/jobs/7392741185) with this move to static linking, which this Android change now fixes, buttaface/swift-android-sdk#94.

### DIFF
--- a/products/libllbuild/CMakeLists.txt
+++ b/products/libllbuild/CMakeLists.txt
@@ -57,14 +57,6 @@ install(TARGETS libllbuild
   RUNTIME DESTINATION bin
   COMPONENT libllbuild)
 
-if(Swift IN_LIST LLBUILD_SUPPORT_BINDINGS)
-  if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-    install(TARGETS libllbuild
-      DESTINATION lib/swift/pm/llbuild
-      COMPONENT libllbuildSwift)
-  endif()
-endif()
-
 add_custom_target(install-libllbuild
                   DEPENDS libllbuild
                   COMMENT "Installing libllbuild..."
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,9 @@ else()
   endif()
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL Android)
+  set(CMAKE_HAVE_LIBC_PTHREAD TRUE)
+endif()
 set(THREADS_PREFER_PTHREAD_FLAG FALSE)
 find_package(Threads REQUIRED)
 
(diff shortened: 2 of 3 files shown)


buttaface:
@johari, would be good to fix this before the upcoming 5.9 branch, let me know what you think.

johari:
cc/ @owenv @jakepetroules @dmbryson and @compnerd

products/libllbuild/CMakeLists.txt
```diff
@@ -57,14 +57,6 @@ install(TARGETS libllbuild
   RUNTIME DESTINATION bin
   COMPONENT libllbuild)
```
compnerd, line 58: I think that we should nix this part as well.  Although it doesn't really matter much for Windows distributions as the packaging of the installer will drop it, it does help keep things clean.  Does it make sense to continue distributing the headers?  Or should we also be dropping those?  Sorry about this mess I created!

compnerd (approved):
I don't see a problem with the diff, I think that it can go further though and reduce even more of the installation logic.

Thanks for cleaning this up @buttaface!

products/libllbuild/CMakeLists.txt
buttaface: I don't know if the rest is needed or not, haven't looked into it.

buttaface:
@johari, would you please run the CI and merge?

johari:
@swiftci please smoke test

buttaface:
I believe it's called @swift-ci.

dmbryson:
@swift-ci please smoke test

johari:
Let's ensure it doesn't break downstream dependencies before merging it

buttaface:
Ping, can be merged now?

johari:
Making a quick inquiry with @bnbarham and we should be good to go.

PR #871, opened 2023-03-07
Crash in commandCannotBuildOutputDueToMissingInputs reported by ubsan

jakepetroules:
rdar://106367348

### DIFF
--- a/products/libllbuild/BuildSystem-C-API.cpp
+++ b/products/libllbuild/BuildSystem-C-API.cpp
@@ -351,7 +351,7 @@ public:
       }
     }
 
-    llb_build_key_t** data() { return (llb_build_key_t **)&keys[0]; }
+    llb_build_key_t** data() { return (llb_build_key_t **)keys.data(); }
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci please smoke test

products/libllbuild/BuildSystem-C-API.cpp
```diff
@@ -351,7 +351,7 @@ class CAPIBuildSystemFrontendDelegate : public BuildSystemFrontendDelegate {
       }
     }
 
-    llb_build_key_t** data() { return (llb_build_key_t **)&keys[0]; }
+    llb_build_key_t** data() { return (llb_build_key_t **)keys.data(); }
```
johari, line 354: Does this actually fix the problem?

From `[1]`:
> If [size()](https://en.cppreference.com/w/cpp/container/vector/size) is ​0​, data() may or may not return a null pointer.

`[1]`: https://en.cppreference.com/w/cpp/container/vector/data

johari (approved):

PR #873, opened 2023-03-28
Various set of SPM based build updates for Windows

compnerd:
Update the build for Windows to get SPM based builds working better.  This should make it easier to work with llbuild and SPM when developing with SPM.

PR #874, opened 2023-03-28
Various build fixes to get llbuild to build with SPM on Windows

compnerd:
This should be rebased upon #873.  It simultaneously requires apple/swift-package-manager#6345 to get the fix for the `..` header search path handling corrected.  Update the build rules for GTest to avoid duplicate definitions, add build rules for GMock, update the tests to use GMock as a dependency.  This set of updates enables building llbuild with SPM on Windows (with some minor hand holding for SQLite and overlinking of swiftrt.obj).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -176,6 +176,10 @@ let package = Package(
                 "gtest-test-part.cc",
                 "gtest-typed-test.cc",
                 "gtest.cc",
+            ],
+            cxxSettings: [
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

jakepetroules (approved):

PR #881, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

### DIFF
--- a/include/llvm/Config/config.h
+++ b/include/llvm/Config/config.h
@@ -18,7 +18,7 @@
 #define ENABLE_CRASH_OVERRIDES 1
 
 /* Define to 1 if you have the `backtrace' function. */
-#if !defined(__ANDROID__) && !defined(__OpenBSD__) && defined(__GLIBC_PREREQ)
+#if __has_include(<execinfo.h>)
 #define HAVE_BACKTRACE TRUE
 #endif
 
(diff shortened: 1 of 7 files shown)


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

include/llvm/Config/config.h
```diff
@@ -18,7 +18,7 @@
 #define ENABLE_CRASH_OVERRIDES 1
 
 /* Define to 1 if you have the `backtrace' function. */
-#if !defined(__ANDROID__) && !defined(__OpenBSD__)
+#if !defined(__ANDROID__) && !defined(__OpenBSD__) && defined(__GLIBC_PREREQ)
```
al45tair, line 21: Maybe this should be `#if __has_include(<execinfo.h>)`? That's where `backtrace` is declared, and would avoid "knowing" that it doesn't exist on Android or OpenBSD.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

PR #887, opened 2023-07-28
Add new Swift API for creating CustomTasks from byte arrays instead of strings

owenv:
Custom task data is often unstructured, and won't be valid UTF-8. Exposing API that works with data instead of strings avoids the need for hacks like base-64 encoding when creating them using the Swift API, and makes it easier to avoid UTF8 encoding overhead.

Also, adopt the use of a Swift user module version in llbuild, allowing clients to more easily stage in use of new API

rdar://112936539


owenv:
@swift-ci please test

owenv:
@swift-ci please smoke test

jakepetroules (approved):

PR #899, opened 2023-12-07
build: bump minimum required CMake version

compnerd:
Update the minimum CMake version to 3.19.  Swift requires 3.19.x for the build, so update the minimum version to match.  This allows us to drop some of the workarounds that we had for older CMake releases.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

jakepetroules (approved):

PR #910, opened 2024-04-23
Escape = characters in interpolated variables in sed expressions since = is used as the delimiter character

jakepetroules:
This prevents the script from failing when any of these paths contain an = character.


jakepetroules:
@swift-ci please smoke test

owenv (approved):

PR #911, opened 2024-04-23
Mark relevant script phases inputs/outputs or as always out of date

jakepetroules:
Most of these don't declare inputs/outputs and it would be nontrivial to do so at this time.


jakepetroules:
@swift-ci please smoke test

PR #912, opened 2024-04-23
Fix the tests build by adding missing availability

jakepetroules:


jakepetroules:
@swift-ci please smoke test

PR #915, opened 2024-05-02
When computing a directory tree signature, don't follow symlinks to parent directories

owenv:
This could cause us to get stuck in an infinite loop when scanning such a symlink as part of a directory tree


owenv:
@swift-ci please smoke test

dmbryson (approved):

PR #919, opened 2024-07-03
Import new Android overlay

finagolfin:
This was recently added in swiftlang/swift#72161 and swiftlang/swift#72634. I've been building it with this patch on my daily Android CI since then, finagolfin/swift-android-sdk#151, and all the same tests pass as before when I build natively on Android too.

@owenv, should be an easy review.


owenv (approved):
LGTM

owenv:
@swift-ci please smoke test

PR #923, opened 2024-07-15
Remove invalid path check

jakepetroules:
These paths are expected to be absolute, but that should be handled by the `fs` object.


jakepetroules:
@swift-ci please smoke test

owenv (approved):

dmbryson (approved):

PR #928, opened 2024-07-26
Revert "Provide more powerful `isResultValid` API"

dmbryson:
Reverts swiftlang/swift-llbuild#926

Caused downstream failures that need to be investigated.


dmbryson:
@swift-ci please smoke test

jakepetroules (approved):

PR #930, opened 2024-08-06
Add support for building on Windows using SwiftPM

jakepetroules:
This adds a dependency on the new swift-toolchain-sqlite package, which allows llbuild to build using `swift build` on Windows (including tests) without any additional flags or preinstalled dependencies.

rdar://133338086

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,9 +1,10 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.3
 
 // This file defines Swift package manager support for llbuild. See:
 //  https://github.com/swiftlang/swift-package-manager/tree/master/Documentation
 
 import PackageDescription
+import class Foundation.ProcessInfo
 
 let package = Package(
     name: "llbuild",
@@ -68,7 +69,7 @@ let package = Package(
             name: "llbuildSwift",
             dependencies: ["libllbuild"],
             path: "products/llbuildSwift",
-            exclude: []
+            exclude: ["CMakeLists.txt"]
         ),
 
         // MARK: Components
@@ -80,9 +81,14 @@ let package = Package(
         ),
         .target(
             name: "llbuildCore",
-            dependencies: ["llbuildBasic"],
+            dependencies: [
+                "llbuildBasic",
+                .product(name: "CSQLite", package: "swift-toolchain-sqlite", condition: .when(platforms: [.windows])),
+            ],
             path: "lib/Core",
-            linkerSettings: [.linkedLibrary("sqlite3")]
+            linkerSettings: [
+                .linkedLibrary("sqlite3", .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .visionOS, .macCatalyst, .linux]))
+            ]
         ),
         .target(
             name: "llbuildBuildSystem",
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -213,6 +221,9 @@ let package = Package(
                 .headerSearchPath(".."),
                 .headerSearchPath("../include"),
                 .headerSearchPath("../../googletest/include"),
+            ],
+            linkerSettings: [
+                .linkedLibrary("swiftCore", .when(platforms: [.windows])), // for swift_addNewDSOImage
```
jakepetroules, line 226: Have to add this otherwise I get linker errors about "missing symbol swift_addNewDSOImage" -- @compnerd any chance you know why this is an issue on Windows (and WebAssembly) in particular?

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please smoke test

Package.swift
```diff
@@ -29,6 +29,9 @@ let package = Package(
             name: "llbuildAnalysis",
             targets: ["llbuildAnalysis"]),
     ],
+    dependencies: [
```
owenv, line 32: We should start respecting the SWIFTCI_USE_LOCAL_DEPS env var so that we can build against a local copy of the dependency in Swift CI

owenv: Example from swift-driver: https://github.com/swiftlang/swift-driver/blob/6a507035ea6a75dc865cc68a7e69f8f42c176595/Package.swift#L161

jakepetroules: Good point, added.

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please test

owenv:
The current round of CI failures should resolve once the new repo is added to update-checkout and related configs

jakepetroules:
> The current round of CI failures should resolve once the new repo is added to update-checkout and related configs

Opened swiftlang/swift#75743 for that

owenv (approved):
LGTM once CI is configured to check everything out

Package.swift
compnerd: It is because you are using swiftc as the linker driver. This will add in the swift registrar (`swiftrt.obj` or `swiftrt.o`) which requires that you overlink against swiftCore. I think that we want to build sqlite statically and not as a DLL.

jakepetroules: Ah, got it. I think in the longer term we probably should change the build system to use swift as the linker driver only if there's any Swift sources built in the target, otherwise clang/clang++?

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please test Windows

jakepetroules:
@swift-ci please test Windows

dmbryson (approved):

Package.swift
compnerd: Yes, that would make sense. We _could_ also use `-nostdlib` as a flag for the driver to exclude the registrar and the runtime.

Package.swift
```diff
@@ -233,6 +242,16 @@ let package = Package(
     cxxLanguageStandard: .cxx14
 )
 
+if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
```
rauhul, line 245: not sure if this is gated by a tools version thats too old, but I think you should be able to use `Context.environment` instead of pulling in `Foundation`

jakepetroules: Didn't know that existed! I guess it's been there since Swift 5.6. But we should probably update this systematically across all apple/swiftlang repos, it appears in a _lot_ of places.

Issue #936, opened 2024-09-05
Builds are completely serialized on Windows


jakepetroules:
This is a regression introduced by #597, where the following line https://github.com/swiftlang/swift-llbuild/commit/3900ed4e80d91d07eb8c502b7213b551ffbdbb81#diff-bd65cf12ed29b52c530ff04d2c678eb66883eddfcb44d5dbb8fe499f25865e93R219 presumably meant to use `std::max` instead of `std::min`.

The impact of this is that `estimateTaskLimits` falls back to a default case of returning only a single lane, regardless of the number of requested lanes, and consequently all builds are serialized.

PR #597, opened 2019-11-01
Fix "Too many open files" if too small of a system limit is encountered

vlm:
* The idea of this PR is that we can estimate the concurrency limits afforded by the system by examining its max open files limit. Based on that we set the number of lanes and the number of background tasks.
* This PR requires https://github.com/apple/swift-llbuild/pull/596 for correctness, but there's no code interdependency.

rdar://55670543

PR #937, opened 2024-09-05
Improve build performance by a factor of $CORE_COUNT

jakepetroules:
This fixes a regression introduced by #597, where the following line 3900ed4#diff-bd65cf12ed29b52c530ff04d2c678eb66883eddfcb44d5dbb8fe499f25865e93R219 presumably meant to use std::max instead of std::min.

The impact of this is that estimateTaskLimits falls back to a default case of returning only a single lane, regardless of the number of requested lanes, and consequently all builds are serialized.

Now, we simply return the value of _getmaxstdio directly. The documentation does not indicate that this function can fail or return a negative or zero value, and attempting to set the value with _setmaxstdio will terminate the process if given any value less than _IOB_ENTRIES (3).

Closes: #936

### DIFF
--- a/lib/Basic/PlatformUtility.cpp
+++ b/lib/Basic/PlatformUtility.cpp
@@ -220,8 +220,7 @@ int sys::write(int fileHandle, void *destinationBuffer,
 // Get the current process' open file limit. Returns -1 on failure.
 llbuild_rlim_t sys::getOpenFileLimit() {
 #if defined(_WIN32)
-  int value = _getmaxstdio();
-  return std::min(0, value);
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci please test

compnerd (approved):
Thank you!

owenv (approved):

vlm (approved):

dschaefer2 (approved):

dmbryson (approved):

dmbryson:
@swift-ci please smoke test

PR #942, opened 2024-09-10
Tritium README tweaks

dmbryson:
- Fixup image link
- Disambiguate system unification from graph unification


dmbryson:
@swift-ci please smoke test

PR #949, opened 2024-10-28
Add some feature flags to control use of optional features in the build

jakepetroules:
This will allow building llbuild with a Swift Static SDK using Musl libc, using SwiftPM.


dmbryson (approved):

jakepetroules:
@swift-ci please test windows

PR #952, opened 2024-11-15
[Tritium] Fix label target name handling

dmbryson:


dmbryson:
@swift-ci please smoke test

PR #954, opened 2024-11-17
Fix the SwiftPM build on Android

jakepetroules:


jakepetroules:
@swift-ci please smoke test

dmbryson (approved):

finagolfin:
What is this all about: any reason you're adding all these Android changes lately, @jakepetroules? You might want to run it by me first, as this affects the native toolchain for Android that I distribute.

Specifically, I've been avoiding adding any mention of the `libandroid-spawn` library, which is specific to the Termux app, because it's not needed for Android API 29 or later. Also, the `curses` dependency could be cross-compiled for Android, as the Termux app does, so I linked that here.

The SQLite change is good, I was planning on submitting that myself after more testing.

jakepetroules:
Hey @finagolfin, sorry I'm just seeing this now, I might've missed the email notification back in November.

To answer your question: just trying to test on a wider variety of platforms these days for good measure, since I know you and others are actively working on Android support for Swift.

I will try to keep you in mind for reviews going forward and am happy to delegate to your expertise if adding the `libandroid-spawn` library isn't the right thing to do. Would you mind preparing a PR to make any adjustments needed in this area?

Unfortunately I forgot what the detail was at this point but something in the pipeline was requiring a specific Android API version to be used throughout the build stack so I used some value < 29; that's something we probably need to parameterize better?

finagolfin:
> Would you mind preparing a PR to make any adjustments needed in this area?

I just [built the latest snapshot of the 6.1 toolchain for Android](https://github.com/finagolfin/termux-packages/actions/runs/12837102931) and this pull did not affect it, so I think we can leave it in for now.

> Unfortunately I forgot what the detail was at this point but something in the pipeline was requiring a specific Android API version to be used throughout the build stack so I used some value < 29; that's something we probably need to parameterize better?

You're using my SDK bundle to build for Android? It currently only supports Android API 24, but I recently tried modifying the destination config in that bundle manually after installation and it is easy to add the paths for it to work with later APIs too, as the bundled Swift runtime and Termux libraries built against API 24 should continue working for later APIs too.

Anyway, thanks for buildng for Android and all your review of my pulls for this port over the years: I'm just a ping away on any further Android changes.

jakepetroules:
> You're using my SDK bundle to build for Android?

Yep. Any plans to make it support arbitrary API levels instead of being fixed to 24? Is that feasible?

finagolfin:
Unfortunately, since the API level is specified in the triple, I'd have to [add 18 more lines like this](https://github.com/finagolfin/swift-android-sdk/blob/2c349c5bb91feb87359fc85ee8ca4e0908c80241/.github/workflows/sdks.yml#L227) to the JSON config for every API level greater than 24 supported, even though all would build against the same Android swiftmodules and link against the same Swift runtime libraries built against API 24.

I suppose it's not a big deal since it's an easy fix and nobody will ever look at that then giant JSON file, but such a hack still bothers me. 😉

I'll open an issue soliciting feedback from users of my Android SDK bundle, thanks for asking.

PR #960, opened 2024-12-05
[llbuild3] Fix an issue which caused action cache misses to prevent b…

### LINKED ISSUE #958: [llbuild3] Builds which get a nil response from a TActionCache implementation never proceed
With the current llbuild3 sources, builds which get a nil response from a TActionCache implementation never proceed -- they wait for further tasks endlessly. This appears to be because the necessary callback isn't called if there's a cache miss. The TActionCache protocol implementation of ExtActionCache should be updated to properly call the callback when there's a cache miss.

cwakamo:
…uilds from completing.

The ExtActionCache implementation for the TActionCache protocol failed to call the callback on a cache miss. This meant that cache misses, as are the case in fully-clean builds, would never allow the build to proceed to any actual work.

This commit resolves #958.

### DIFF
--- a/src/llbuild3/core/ActionCache.swift
+++ b/src/llbuild3/core/ActionCache.swift
@@ -44,6 +44,9 @@ extension TActionCache {
             let bytes = try value.serializedData()
             handler(std.string(fromData: bytes), std.string())
             return
+          } else {
+            handler(std.string(), std.string())
+            return
           }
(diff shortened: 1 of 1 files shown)


cwakamo:
@swift-ci please smoke test

dmbryson (approved):

PR #961, opened 2024-12-05
[llbuild3] Add support for passing a TCASDatabase to TEngine.

### LINKED ISSUE #959: [llbuild3] Unable to create a TEngine with a TCASDatabase instance
A recent llbuild3 change made it possible to define an implementation of the TCASDatabase protocol, but there's no way to actually _use_ TCASDatabase implementations with TEngine or any other part of llbuild3.

cwakamo:
Exposed the existing `makeExtCASDatabase` function to Swift via SwiftAdaptors.hpp. Added a `casDB` parameter to `TEngine`'s main initializer, and used `makeExtCASDatabase` to make the necessary `CASDatabaseRef` when `casDB` is not nil.

This matches the behavior of the `TActionCache?` parameter to `TEngine`'s main initializer.

This commit resolves #959.

### DIFF
--- a/products/llbuild3/SwiftAdaptors.hpp
+++ b/products/llbuild3/SwiftAdaptors.hpp
@@ -128,6 +128,7 @@ struct ExtCASDatabase {
 
 class CASDatabase;
 typedef std::shared_ptr<CASDatabase> CASDatabaseRef;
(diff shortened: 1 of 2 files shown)


cwakamo:
@swift-ci please smoke test

dmbryson (approved):

PR #968, opened 2025-01-24
build: rename `LLVMDemangle` to `llvmDemangle`

compnerd:
This mirrors the change to `LLVMSupport`. While this is not a robust change, it does make the build more uniform. We should instead be adding inline namespaces and renaming the library as we do in the Swift runtime.


compnerd:
@swift-ci please smoke test

PR #970, opened 2025-01-27
Fix CMake warnings in llbuild

etcwilde:
The llbuild project was emitting a few warnings from the CMake build. This was because it was relying on deprecated modules and behaviors. These patches fix both.

The `PythonInterp` module is deprecated in favor of `Python`, `Python2`, or `Python3`. Given that `llvm-lit` no longer works with Python versions at least below 3.7, I've gone ahead and updated this to look for Python 3.7 or newer when testing is enabled.

The project was also using the CMP0037 `OLD` behavior to silence the warning about the `test` target name being reserved. As of CMake 3.11 (llbuild requires 3.19 or newer), it is only reserved when using the CMake testing infrastructure, which `llbuild` does not. Now, setting this version to `OLD` results in a warning about the old behavior being deprecated. It isn't needed, so I've removed it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ set(LLBUILD_LIBDIR_SUFFIX "${LIB_SUFFIX}" CACHE STRING "Set default library fold
 if(BUILD_TESTING)
   find_package(Lit)
   find_package(FileCheck)
-  find_package(PythonInterp)
+  find_package(Python3 3.7 COMPONENTS Interpreter)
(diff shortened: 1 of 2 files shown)


etcwilde:
@swift-ci please test

compnerd (approved):

dmbryson (approved):

etcwilde:
@swift-ci please smoke test Linux

etcwilde:

PR #976, opened 2025-02-23
[llbuild3] Flesh out handling of requests in cached task transitions

dmbryson:


dmbryson:
@swift-ci please smoke test

PR #978, opened 2025-03-11
[llbuild3] Add Ackermann test

dmbryson:
Also adds some basic perf counters to the engine


dmbryson:
@swift-ci please smoke test

PR #979, opened 2025-04-05
Update OpenBSD support for swiftpm.

3405691582:
Some of this ends up being in common with the FreeBSD stanza, so lightly refactor.


3405691582:
@michael-yuji this should be a no-op, but just FYSA.

3405691582:
@swift-ci please smoke test.

dmbryson (approved):

PR #981, opened 2025-04-30
Add support for thread-safe chdir in process spawning

jakepetroules:
This is needed as a fallback to support Amazon Linux 2 and OpenBSD.

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -48,27 +48,6 @@
 #ifdef __APPLE__
 #include <pthread/spawn.h>
 #include "TargetConditionals.h"
-#if !TARGET_OS_IPHONE
-extern "C" {
-  // Provided by System.framework's libsystem_kernel interface
-  extern int __pthread_chdir(const char *path);
-  extern int __pthread_fchdir(int fd);
-}
-
-/// Set the thread specific working directory to the given path.
-int pthread_chdir_np(const char *path)
-{
-  return __pthread_chdir(path);
-}
-
-/// Set the thread specific working directory to that of the given file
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
Draft for now as I'm not confident this is safe at all.

jakepetroules:
@swift-ci test

dmbryson:
Unfortunately I think this is likely to run afoul of open file descriptor inheritance. I think we may want to use a trampoline binary to accomplish this.

jakepetroules:
> Unfortunately I think this is likely to run afoul of open file descriptor inheritance. I think we may want to use a trampoline binary to accomplish this.

I think that's true even for the current posix_spawn implementation because POSIX_SPAWN_CLOEXEC_DEFAULT is an Apple-specific extension.

jakepetroules:
@swift-ci test

jakepetroules:
OK, I'm fairly confident this is sufficiently close to the existing posix_spawn based code path in llbuild (acknowledging that the posix_spawn path is also probably not perfect in one way or another) so as to get us unblocked for Amazon Linux 2 and OpenBSD.

Please note that this does NOT change behavior for macOS, Linux with a sufficiently new (2.29+) Glibc, or any other platform which has posix_spawn_file_actions_addchdir[_np].

owenv (approved):

dmbryson (approved):

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci smoke test

jakepetroules:
@swift-ci test windows

jakepetroules:
@swift-ci test windows platform

PR #984, opened 2025-05-06
[FreeBSD] Fix target conditionals for FreeBSD aarch64 builds

CodingCarpincho:
Despite being declared in the header unistd.h, the function `sbrk` is not available on FreeBSD when using the aarch64 architecture because FreeBSD made the decision to drop support for this deprecated function when porting the system to ARM.

The configuration script incorrectly hard-codes the support for this function as being always available, which is not correct. We need to ensure that it is not used on FreeBSD aarch64.

### DIFF
--- a/include/llvm/Config/config.h
+++ b/include/llvm/Config/config.h
@@ -131,7 +131,11 @@
 /* #undef HAVE_LSEEK64 */
 
 /* Define to 1 if you have the `mallctl' function. */
+#if defined(__FreeBSD__)
+#define HAVE_MALLCTL 1
+#else
 /* #undef HAVE_MALLCTL */
+#endif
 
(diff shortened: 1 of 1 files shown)


jakepetroules (approved):

jakepetroules:
@swift-ci smoke test

jakepetroules:
@swift-ci smoke test linux

dmbryson (approved):

jakepetroules:
@swift-ci smoke test linux

jakepetroules:
@swift-ci smoke test linux

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci smoke test linux

PR #985, opened 2025-05-07
Return ENOSYS for posix_spawn_file_actions_addchdir on older Android versions

marcprux:
This fixes the Android build error introduced in https://github.com/swiftlang/swift-llbuild/commit/9087bdf8c7d4c95402d6d811745ca5ff3e149741.

According to https://android.googlesource.com/platform/bionic/+/master/docs/status.md, `posix_spawn_file_actions_addchdir_np` was only added in API level 34. There seems to be no equivalent for earlier versions, so we just return `ENOSYS` for lower API levels (such as API level 29, which the Android SDK is targeting).

### DIFF
--- a/lib/Basic/Subprocess.cpp
+++ b/lib/Basic/Subprocess.cpp
@@ -102,6 +102,9 @@ static int posix_spawn_file_actions_addchdir_polyfill(posix_spawn_file_actions_t
   //  - FreeBSD 13.1 (May 2022)
   //  - Android 14 (October 2023)
   return posix_spawn_file_actions_addchdir_np((posix_spawn_file_actions_t *)file_actions, path);
+#elif defined(__ANDROID__)
+  //  - Android < 14
+  return ENOSYS;
 #else
   // Standardized posix_spawn_file_actions_addchdir version (POSIX.1-2024, June 2024) available in:
   //  - Solaris 11.4 (August 2018)


marcprux:
Thoughts, @finagolfin and @jakepetroules?

finagolfin:
@swift-ci test

dmbryson (approved):

dmbryson:
@swift-ci please smoke test

lib/Basic/Subprocess.cpp
```diff
@@ -102,6 +102,9 @@ static int posix_spawn_file_actions_addchdir_polyfill(posix_spawn_file_actions_t
   //  - FreeBSD 13.1 (May 2022)
   //  - Android 14 (October 2023)
   return posix_spawn_file_actions_addchdir_np((posix_spawn_file_actions_t *)file_actions, path);
+#elif defined(__ANDROID__)
```
jakepetroules, line 105: Actually, would you mind moving this up below the glibc case at the top, and write like:

```c
#elif defined(__ANDROID__) && __ANDROID_API__ < 34
```

...to keep the "missing" cases consolidated at the top?

marcprux: > Actually, would you mind moving this up below the glibc case at the top

Done!

finagolfin: You missed moving it up. I'd also prefer if you combined it with the glibc and OpenBSD cases, since all return the same error.

You can also remove checking the API level is greater than 34 for the affirmative case above.

marcprux: > I'd also prefer if you combined it with the glibc and OpenBSD cases, since all return the same error.

You mean like this? It seems rather tortuous… 

```
#if (defined(__GLIBC__) && !__GLIBC_PREREQ(2, 29)) || (defined(__OpenBSD__)) || (defined(__ANDROID__) && __ANDROID_API__ < 34)
  return ENOSYS;
```

finagolfin: Yes, either way, Jake asked to move it up above. I prefer it combined like that too.

lib/Basic/Subprocess.cpp
```diff
@@ -79,20 +79,19 @@ int pthread_fchdir_np(int fd)
 // Implementation mostly copied from _CFPosixSpawnFileActionsChdir in swift-corelibs-foundation
 static int posix_spawn_file_actions_addchdir_polyfill(posix_spawn_file_actions_t * __restrict file_actions,
                                                       const char * __restrict path) {
-#if defined(__GLIBC__) && !__GLIBC_PREREQ(2, 29)
+#if (defined(__GLIBC__) && !__GLIBC_PREREQ(2, 29)) || (defined(__OpenBSD__)) || (defined(__ANDROID__) && __ANDROID_API__ < 34)
   // Glibc versions prior to 2.29 don't support posix_spawn_file_actions_addchdir_np, impacting:
   //  - Amazon Linux 2 (EoL mid-2025)
+  // Currently missing as of:
+  //  - OpenBSD 7.5 (April 2024)
+  //  - Android <= 14
```
jakepetroules, line 87: While you're here, could you move the QNX 8 comment here too?. The function is actually missing on that platform. It is mistakenly labeled as supported due to an error in earlier refactoring.

lib/Basic/Subprocess.cpp
```diff
@@ -79,20 +79,19 @@ int pthread_fchdir_np(int fd)
...
   // Conditionally available on macOS if building with a deployment target older than 10.15
   if (__builtin_available(macOS 10.15, *)) {
     return posix_spawn_file_actions_addchdir_np(file_actions, path);
   }
   return ENOSYS;
-#elif defined(__OpenBSD__)
-  // Currently missing as of:
-  //  - OpenBSD 7.5 (April 2024)
-  return ENOSYS;
 #elif defined(__GLIBC__) || defined(__APPLE__) || defined(__FreeBSD__) || (defined(__ANDROID__) && __ANDROID_API__ >= 34) || defined(__musl__)
```
jakepetroules, line 95: The __ANDROID_API__ >= 34 bit is no longer needed.

lib/Basic/Subprocess.cpp
marcprux: I assume you also want to add a check for `defined(__QNX__)` for that block? I've moved the comment and added the check in https://github.com/swiftlang/swift-llbuild/pull/985/commits/2cfc9eb97094051f7fef4038765c72ea5f7e1073

finagolfin:
@swift-ci please smoke test

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci smoke test linux

finagolfin:
@swift-ci smoke test linux

finagolfin:
@swift-ci smoke test linux

finagolfin:
@swift-ci smoke test linux

finagolfin:
@swift-ci smoke test windows

PR #988, opened 2025-05-23
llbuildSwift: wire up dependencies only when available

compnerd:
In the case that `Foundation` and `dispatch` are not passed along, wiring up the dependencies results in an injected `-lFoundation.lib` and `-lswiftDispatch.lib` on Windows. Only add the dependencies if we have the associated CMake targets.


compnerd:
@swift-ci please test

dmbryson:
@swift-ci please smoke test

compnerd:
@swift-ci please test Windows platform

compnerd:
@swift-ci please test Windows platform

PR #992, opened 2025-06-02
[6.2] fix: typo in the word `dependency`

jakepetroules:
Bring this back to 6.2 as well so swift-build can adopt the corrected naming sooner rather than later.


jakepetroules:
@swift-ci test

dmbryson (approved):

PR #1003, opened 2025-07-22
Remove `pipes` module from `pseudo-swiftc` test

sookach:
The pipes module was removed in Python 3.13. build-script tripped on this test when trying to build a toolchain on Fedora 41.

### DIFF
--- a/tests/SwiftBuildTool/Inputs/pseudo-swiftc
+++ b/tests/SwiftBuildTool/Inputs/pseudo-swiftc
@@ -4,7 +4,7 @@
 import argparse
 import json
 import os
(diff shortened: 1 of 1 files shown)


jakepetroules (approved):

etcwilde (commented):
Do we have anything that requires something earlier than python 3.3? If so, we may need to do something like:

```python
try:
    from shlex import quote as _quote
except ImportError:
    from pipes import quote as _quote
```
and then call `_quote`. If not, then I think that this is fine.

shahmishal:
@swift-ci smoke test

PR #1004, opened 2025-08-07
use Win32 file APIs and widePath for file ops to support long filenames

daveinglis:


daveinglis:
@swift-ci test

lib/Basic/PlatformUtility.cpp
```diff
@@ -38,9 +38,10 @@ using namespace llbuild::basic;
 
 bool sys::chdir(const char *fileName) {
 #if defined(_WIN32)
-  llvm::SmallVector<llvm::UTF16, 20> wFileName;
-  llvm::convertUTF8ToUTF16String(fileName, wFileName);
-  return SetCurrentDirectoryW((LPCWSTR)wFileName.data());
+  llvm::SmallVector<wchar_t, 128> wFileName;
```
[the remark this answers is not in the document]
daveinglis: updated to use MAX_PATH

owenv (approved):

daveinglis:
@swift-ci please smoke test

lib/Basic/PlatformUtility.cpp
```diff
@@ -38,9 +38,10 @@ using namespace llbuild::basic;
 
 bool sys::chdir(const char *fileName) {
 #if defined(_WIN32)
-  llvm::SmallVector<llvm::UTF16, 20> wFileName;
-  llvm::convertUTF8ToUTF16String(fileName, wFileName);
-  return SetCurrentDirectoryW((LPCWSTR)wFileName.data());
+  llvm::SmallVector<wchar_t, MAX_PATH> wFileName;
```
jakepetroules, line 41: What is MAX_PATH? Note that the true max path (with the `\\?\` prefix) is 32767, not 260.

Or is that just an initial allocation size, and it grows dynamically as needed?

lib/Basic/PlatformUtility.cpp
```diff
@@ -164,15 +171,72 @@ int sys::read(int fileHandle, void *destinationBuffer,
...
+  }
+  return -1;
 #else
   return ::rmdir(path);
 #endif
 }
 
 int sys::stat(const char *fileName, StatStruct *buf) {
 #if defined(_WIN32)
-  return ::_stat(fileName, buf);
```
jakepetroules, line 175: @compnerd Would it be simpler to use _wstat, _wrmdir, etc., vs the Win32 functions? They support long paths too, as they are just wrappers around Win32 anyways.

lib/Basic/PlatformUtility.cpp
daveinglis: yes, this is just the initial size

lib/Basic/PlatformUtility.cpp
daveinglis: so there would need some char to wchar_t conversion here to use those, so Im not sure it saves that much since the widePath API does this already.

daveinglis: do these _w functions accept the //?/ prefixed paths?

compnerd: The Microsoft implementation does support the NT style paths I believe but they are disjoint from Win32. e.g. using`SetEnvironmentVariableW` and `wgetenv` would give unexpected results.i think that using win32 might unify some behaviors.

bnbarham:
I mentioned this in the revert, but the issue here is that PR testing was testing tip rather than the PR itself. I'll edit this comment once that's fixed.

PR #1007, opened 2025-08-15
Revert "use Win32 file APIs and widePath for file ops to support long filenames"

bnbarham:
Reverts swiftlang/swift-llbuild#1004. Possibly the cause of Windows CI failures, will run cross PR testing to check.


bnbarham:
https://ci-external.swift.org/job/swift-PR-windows/44372/ is the swift cross-pr testing run. Though I'm not sure how they could be different 🤔

PR #1005, opened 2025-08-07
build: rename `libllbuild` to the proper `liblibllbuild`

compnerd:
Because `llbuild` is the executable, we cannot have another target with the name `llbuild` which is the library target. As such, the library is named `libllbuild`. This changes the output name of the library to `liblibllbuild`. This was being overridden which caused some trouble for builds which attempted to skirt the dependencies. Rename the output to the proper name.

PR #1011, opened 2025-08-30
Remove /INCREMENTAL:NO flag for llbuild-tool and libllbuild targets on Windows

jakepetroules:
These were originally added 6 years ago in #475 to "Disable incremental linking so llbuild.ilk's don't conflict" because at the time the command line tool and library targets were given the same OUTPUT_NAME, which conflicted. In #1005 these targets were given distinct names, so the workaround should no longer be necessary.

This helps unblock swiftlang/swift-build#757 because this flag is currently added without regard for the specific compiler driver in use, so the flag might not work based on whether clang-cl.exe or cl.exe vs clang.exe (where it would require a -Xlinker prefix), is used.

### DIFF
--- a/products/libllbuild/CMakeLists.txt
+++ b/products/libllbuild/CMakeLists.txt
@@ -21,10 +21,6 @@ target_link_libraries(llbuild PRIVATE
   llvmSupport
   SQLite::SQLite3)
 
-if(CMAKE_SYSTEM_NAME STREQUAL Windows)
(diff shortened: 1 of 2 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci smoke test Linux

owenv (approved):

PR #1013, opened 2025-09-17
Remove swift-build-tool executable

owenv:
Clients have long-since moved to using llbuild's library interface, and this tool is no longer used to bootstrap the toolchain (Its not the same thing as swift-build)


owenv:
@swift-ci test

jakepetroules (approved):

PR #1018, opened 2025-11-13
[Buildsystem] Replace pipes with shlex for command quoting

da-viper:
The pipes modules was removed in [python 3.13](https://docs.python.org/3/library/pipes.html)  causing the fedora build to fail. 

replace with `shlex` 

https://ci.swift.org/view/all/job/oss-swift-6.2-package-fedora-41


dmbryson:
@swift-ci test

da-viper:
@swift-ci test macOS

PR #1033, opened 2026-07-14
Ensure a rule scanning cycle does not leave behind dangling scan records

owenv:
If two rule scans form a cycle but there are no tasks running, the build could reach the end of executeTasks without successfully computing the requested key. I've observed this very rarely and unreliably in a complex project which I was unable to reduce to a smaller example. When it occurs, rule scan records from the "stuck" build containing dangling pointers would persist to the next build and lead to crashes when accessed, since they were never marked as cancelled. For now, if we detect this scenario, diagnose an error asking the user to file a bug, and call `cancelRemainingTasks` to cleanup any active rule scan records so we can recover without a crash.

### DIFF
--- a/lib/Core/BuildEngine.cpp
+++ b/lib/Core/BuildEngine.cpp
@@ -1066,6 +1066,18 @@ private:
           }
         }
 
+        // If two rule scans form a cycle but there are no tasks running, the build could reach this point
+        // without successfully computing the requested key. I've observed this very rarely and unreliably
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci smoke test

owenv:
@swift-ci test linux

dmbryson (approved):
Hmm, so a cycle in scanning that hasn’t left any pending tasks.  🤔

At any rate, the clean up looks reasonable.

owenv:
@swift-ci test linux

PR #1037, opened 2026-08-31
Update deployment targets

rconnell9:
Update deployment targets


rjmansfield (approved):

Xcode/Configs/Common.xcconfig
```diff
@@ -49,7 +49,7 @@ ALLOW_TARGET_PLATFORM_SPECIALIZATION = YES
 SUPPORTED_PLATFORMS = $(AVAILABLE_PLATFORMS)
 SUPPORTS_MACCATALYST = YES
 
-MACOSX_DEPLOYMENT_TARGET = 10.13
+MACOSX_DEPLOYMENT_TARGET = 15.0
 IPHONEOS_DEPLOYMENT_TARGET = 9.0
```
owenv: We could update these deployment targets too (all to 18), although I don't think they get used in practice

owenv (approved):

Xcode/Configs/Common.xcconfig
rconnell9: Having to do a Package.swift migration as well, since the current tools version is too old for the new deployment targets

PR #1047, opened 2026-09-15
Add an API for constructing a build description in memory, equivalent to a parsed manifest

rconnell9:
Add an API for constructing a build description in memory, equivalent to a parsed manifest.

Today, a client that builds a manifest programmatically currently has to serialize it to JSON and hand llbuild a file to parse back. On large projects that round-trip can cost seconds. This adds a `BuildDescriptionBuilder` clients can use directly, along with C API bindings (`llb_buildsystem_create_with_description` and the`llb_buildsystem_description_builder_*` family).

Both the in-memory API and YAML file paths converge in the BuildDescriptionBuilder, so the mechanisms for tool lookups, signature computation etc are the same.  What this means is that a llbuild description, whether passed in-memory or by file share identical signatures. In theory you could flip between each and there would be no impact to an incremental build.

### New build file keys

Also included are **environment bases**: a command that points at a shared named environment table and carries only its own overrides. Real manifests repeat huge environments, and the WebKit holds 5.73M bindings.

The `signature` attribute is deprecated. It is still honored, and nothing that uses it has to change.

Clients currently hash their own command lines and hand the digest over in`signature`, which replaced llbuild's computation outright. That meant every build paid for a full digest of every command up front, before the build system had decided anything was out of date. Two new attributes (signature-ignored-args and additional-signature-data) let a client keep the influence it actually needs without taking over the computation:

### Misc

This change is intended to be purely additive and opt-in.  It should not have any impact to exiting clients.

This change bumps the LLBUILD_C_API_VERSION 20.

### Perf

The included BuildDescriptionLoadPerfTest is disabled by default.  Below are the results using some contrived and some real-world manifests.

| profile | commands | env bindings | build file | |
| --- | ---: | ---: | ---: | --- |
| `webkit` | 25000 | 5740000 | 506.8 MiB | modeled on a measured 517 MiB WebKit manifest |
| `node-heavy` | 40000 | 0 | 104.8 MiB | no environment at all; node names are the whole cost |
| `balanced` | 5000 | 60000 | 6.9 MiB | the ordinary case, as a regression guard |

| profile | route | p50 | p90 | vs parse |
| --- | --- | ---: | ---: | ---: |
| `webkit` | parse | 2156.6 ms | 2525.2 ms | 1.00x |
| `webkit` | builder | 368.9 ms | 376.7 ms | 5.85x |
| `webkit` | builder + bases | 102.6 ms | 109.9 ms | 21.02x |
| `node-heavy` | parse | 289.6 ms | 301.4 ms | 1.00x |
| `node-heavy` | builder | 148.9 ms | 150.4 ms | 1.94x |
| `balanced` | parse | 32.7 ms | 33.7 ms | 1.00x |
| `balanced` | builder | 8.4 ms | 8.6 ms | 3.89x |
| `balanced` | builder + bases | 6.8 ms | 7.0 ms | 4.79x |


rconnell9:
@swift-ci test

rconnell9:
@swift-ci smoke test

include/llbuild/BuildSystem/BuildDescription.h
```diff
@@ -120,6 +156,13 @@ class BuildDescription {
   typedef llvm::StringMap<std::unique_ptr<Tool>> tool_set;
 
 private:
+  /// The shared environment tables commands may reference by name.
+  ///
+  /// Commands hold bare `EnvironmentBase*` pointers into this set, so it has to
```
owenv: this comment seems unnecessary

include/llbuild/BuildSystem/ShellCommand.h
```diff
@@ -125,8 +152,23 @@ class ShellCommand : public ExternalCommand {
...
-  getEnv() const { return env; }
+  getEnvOverrides() const { return envOverrides; }
+
+  /// The shared environment table this command's bindings layer over, if any.
+  const EnvironmentBase* getEnvBase() const { return envBase; }
+
+  /// Append the effective environment, the base composed with this command's
+  /// own bindings, to `result`.
+  ///
+  /// Deliberately not cached: it is needed only when the command is about to
```
owenv: this seems overly verbose

unittests/BuildSystem/BuildDescriptionLoadPerfTest.cpp
```diff
@@ -0,0 +1,731 @@
+//===- unittests/BuildSystem/BuildDescriptionLoadPerfTest.cpp -------------===//
```
owenv, line 1: I'm not sure this is worth checking in since we don't have infra to run it on at the moment

owenv (approved):
This is a big change, but the impact to the existing manifest loading path is small and the proposed API lgtm. I think we can continue refining it as we experiment with adopting it

owenv:
@swift-ci test windows

rconnell9:
@swift-ci smoke test

rconnell9:
@swift-ci test windows

rjmansfield (approved):
lgtm

rconnell9:
@swift-ci test

rconnell9:
@swift-ci smoke test

rconnell9:
@swift-ci test windows

PR #1048, opened 2026-09-18
[llbuildSwift] Make sure a command can be both `ExternalDetachedCommand` and `ProducesCustomBuildValue`

akyrtzi:
Previously only non-detached commands could be setup to have `ProducesCustomBuildValue` overrides.

There will be an upcoming test on the SwiftBuild side that this change will make pass.


akyrtzi:
@swift-ci smoke test

owenv (approved):

